PR #30, opened 2023-03-15
wrapper update with fixes

vhnatyk:
- Adds required wrappers for batched ntt/ecntt
- Cleanup
- Fixed tests
- Removed wip and danksharding-related stuff, drafts and stubs


alxiong:
is there any reason why the wrappers for EC-NTT in rust bindings are removed?
Or maybe any plan to add it?

omershlo:
indeed - it will be added back! 
@LeonHibnik @jeremyfelder @vhnatyk please share estimated timeline and workarounds

DmytroTym:
@alxiong we'll add them, we just weren't aware of any users, thanks for letting us know!

alxiong:
> we just weren't aware of any users

e.g. `fast-danksharding` which depends on a much older version of icicle (before it was cleaned up) uses `ecntt_batch`, [see here](https://github.com/ingonyama-zk/fast-danksharding/blob/f771e8cabc6f64d59c253827d2c0755f38ec9a61/fast-danksharding/src/fast_danksharding.rs#L70)

DmytroTym:
@alxiong yea that's true, we were going to reintroduce EC NTT for a new danksharding version should it be developed.

PR #64, opened 2023-04-19
New api

DmytroTym:
An attempt at creating new api based on data staying inside CUDA


icicle/appUtils/msm/msm.cu
```diff
@@ -180,7 +181,8 @@ void bucket_method_msm(unsigned bitsize, unsigned c, S *h_scalars, A *h_points,
   //split scalars into digits
   NUM_THREADS = 1 << 10;
   NUM_BLOCKS = (size * (nof_bms+1) + NUM_THREADS - 1) / NUM_THREADS;
-  split_scalars_kernel<<<NUM_BLOCKS, NUM_THREADS>>>(bucket_indices + size, point_indices + size, scalars, size, msm_log_size, nof_bms, bm_bitsize, c); //+size - leaving the first bm free for the out of place sort later
+  split_scalars_kernel<<<NUM_BLOCKS, NUM_THREADS>>>(bucket_indices + size, point_indices + size, on_device ? scalars : d_scalars, size, msm_log_size,
```
HadarIngonyama, line 184: instead of using "?" every time just use the same pointer. decide only at the beginning to copy or not. It's also confusing because d_scalars and scalars are actually both on device..

icicle/appUtils/msm/commit.cu
```diff
@@ -0,0 +1,57 @@
...
+ */
+extern "C"
+projective_t* commit_cuda(scalar_t* d_scalars, affine_t* d_points, size_t count, size_t device_id = 0)
+{
+    try
+    {
+        projective_t* d_out;
+        cudaMalloc(&d_out, sizeof(projective_t));
+        // TODO: set c depending on `count` instead of just 10
+        bucket_method_msm(scalar_t::NBITS, 10, d_scalars, d_points, count, d_out, true);
```
HadarIngonyama, line 22: the wrapper "large_msm" should be used instead of directly using "bucket_methd_msm" and the value of c needs to be determined there

icicle/appUtils/msm/msm.cu
```diff
@@ -144,16 +144,17 @@ __global__ void final_accumulation_kernel(P* final_sums, P* final_results, unsig
 
 //this function computes msm using the bucket method
 template <typename S, typename P, typename A>
-void bucket_method_msm(unsigned bitsize, unsigned c, S *h_scalars, A *h_points, unsigned size, P*h_final_result){
+void bucket_method_msm(unsigned bitsize, unsigned c, S *scalars, A *points, unsigned size, P* final_result, bool on_device) {
```
HadarIngonyama, line 147: should be separated to input_on_device and output_on_device

HadarIngonyama (changes requested):

vhnatyk (changes requested):
Overall the intention to create more convenient and more "research-friendly" higher-level APIs is cool :+1:, 

Must admit though - it feels like it needs to be a separate library similar to [poly-commit](https://github.com/arkworks-rs/poly-commit) by Arkworks for example. Also a lot of cudaMalloc/cudaMemcpy combined with parameters and if/else - feels like could be a source of a lot of bugs :bug: 

A simplified version of the flow for example for [Fast danksharding](https://github.com/ingonyama-zk/fast-danksharding) would be much simpler and maintainable - as a lot of the variables would be simply local - and all sitting on the device without extra if/host wrapping calls. So we could do those available externally for example on community requests via GitHub feature requests/issues etc.

icicle/appUtils/msm/msm.cu
DmytroTym: I kind of assume that everything is on device for simplicity

icicle/appUtils/vector_manipulation/ve_mod_mult.cu
```diff
@@ -7,92 +7,6 @@
 #include "ve_mod_mult.cuh"
 
 
-#define MAX_THREADS_PER_BLOCK 256
```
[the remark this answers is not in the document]
DmytroTym: Changed to only prototypes in lde.cuh but not in ve_mod_mult.cuh because if I try, a linking error appears. It basically says that functions from ve_mod_mult.cuh can't be found. I'm not a C++ expert but it seems like this can be relevant: https://stackoverflow.com/questions/495021/why-can-templates-only-be-implemented-in-the-header-file

icicle/appUtils/ntt/ntt.cuh
```diff
@@ -83,14 +82,55 @@ template < typename T > T * template_reverse_order(T * arr, uint32_t n, uint32_t
...
+
 /**
  * Cooley-Tuckey butterfly kernel. 
  * @param arr array of objects of type E (elements). 
  * @param twiddles array of twiddle factors of type S (scalars). 
  * @param n size of arr. 
  * @param n_twiddles size of omegas.
  * @param m "pair distance" - indicate distance of butterflies inputs.
- * @param i Cooley-TUckey FFT stage number.
+ * @param i Cooley-Tuckey FFT stage number.
```
vhnatyk, line 133: Tokey :)

icicle/appUtils/ntt/ntt.cuh
```diff
@@ -321,10 +360,13 @@ __global__ void ntt_template_kernel(E *arr, uint32_t n, S *twiddles, uint32_t n_
       uint32_t k = i + j + shift_s;
 
       uint32_t offset = (task / chunks) * n;
-      E u = arr[offset + i + j];
-      E v = twiddles[j * n_twiddles_div] * arr[offset + k];
-      arr[offset + i + j] = u + v;
-      arr[offset + k] = u - v;
+      if (!rev)
```
vhnatyk, line 363: pls review and try to use a local variable if possible - it like 2 read + 2 write (not counting twiddles) instead of 4 + 6 (including conditional) to global memory

vhnatyk (commented):
hmm
- fails to compile tests in release  `cargo test --release` 
- tests passing, but with `cargo test -- --test-threads=1`, as we discussed I suggest to make those failing optional, add  info to readme with feature and open issue
- `cargo build --release` builds successfully :+1:

icicle/appUtils/msm/commit.cu
```diff
@@ -0,0 +1,57 @@
...
+ * Commit to a batch of polynomials using the MSM.
+ * Note: this function just calls the MSM, it doesn't convert between evaluation and coefficient form of scalars or points.
+ * @param d_out Ouptut point to write the results to.
+ * @param d_scalars Scalars for the MSMs of all polynomials. Must be on device.
+ * @param d_points Points for the MSMs. Must be on device. It is assumed that this set of bases is used for each MSM.
+ * @param count Length of `d_points` array, `d_scalar` has length `count` * `batch_size`.
+ * @param batch_size Size of the batch.
+ */
+extern "C"
+projective_t* commit_batch_cuda(scalar_t* d_scalars, affine_t* d_points, size_t count, size_t batch_size, size_t device_id = 0)
```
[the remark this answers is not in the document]
vhnatyk: fixed

icicle/appUtils/ntt/ntt.cuh
DmytroTym: You're right, the old version is slightly faster, checked with criterion benchmarks for NTT (will include them in the next commit). Thought that removing 1 local variable would be worth it, but it's apparently outweighed by extra global memory accesses

DmytroTym:
Linking issue solved for now by just transferring code from keg.cu to mom.cu, thx @vhnatyk

vhnatyk (approved):

HadarIngonyama (approved):

PR #88, opened 2023-05-28
MSM for large sizes

HadarIngonyama:
- Fixed bugs in 2 kernels of the MSM, the kernels had threads that were accessing out-of-bound memory.
- The value of C is now chosen more optimally - as a function of the MSM size.
- Added a CUDA test file for debugging the MSM. Classes of dummy points and scalars can be used.
- MSM tested up to 2^24 against arkworks.

### DIFF
--- a/icicle/appUtils/msm/Makefile
+++ b/icicle/appUtils/msm/Makefile
@@ -1,0 +1,4 @@
+test_msm:
+	mkdir -p work
+	nvcc -o work/test_msm -I. tests/msm_test.cu
+	work/test_msm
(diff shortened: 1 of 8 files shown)


DmytroTym:
Generating a large number of random points takes a really long time so I changed it to this:
`pub fn generate_random_points_bls12_381(
    count: usize,
    mut rng: Box<dyn RngCore>,
) -> Vec<PointAffineNoInfinity_BLS12_381> {
    if count < (1 << 15) {
        return (0..count)
        .map(|_| Point_BLS12_381::from_ark(G1Projective_BLS12_381::rand(&mut rng)).to_xy_strip_z())
        .collect();
    }
    let a: Vec<PointAffineNoInfinity_BLS12_381> = (0..1 << 15)
        .map(|_| Point_BLS12_381::from_ark(G1Projective_BLS12_381::rand(&mut rng)).to_xy_strip_z())
        .collect();
    vec![a; ((count - 1) >> 15) + 1].concat()[..count].to_vec()
}`
Also, seems like (at least on our 3090Ti) sizes of 2^25 and 2^26 work as well. Haven't tested anything larger.

DmytroTym (approved):

PR #102, opened 2023-06-08
Fix: Docs, Curve generation script

### LINKED ISSUE #101: CUDA compilation documentation is incorrect
The [build instructions for cuda](https://github.com/ingonyama-zk/icicle#build-and-usage) in the README are not updated to reflect the reorganization of code after supporting multiple curves.

jeremyfelder:
## Describe the changes

This PR:
- Updates the python curve generation script to reflect recent changes and using Template instead of formatted strings
- Updates the constant params in curves
- Updates documentation on compiling the CUDA side

## Linked Issues

Resolves #101


curve_parameters/new_curve_script.py
```diff
@@ -33,157 +15,251 @@ def to_hex(val, length):
...
-    file_content+="}\n"
-    return file_content
+        s += f'0x{c}, '
+        
+    return s[:-2]
+
+
+def get_root_of_unity(modulus_p, order: int) -> int:
+    assert (modulus_p - 1) % order == 0
+    return pow(5, (modulus_p - 1) // order, modulus_p)
```
ImmanuelSegol, line 25: why 5?

can u use a roo_of_unity?

this is what my json looks like for bn254:

```
{
    "curve_name" : "bn254",
    "modolus_p" : 21888242871839275222246405745257275088548364400416034343698204186575808495617,
    "bit_count_p" : 254,
    "limb_p" :  8,
    "ntt_size" : 16,
    "modolus_q" : 21888242871839275222246405745257275088696311157297823662689037894645226208583,
    "bit_count_q" : 254,
    "limb_q" : 8,
    "weierstrass_b" : 3,
    "gen_x" : 1,
    "gen_y" : 2,
    "root_of_unity": 421743594562400382753388642386256516545992082196004333756405989743524594615
}
```

I then did:

```
def get_root_of_unity(order: int) -> int:
    assert (modolus_p - 1) % order == 0
    return pow(root_of_unity, (1 << ntt_size) // order, modolus_p)
```

jeremyfelder: @ImmanuelSegol I'm not sure.. the only change on these lines is whitespace.

@guy-ingo Whats the reason behind a hard coded 5 here?

DmytroTym: @jeremyfelder yeah 5 being hard-coded there is not your change. But IMO we do need to use root of unity and not just 5 because first of all, 5 might not be the primitive root of the multiplicative subgroup of some fields, which would mean that it generates invalid twiddle factors. And secondly, we will want to use the exact same twiddle factors that the applications we want to support use, so being able to choose here is important

DmytroTym (dismissed):
Looks good to me

guy-ingo (dismissed):

guy-ingo (approved):

vhnatyk (changes requested):
still no regenerated omegas for bn254 ? :)

vhnatyk (approved):

DmytroTym:
One question: why is there no `i_squared` and `i_squared_is_negative` fields in json or in curve script?

DmytroTym (approved):

PR #106, opened 2023-06-13
Fix curve config location and link to some namespace struct members

### LINKED ISSUE #103: [BUG]: CUDA test suite fails
## Description

When running the CUDA test suite, the following error is encountered:

`icicle/icicle/primitives/test_kernels.cuh:4:10: fatal error: ../curves/bls12_381.cuh: No such file or directory`

### Reproduce

Follow the instructions for building and testing in the README

## Expected Behavior

The tests should run and pass

## Environment

Please complete the following information:

OS + Version:

Cargo Version:

GPU type:

gkigiermo:
This PR fixes the compilation of the Google tests.
The header with the curve configuration was not pointing on the correct path. 
Some static struct members needed a definition for solving linking problems.

Resolves #103

### DIFF
--- a/icicle/curves/bls12_381/params.cuh
+++ b/icicle/curves/bls12_381/params.cuh
@@ -190,11 +190,4 @@ namespace PARAMS_BLS12_381{
                                                                           0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000};
   static constexpr storage<fq_config::limbs_count> weierstrass_b_g2_im = {0x00000004, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000,
                                                                           0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000};
-
-  constexpr storage<fq_config::limbs_count> fq_config::m;
-  constexpr storage<fq_config::limbs_count> fq_config::modulus;
-  constexpr storage<2*fq_config::limbs_count> fq_config::modulus_wide;
-  constexpr storage<fp_config::limbs_count> fp_config::m;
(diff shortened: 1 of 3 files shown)


icicle/curves/bls12_381/params.cuh
```diff
@@ -190,4 +190,11 @@ namespace PARAMS_BLS12_381{
...
                                                                           0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000};
-}
\ No newline at end of file
+
+  constexpr storage<fq_config::limbs_count> fq_config::m;
+  constexpr storage<fq_config::limbs_count> fq_config::modulus;
+  constexpr storage<2*fq_config::limbs_count> fq_config::modulus_wide;
+  constexpr storage<fp_config::limbs_count> fp_config::m;
+  constexpr storage<fp_config::limbs_count> fp_config::modulus;
+  constexpr storage<2*fp_config::limbs_count> fp_config::modulus_wide;
```
jeremyfelder, line 199: @gkigiermo I think we can remove these if we require c++17 standard in the CMakeLists for both [CXX and CUDA](https://github.com/ingonyama-zk/icicle/blob/main/icicle/CMakeLists.txt#L4-L5) which would require changing the minimum cmake version to 3.18.

jeremyfelder (changes requested):
Thanks for contributing!
If you have time, I'd recommend one small change.

icicle/curves/bls12_381/params.cuh
gkigiermo: You are right; I just did it.

jeremyfelder (approved):

PR #128, opened 2023-07-18
Fix/cudacodegoicile

ImmanuelSegol:
## Describe the changes

This PR...

## Linked Issues

Resolves #


LeonHibnik:
@ImmanuelSegol for setup.sh add instructions to source it

goicicle/setup.sh
```diff
@@ -0,0 +1,27 @@
...
+fi
+
+if [[ "$TARGET_BN254_PATH" != "" ]]; then
+    echo "BN254 found @ $TARGET_BN254_PATH"
+    export LD_LIBRARY_PATH=$TARGET_BN254_PATH
+fi
+
+if [[ "$TARGET_BLS12_381_PATH" != "" ]]; then
+    echo "BN254_PATH @ $TARGET_BLS12_381_PATH"
+    export LD_LIBRARY_PATH=$TARGET_BLS12_381_PATH
```
LeonHibnik, line 26: LD_LIBRARY_PATH can include other paths so for not breaking it, fix it to add our path to current, not overwrite it

go.mod
jeremyfelder, line 1: Whats the purpose for moving this to the root of the repo? I think it would make sense to have both Go and Rust code in a bindings directory at some point

LeonHibnik: it's a way to add the cuda code to the go package repo

PR #130, opened 2023-07-20
Fix/curves

ImmanuelSegol:
## Describe the changes

This PR...

## Linked Issues

Resolves #


goicicle/curves/bls12377/msm.go
jeremyfelder, line 1: Two things about the API
- CommitG2Batch is missing
- Its confusing to have both Commit and MSM as function names when the only different is where the input data is...maybe it makes more sense to expose a single MSM API with an `onDevice` flag to switch between which cpp function is called?

jeremyfelder (commented):
Only left comments on BLS12_377 but it looks like each comment would apply to the other curves as well.

goicicle/curves/bls12377/g1.go
```diff
@@ -1,395 +1,317 @@
...
+}
 
-	return &ScalarField{s}
+func (a *G1ScalarField) Equals(b *G1ScalarField) bool {
+	for i, v := range a.S {
+		if b.S[i] != v {
+			return false
+		}
+	}
+	return true
```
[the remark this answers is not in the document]
ImmanuelSegol: do we have this in C++? I cant find it, I also don't see this in g2.

goicicle/curves/bls12377/msm.go
ImmanuelSegol: - thanks added
- I opened an issue for this, I want there to be a CUDA and Golang PR which will sort this issue out.

jeremyfelder (approved):

PR #132, opened 2023-07-24
Add language formatters

jeremyfelder:
## Describe the changes

This PR adds a git hook script that will run formatters for Rust, Go, and CUDA/CPP/C files.
The README has been updated to instruct how this script should be applied to git hooks.

NOTE: clippy was not included for Rust for the following reasons:
- it also reports warnings from CUDA/CPP/C
- not all of its warnings/errors are easily fixable by `clippy --fix`
- It takes a very long time and would disrupt the development process. It will be run in CI


ImmanuelSegol (approved):

scripts/pre-commit.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+# Find CUDA and CPP files and apply clang-format 
+find ./ -iname *.h -or -iname *.cuh -or -iname *.cu | xargs clang-format -i -style=file
```
vhnatyk, line 4: what about .c / .cpp files?

jeremyfelder: > what about .c / .cpp files?

Can definitely add. I didn't see that we have any files that are strictly .c/.cpp right now so I didn't put them in.

jeremyfelder: @vhnatyk I kept the single script for this since each language tries to write their own scripts to the .git/hooks directory and they were conflicting with each other. I added a README section on how to configure the hook scripts

scripts/hooks/pre-push
```diff
@@ -0,0 +1,30 @@
...
+# Run go fmt across all Golang packages
+if [[ $(go list ./... | xargs go fmt) ]];
+then
+    echo "🚨 There are Golang files that need formatting."
+    echo "Please commit the formatted files"
+    status=1
+fi
+
+# Run cargo fmt on Rust files
+if [[ $(cargo fmt --check) ]];
```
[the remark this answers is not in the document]
vhnatyk: a bit uncommon - but tried it better - and seems indeed more convenient to have formatting manual and in separate commit

vhnatyk (approved):

PR #134, opened 2023-07-27
Karatsuba 1.0

DmytroTym:
This PR introduces a new improved multiplier

### DIFF
--- a/icicle/primitives/field.cuh
+++ b/icicle/primitives/field.cuh
@@ -197,23 +197,25 @@ template <class CONFIG> class Field {
       }
     }
 
+    template <bool SUBTRACT, bool CARRY_OUT> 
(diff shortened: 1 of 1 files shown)


IdoAtlas (commented):
have no comments on code, if it works correctly that's fine.

first, it important that we will have stand-alone benchmark of multiplier for 384 bit field and 256 bit field
each benchmark measures throughput (mults / sec, mod mults / sec) and power (to calculate mod mults / watt). should have those target optimization parameters in front of our eyes at every optimization iteration.

for future improvements:
1. detailed performance breakdown of the mod-mult (how much does each part takes?)
specific points:
* i encountered a way to get rid of the barrel-shifts, want to see what is the actual impact, if this is a signficant problem let's implement the new method.
* reduce redundency and msb multiplier

2. understanding the karatsuba gains, where do they come from? can we benefit from breaking it another level? can we use ADK for better results?

DmytroTym:
Closed in favour of https://github.com/ingonyama-zk/icicle/pull/289

PR #143, opened 2023-08-11
Correct MSM for weird scalar distributions

DmytroTym:
This PR resolves the correctness issues in MSMs with skewed distributions of scalars observed in different gnark circuits. If I have some time on the weekend, I'll also follow-up with performance improvements for the most skewed cases, but this can be merged as-is to unblock support for all gnark circuits.


icicle/appUtils/msm/msm.cu
```diff
@@ -135,9 +135,9 @@ __global__ void find_cutoff_kernel(unsigned *v, unsigned size, unsigned cutoff,
       result[0] = i+1;
       return;
     }
-    if (i == size - 1) {
-      result[0] = 0;
-    }
+  }
+  if (v[size - 1] > cutoff) {
+    result[0] = size;
```
HadarIngonyama, line 140: why did you change this? also, now all the threads write to result[0] instead of a single thread..

icicle/appUtils/msm/msm.cu
```diff
@@ -427,6 +427,7 @@ void bucket_method_msm(unsigned bitsize, unsigned c, S *scalars, A *points, unsi
   unsigned bucket_th = large_bucket_factor*avarage_size;
   unsigned *nof_large_buckets;
   cudaMallocAsync(&nof_large_buckets, sizeof(unsigned), stream);
+  cudaMemset(nof_large_buckets, 0, sizeof(unsigned));
```
HadarIngonyama, line 430: why is this needed?

icicle/appUtils/msm/msm.cu
```diff
@@ -479,19 +481,22 @@ void bucket_method_msm(unsigned bitsize, unsigned c, S *scalars, A *points, unsi
...
+  if (h_nof_buckets_to_compute > h_nof_large_buckets) {
+    NUM_THREADS = 1 << 8;
+    NUM_BLOCKS = (h_nof_buckets_to_compute-h_nof_large_buckets + NUM_THREADS - 1) / NUM_THREADS;
+    accumulate_buckets_kernel<<<NUM_BLOCKS, NUM_THREADS, 0, stream>>>(buckets, sorted_bucket_offsets+h_nof_large_buckets, sorted_bucket_sizes+h_nof_large_buckets, sorted_single_bucket_indices+h_nof_large_buckets, point_indices, 
+                                                            d_points, nof_buckets, h_nof_buckets_to_compute-h_nof_large_buckets, c+bm_bitsize, c);
+  }
+
+  // all the large buckets need to be accumulated before the final summation
   cudaStreamSynchronize(stream2);
-  cudaStreamDestroy(stream2);
```
HadarIngonyama, line 487: why not destroy the second stream?

HadarIngonyama (commented):
So from what I see this PR adds small fixes to the MSM and adds a rust tests that enables testing different distributions? But doesn't include yet a solution for separating very large scalars and using SSM?

icicle/appUtils/msm/msm.cu
DmytroTym: Tbh I deleted it by accident (along with cudaStreamSyncronize() which gave me a lot of extra debug time lol). I'll put it back

icicle/appUtils/msm/msm.cu
DmytroTym: Tbh I don't understand the original version. First of all, `i` is never `size - 1` because the loop terminates at `i<min(start_index+run_length,size-1)`, so your block is never executed, this breaks correctness (for 0;1;-1 distribution for example). And judging from the condition, it is supposed to work when all the nonzero buckets are large, right? In that case, why would `result` be 0 and not `size`? In the case of 0;1;-1 distribution I would like -1 buckets to be treated as large and not as small

DmytroTym: About threads writing simultaneously - I agree that looks awkward but because the value is the same for all threads, it's not a data race, and I don't think there's a performance penalty as one thread has to diverge to return the result anyway

icicle/appUtils/msm/msm.cu
DmytroTym: I don't really see how the case of zero large buckets is handled. If there are no large buckets, `find_cutoff_kernel` just doesn't modify the `nof_large_buckets` variable, so it stays whatever it was when allocated on line 429. Because `cudaMallocAsync` [doesn't clear](https://stackoverflow.com/a/62055960) the memory, we potentially  might get gibberish instead of zero in `nof_large_buckets`. Admittedly, I never actually got gibberish but I think it's better to do it the safe way

DmytroTym:
Comparing [sha256 circuit with inputLen=16384](https://github.com/celer-network/zk-benchmark/tree/main/gnark) with and without icicle on our lab machine with RTX 3090Ti.
CPU version without icicle:

> DBG prover done backend=groth16 curve=bn254 nbConstraints=11207992 took=9204.99307


Icicle version using this commit:

> DBG Icicle API: computeH took=477.582926
> DBG Icicle API: MSM BS1 MSM backend=groth16 curve=bn254 nbConstraints=11212315 took=38.229246
> DBG Icicle API: MSM AR1 MSM backend=groth16 curve=bn254 nbConstraints=11212315 took=61.28084
> DBG Icicle API: MSM KRS2 MSM backend=groth16 curve=bn254 nbConstraints=11212315 took=211.067074
> DBG Icicle API: MSM KRS MSM backend=groth16 curve=bn254 nbConstraints=11212315 took=88.270904
> DBG Icicle API: MSM G2 BS backend=groth16 curve=bn254 nbConstraints=11212315 took=74.128782
> DBG Total MSM time backend=groth16 curve=bn254 nbConstraints=11212315 took=501.157817
> DBG prover done; TOTAL PROVE TIME backend=groth16 curve=bn254 nbConstraints=11212315 took=1065.513368

icicle/appUtils/msm/msm.cu
HadarIngonyama: I already answered on slack.. the fix should be i<min(start_index+run_length,size), then the last thread (and only it) always reaches size -1. I put 0 so the regular bucket kernel would run, but I guess putting size is better yeah.

DmytroTym: In this case we'll be accessing `v[size + 1]` which hypothetically might cause a segfault, and even if the condition `if (i == size - 1)` is put before the main body of the loop to avoid this, you'll still need to add `v[size - 1] > cutoff` condition to it, otherwise you're creating a data race, no?
And checking the condition on each iteration of the loop is less efficient than just once in the end too...

HadarIngonyama (approved):

PR #162, opened 2023-08-23
Fixing batched MSM

DmytroTym:
This PR aims to fix batch MSM tests in `dev` at minimum and preferably make batched performance as good as single MSM performance. In which case we might only leave one function to avoid duplication of code between batched/single MSM


jeremyfelder (approved):
Tested using Rust tests

Issue #113, opened 2023-06-28
[FEAT]: Make `i_squared` and `i_squared_is_negative` configurable


jeremyfelder:
## Description

Make `i_squared` and `i_squared_is_negative` configurable from each curve's parameter json file and update the new curve script to correctly templatize these params.

See here for more information: https://github.com/ingonyama-zk/icicle/pull/102#issuecomment-1611308059

PR #188, opened 2023-09-14
Support bw6-761

liuxiaobleach:


DmytroTym:
I'll try to explain my latest commit and since it's quite a bit of code, I'll mention people who wrote the code I modified or are otherwise necessary to ask
First, this PR aims to add BW6-671 curve to icicle. Only CUDA part is implemented in the latest commit, with wrappers (first and foremost go) still to be written. @liuxiaobleach instead of declaring scalar field for this curve separately, I re-used base field of BLS12-377, hopefully that makes sense. I also needed to add twiddles and other NTT-related info to BLS12-377's base field which is not generated by our script, so we should avoid re-using the script for this curve. But I guess it's the case anyway because of the reversed twiddles for the scalar field @jeremyfelder. More generally, I found our script to be way too restrictive in creating a new curve. I feel like @ChickenLover who is now implementing fields for STAKS might feel the same way. It's nice to have something that helps you generate certain parameters but very often we need to edit script-generated files which is probably not a good sign for this approach
@vhnatyk I removed `scalar_field_t` and only left `scalar_t`, I don't really see a reason for the two of them to exist
@ImmanuelSegol I changed the script in several places, some fixes, but also to include nonresidue value that was hardcoded before as well as cherrypicking @LeonHibnik 's PR that added Montgomery values. Also please check my edits to README about how to use the script, some of them are just my personal preferences though tbh
A couple of issues with this commit: first, @jeremyfelder some BLS12-381 CUDA files were edited by the script, seems like all the tests pass, but maybe I should revert these changes. Second, @jeremyfelder 4 coset evaluation themed tests fail for BLS12-377, I'm assuming because of the flipped twiddles. @ChickenLover I had to comment out `poseidon_multi_cuda_bls12_381` function and corresponding test because there's no extern equivalent on the CUDA side

icicle/curves/bls12_381/supported_operations.cu
```diff
@@ -1,5 +1,4 @@
 #include "lde.cu"
 #include "msm.cu"
-#include "poseidon.cu"
 #include "projective.cu"
-#include "ve_mod_mult.cu"
+#include "ve_mod_mult.cu"
```
ImmanuelSegol, line 4: best to add another line

icicle/curves/index.cu
```diff
@@ -1,3 +1,4 @@
 #include "bls12_377/supported_operations.cu"
 #include "bls12_381/supported_operations.cu"
-#include "bn254/supported_operations.cu"
\ No newline at end of file
+#include "bn254/supported_operations.cu"
+// #include "bw6_761/supported_operations.cu"
```
ImmanuelSegol, line 4: why is this commented out ?

src/test_bls12_381.rs
```diff
@@ -9,14 +9,14 @@ use rustacuda_core::DevicePointer;
...
 extern "C" {
-    fn poseidon_multi_cuda_bls12_381(
-        input: *const ScalarField_BLS12_381,
-        out: *mut ScalarField_BLS12_381,
-        number_of_blocks: usize,
-        arity: c_uint,
-        device_id: usize,
-        stream: usize, // TODO: provide a real stream
-    ) -> c_uint;
+    // fn poseidon_multi_cuda_bls12_381(
```
ImmanuelSegol, line 12: why is all this commented out?

ImmanuelSegol (changes requested):

jeremyfelder:
> first, @jeremyfelder some BLS12-381 CUDA files were edited by the script, seems like all the tests pass, but maybe I should revert these changes.

This is fine. I think those functions were missing from this curve since we haven't run the generation script on all curves in a little bit.

>Second, @jeremyfelder 4 coset evaluation themed tests fail for BLS12-377, I'm assuming because of the flipped twiddles.

I assume you are talking about Rust tests. Makes sense from flipped twiddles. Shouldn't be too difficult to add an option to reverse the twiddles in the short term. Long term we the solution of generating twiddles on the fly.

curve_parameters/new_curve_script.py
```diff
@@ -14,25 +17,28 @@ def to_hex(val: int, length):
     n = 8
     chunks = [x[i:i+n] for i in range(0, len(x), n)][::-1]
     s = ""
-    for c in chunks:
+    for c in chunks[:length // n]:
```
jeremyfelder, line 20: Why is this necessary?

src/test_bls12_381.rs
jeremyfelder: >I had to comment out poseidon_multi_cuda_bls12_381 function and corresponding test because there's no extern equivalent on the CUDA side

curve_parameters/new_curve_script.py
DmytroTym: It's for a weird corner case. Most of the time, `length // n` is the length of `chunks`, but sometimes the number `val` passed into `to_hex` function can be larger than expected. For example, `4 * modulus` might be 257 bits long for 255-bit moduli. In which case an extra limb will be added leading to a compile error. Of course, trimming `chunks` like I'm doing is also unsafe - if `4 * modulus` in the above example is used at any point, this will lead to a silent failure during runtime. Currently, the only field with this issue is the scalar field of BLS12-381 curve which doesn't use `4 * modulus` anywhere. But in principle this is problematic and we should probably rewrite `field.cu` in a way that allows removing such overflowing values from the params file with compile time errors when trying to use them

icicle/curves/index.cu
DmytroTym: Cos the wrappers aren't ready yet and I figured that this line is useless for now. I'll reactivate it once wrappers are implemented

DmytroTym:
Closes https://github.com/ingonyama-zk/icicle/issues/191 and https://github.com/ingonyama-zk/icicle/issues/113

src/test_bls12_381.rs
jeremyfelder: @DmytroTym It seems there is a corresponding method in cuda that is extern "C": https://github.com/celer-network/icicle/blob/main/icicle/curves/bls12_381/poseidon.cu#L10

Can we uncomment these functions and the test?

icicle/curves/bls12_377/params.cuh
```diff
@@ -142,6 +142,7 @@ namespace PARAMS_BLS12_377 {
 
   struct fq_config {
     static constexpr unsigned limbs_count = 12;
+    static constexpr unsigned omegas_count = 48;
```
jeremyfelder, line 145: This is nitpicky, but it looks like this was manually changed? Do we want to also change this [here](https://github.com/celer-network/icicle/blob/main/curve_parameters/bls12_377.json#L6)

goicicle/curves/bw6761/msm_test.go
```diff
@@ -0,0 +1,367 @@
...
+	LOG_MSM_SIZES := []int{20, 21, 22, 23, 24, 25, 26}
+
+	for _, logMsmSize := range LOG_MSM_SIZES {
+		msmSize := 1 << logMsmSize
+		points := GeneratePoints(msmSize)
+		scalars := GenerateScalars(msmSize, false)
+
+		out_d, _ := goicicle.CudaMalloc(96)
+
+		pointsBytes := msmSize * 64
```
DmytroTym, line 172: Pretty sure there's a problem somewhere in here, too many hardcoded values

icicle/curves/bls12_377/params.cuh
DmytroTym: This is actually a situation not accounted for by the script. You can notice that `fq_config` is not the scalar field, but the base field. And the script only allows for NTT parameters in the scalar field because most of the time, scalar field is the only one we compute NTTs over. BLS12-377 is special because its base field is designed to be the scalar field of another curve - BW6-671. That's why `omegas_count` and twiddles have been added to the base field by hand here.
An alternative option would be to repeat this field in the BW6-671 config file which would align the codebase with the script better. But we decided not to do this to avoid code duplication, plus I think having one less field would decrease compilation time and the size of our static lib at least a bit.

src/curves/bw6_761.rs
```diff
@@ -0,0 +1,329 @@
+use std::ffi::c_uint;
```
DmytroTym, line 1: Oh, this file shouldn't be here, somehow forgot to delete it or/and missed it's autogenerated reintroduction

src/test_bw6_761.rs
```diff
@@ -0,0 +1,1482 @@
+use std::ffi::{c_int, c_uint};
```
DmytroTym, line 1: I distinctly remember deleting this :/

src/test_bls12_381.rs
DmytroTym: Ah, you're right, I think the issue is just that `poseidon.cu` is not included into `supported_operations.cu` for bls12-381. Does it make sense @ChickenLover ? Am trying to include and rebuild now. Since we're fixing Rust tests, can I also `#[ignore]` the currently failing bls12-377 tests, until we fix it? It's an eyesore :(

DmytroTym (dismissed):
This is not a fully complete implementation of BW6-671 as Rust wrappers are still missing, but for the sake of C++ and Golang I think this can be merged into main. One slight issue I have with Golang wrappers is that the tests are pretty weak and don't cover all functionality as well as don't check correctness agains gnark (or any other CPU library). This is because interaction with gnark has been factored into the iciclegnark repo.
Either way, we have correctness in gnark's groth16 prover, and IMO it's best to address aforementioned issues as part of the new API.

goicicle/curves/bw6761/msm_test.go
DmytroTym: But we discussed offline that these things will be fixed down the line, for now we don't really use this function

ImmanuelSegol (dismissed):

jeremyfelder (approved):

Issue #193, opened 2023-09-19
[BUG]: insufficient dependencies 


maciejskorski:
## Description

I can't build tests although dependencies seem OK. 

### Reproduce

```bash
mkdir -p build
cmake -S . -B build
cmake --build build
cd build && ctest

[ 18%] Built target gtest
[ 36%] Built target gtest_main
[ 45%] Building CUDA object CMakeFiles/primitives_test.dir/primitives/test.cu.o
/usr/include/boost/multiprecision/cpp_int/serialize.hpp(49): error: function "boost::serialization::make_nvp(const char *, T &) [with T=bool]" returns incomplete type "const boost::serialization::nvp<bool>"
     ar & make_nvp("sign", s);
                  ^

/usr/include/boost/multiprecision/cpp_int/serialize.hpp(52): error: function "boost::serialization::make_nvp(const char *, T &) [with T=std::size_t]" returns incomplete type "const boost::serialization::nvp<std::size_t>"
     ar & make_nvp("byte-count", byte_count);
                  ^

/usr/include/boost/multiprecision/cpp_int/serialize.hpp(62): error: function "boost::serialization::make_nvp(const char *, T &) [with T=unsigned char]" returns incomplete type "const boost::serialization::nvp<unsigned char>"
           ar & make_nvp("byte", byte);
                        ^

3 errors detected in the compilation of "/home/icicle/icicle/primitives/test.cu".
make[2]: *** [CMakeFiles/primitives_test.dir/build.make:77: CMakeFiles/primitives_test.dir/primitives/test.cu.o] Error 2
make[1]: *** [CMakeFiles/Makefile2:137: CMakeFiles/primitives_test.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```

## Expected Behavior

I would expect the compilation to go through. 

I think it would be nice to demonstrate on CI/CD that the code actually builds in a fresh environment.

## Environment

Please complete the following information:

OS + Version: Linux 4654ec4915e6 5.19.0-1025-aws #26~22.04.1-Ubuntu SMP Mon Apr 24 01:58:15 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

GPU type: Tesla V100-SXM2-16GB

CUDA Version (nvcc): Cuda compilation tools, release 12.2, V12.2.91

CMake Version: 3.27.5

ImmanuelSegol:
@maciejskorski Thanks for reporting, I'm looking into this now.

[...]

PR #197, opened 2023-09-19
fix cuda test - remove boost

ImmanuelSegol:
## Describe the changes

This PR...

## Linked Issues

Resolves #
#193

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+# Use the specified base image
+FROM nvidia/cuda:12.0.0-devel-ubuntu22.04
+
+# Update and install dependencies
+RUN apt-get update && apt-get install -y \
+    cmake \
+    protobuf-compiler \
+    curl \
+    build-essential \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install Rust
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+ENV PATH="/root/.cargo/bin:${PATH}"
+
+# Install Golang
+ENV GOLANG_VERSION 1.21.1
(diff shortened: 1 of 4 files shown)


icicle/primitives/test.cu
```diff
@@ -406,92 +395,6 @@ TEST_F(PrimitivesTest, INGO_MP_MULT)
   }
 }
 
-TEST_F(PrimitivesTest, INGO_MP_MSB_MULT)
```
vhnatyk, line 409: the primitive tests are important - maybe we can pre-install or replace boost with something more common

vhnatyk: since we don't need the library for comparing to be CUDA compatible - maybe GMP will work - should be ultimately common on Ubuntu (or at least I can't think of any more common library like that 🙂 ) @DmytroTym  wdyt?

maciejskorski: @vhnatyk 
Keeping in mind that I am not a member but a user - I understand the fix is to remove tests that don't compile, but this sounds to me like a weak solution. Wouldn't be possible to upgrade the tests so that they are compatible?

Dockerfile
```diff
@@ -0,0 +1,28 @@
+# Use the specified base image
+FROM nvidia/cuda:12.0.0-devel-ubuntu22.04
+
+# Update and install dependencies
+RUN apt-get update && apt-get install -y \
+    cmake \
```
maciejskorski, line 6: My two cents: although I suggested a Docker I think it would be good to implement it under a different PR because this requires some more testing. The first example - taking cmake from Ubuntu 20.x repo results in an old version causing compilation problems.

icicle/primitives/test.cu
vhnatyk: ok resolving then - otherwise all good @ImmanuelSegol  👍🏻

vhnatyk (approved):

PR #214, opened 2023-09-20
convert some of the todos to issues :)

ImmanuelSegol:
## Describe the changes

This PR removes TODOs in favor of issues I created.

## Linked Issues

Resolves #


jeremyfelder:
I'm not sure this is the best idea for all TODOs. There are now areas of the code that look very strange and not everyone knows all of the issues to understand why they are like that. I think this might lead to removing or adding things that are not ready yet

Issue #194, opened 2023-09-19
[IMPROVEMENTS] - remove curve name prefixes


ImmanuelSegol:
Remove curve name prefix (example: bn254_msm).

The package name should contain the curve prefix and functions within the package should just carry their own names.

DmytroTym:
We never fully removed them but made them as invisible as possible. We still need different names on the boundary between C++ and wrappers so I guess the current version is as good as it gets.

Issue #213, opened 2023-09-20
[FEAT] Implement error codes


ImmanuelSegol:
- Define error codes for different classes of errors. 
- Make sure errors return information about the location they have been emmited from
- Create verbose error messages.

DmytroTym:
We have some duplicates (or near-duplicates) for this one: https://github.com/ingonyama-zk/icicle/issues/109, https://github.com/ingonyama-zk/icicle/issues/38 and https://github.com/ingonyama-zk/icicle/issues/190

vhnatyk:
Added in v1.0.0

PR #217, opened 2023-09-20
Cuda refactoring

DmytroTym:
This PR aims to refactor APIs on CUDA side as well as on the wrapper side


DmytroTym:
@HadarIngonyama pls take a look at the new MSM interfaces and see if they make sense to you.
Due to numerous breaking changes, wrappers don't work rn, but you can try googletests or just building shared libraries using cmake:

- In googletests, add `-DBUILD_TESTS=ON -DCURVE=254`  to the usual `cmake -S . -B build` (if you want to test bn254 curve).
- For building the `.a` library, run from `icicle/icicle`: `cmake -DBUILD_TESTS=OFF -DCURVE=254 -DCURVE_NAME=bn254 -DLIBRARY_OUTPUT_DIRECTORY={folder to save the library into}`

DmytroTym:
Closes https://github.com/ingonyama-zk/icicle/issues/194

icicle/appUtils/msm/msm.cu
```diff
@@ -880,109 +882,137 @@ void batched_bucket_method_msm(
...
+/**
+ * Extern version of [msm_internal](@ref msm_internal) function with the following values of template parameters 
+ * (where the curve is given by `-DCURVE` env variable during build):
+ *  - `S` is the [scalar field](@ref scalar_t) of the curve;
+ *  - `A` is the [affine representation](@ref affine_t) of curve points;
+ *  - `P` is the [projective representation](@ref projective_t) of curve points.
+ * @returns 0 if no error occured and -1 otherwise.
+ */
+extern "C" int msm_internal_cuda(
+  curve_config::scalar_t scalars[],
```
vhnatyk, line 929: isn't curve_config::scalar_t scalars* more common in our code?

icicle/appUtils/msm/msm.cuh
```diff
@@ -1,49 +1,74 @@
...
-  unsigned msm_size,
-  P* final_results,
-  bool on_device,
-  cudaStream_t stream);
+/**
+ * @struct MSMConfig
+ * Struct that encodes MSM parameters to be passed into the [msm_internal](@ref msm_internal) function.
+ */
+struct MSMConfig {
+  bool scalars_on_device;         /**< True if scalars are on device and false if they're on host. Default value: false. */
```
vhnatyk, line 25: maybe we can denote bool params with some verbal prefix?

icicle/appUtils/msm/msm.cu
DmytroTym: tbh I just copied the code that we had before. can rename for consistency, there aren't significant differences between the two though, right?

DmytroTym:
One thing that's very much missing is error handling (see https://github.com/ingonyama-zk/icicle/issues/213). Can I suggest that it's implemented in a separate PR while for now I'll just use `cudaError_t` from `cuda_runtime_api.h` (like Matter Labs do for instance) as a return type of at least public functions? They'll try to always return `CUDA_SUCCESS` for now but at least the signature will be (more) correct

icicle/appUtils/ntt/ntt.cu
```diff
@@ -0,0 +1,408 @@
...
+    generate_twiddle_factors(d_twiddles, n_twiddles, is_inverse ? S::omega_inv(logn) : S::omega(logn), config.stream);
+  }
+
+  E* d_input;
+  if (on_device) {
+    cudaMallocAsync(&d_input, input_size, config.stream);
+    cudaMemcpyAsync(d_input, input, input_size, cudaMemcpyHostToDevice, config.stream);
+  }
+
+  if ((is_inverse && (config.ordering == Ordering::kNN || config.ordering == Ordering::kNR)) ||
```
vhnatyk, line 367: umm - this feels a bit clunky compared to [other libs](https://github.com/supranational/sppark/blob/fffd7349b29bb086d2a8147b227baf8b9ed006dd/ntt/ntt.cuh#L91-L110) for example

vhnatyk (commented):
frankly still feels a bit raw, was writing a lengthy comment until it disappeared so huh... in short

1) maybe we can create a feature branch like feat/new_api - so I can start Rust API refactoring there since we are two days late now 

2) not quite comfy about ntt api yet - it's close though. Contrary to dedicated wrappers - api universal parameter doesn't seem a welcoming way of new integrations. We can add those later-on of course

3) like a safety net for potentially conflicting params needs test coverage  - otherwise seems a fruitful soil for some nasty bugs. Also while I'm sure we 100% need decimation and butterfly type params for integration convenience and compatibility - maybe it's better to add them later on or comment to not leave them floating unused

4) as there seem to be some uncertainties with LDE API - maybe we can keep it and file issues on a case-by-case basis instead of just removing it right away ;) . Definitely would make merging with the current Gnark Go API easier 

5) separation of cuh and cu files probably can be increased - ZKSync-style? kernels to separate files as well - later when per-file compilation is done - can reduce the scope of changes and compile time further

6) legacy end2end api needs to be removed  - we will have to replace it anyway with one0time/dynamic twidle generation

7)Halo2 integration seems can continue once function prototypes are finalized - no need to wait merging to main

icicle/CMakeLists.txt
```diff
@@ -18,31 +18,76 @@ set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --expt-relaxed-constexpr")
...
+
+  include(FetchContent)
+  FetchContent_Declare(
+    googletest
+    URL https://github.com/google/googletest/archive/refs/tags/v1.13.0.zip
+  )
+  # For Windows: Prevent overriding the parent project's compiler/linker settings
+
+  # boosting lib
+  include_directories("/home/miner/include/boost_1_80_0")
```
vhnatyk, line 72: I just noted a hardcoded path. Seems I missed it at #193

DmytroTym:
Closing as merging to main is not expected anytime soon and we decided to make https://github.com/ingonyama-zk/icicle/tree/feat/233-update-api the main branch for the API refactoring effort.

PR #237, opened 2023-10-03
linting docs

ImmanuelSegol:
## Describe the changes

This PR...

## Linked Issues

Resolves #

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,13 @@ docker build -t <name_of_your_choice> .
 docker run --gpus all -it <name_of_your_choice> /bin/bash
 ```
 
+## Code Formatting 
+
+### C++ and CUDA
+1. Install clang-format: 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -178,6 +178,23 @@ docker build -t <name_of_your_choice> .
 docker run --gpus all -it <name_of_your_choice> /bin/bash
 ```
 
+## Code Formatting
```
jeremyfelder, line 181: Since this is only needed when contributing/changing code and not necessary for using/running icicle, I think it makes sense to leave it in the [section below](https://github.com/ingonyama-zk/icicle#development-contributions).

If we want to be more clear about needing clang-format to be installed, it may make sense to point out that the script requires Rust and Go to be installed as well

jeremyfelder (changes requested):

PR #240, opened 2023-10-03
Cuda refactoring

DmytroTym:
1. Restores `lde` files with minimal functionality.
2. Removes `vector_manipulation` with methods partially moved to `lde` and partially - to `ntt`.
3. Creates a separate device context struct that can be passed to all methods to detail `device_id`, `stream`, `mempool` etc.
In general, I opted for the absolute minimum public API that in theory covers our immediate needs. This is because it's much easier to add new public methods than to remove existing ones so any questionable public functions should be avoided.


icicle/appUtils/ntt/ntt.cu
```diff
@@ -311,14 +333,14 @@ void ntt_inplace_batch_template(
...
-    if (is_coset) batch_vector_mult(coset, d_inout, n, batch_size, stream);
+    if (is_coset) batchVectorMult<E, S><<<num_blocks, num_threads, 0, stream>>>(coset, d_inout, n, batch_size);
 
     num_threads = max(min(n / 2, MAX_NUM_THREADS), 1);
     num_blocks = (n * batch_size + num_threads - 1) / num_threads;
     template_normalize_kernel<E, S>
       <<<num_blocks, num_threads, 0, stream>>>(d_inout, n * batch_size, S::inv_log_size(logn));
   } else {
-    if (is_coset) batch_vector_mult(coset, d_inout, n, batch_size, stream);
+    if (is_coset) batchVectorMult<E, S><<<num_blocks, num_threads, 0, stream>>>(coset, d_inout, n, batch_size);
```
vhnatyk, line 343: the sneak case is a better idea imho - is now sort of an implicit defacto standard for our code?

icicle/utils/error_handler.cuh
```diff
@@ -0,0 +1,32 @@
...
+{
+  cudaError_t err{cudaGetLastError()};
+  if (err != cudaSuccess) {
+    std::cerr << "CUDA Runtime Error at: " << file << ":" << line << std::endl;
+    std::cerr << cudaGetErrorString(err) << std::endl;
+  }
+}
+
+#define CHECK_SYNC_DEVICE_ERROR() syncDevice(__FILE__, __LINE__)
+void syncDevice(const char* const file, const int line)
```
vhnatyk, line 25: maybe let's stick with a snake case? compatible with Rust and ok for C as well.

icicle/CMakeLists.txt
```diff
@@ -44,20 +44,23 @@ if (NOT BUILD_TESTS)
...
     TARGET icicle
     POST_BUILD
-    COMMAND ${CMAKE_OBJCOPY} ARGS --redefine-sym msm_cuda=${CURVE}_msm_cuda --redefine-sym msm_internal_cuda=${CURVE}_msm_internal_cuda ${PROJECT_BINARY_DIR}/CMakeFiles/icicle.dir/appUtils/msm/msm.cu.o
-  )
-  add_custom_command(
-    TARGET icicle
-    POST_BUILD
-    COMMAND ${CMAKE_AR} ARGS -rcs ${LIBRARY_OUTPUT_DIRECTORY}/libingo_${CURVE}.a ${PROJECT_BINARY_DIR}/CMakeFiles/icicle.dir/appUtils/msm/msm.cu.o
+    COMMAND ${CMAKE_OBJCOPY} ARGS --redefine-sym msm_cuda=${CURVE}_msm_cuda ${PROJECT_BINARY_DIR}/CMakeFiles/icicle.dir/appUtils/msm/msm.cu.o
+    COMMAND ${CMAKE_OBJCOPY} ARGS --prefix-symbols=${CURVE}_ ${PROJECT_BINARY_DIR}/CMakeFiles/icicle.dir/appUtils/ntt/ntt.cu.o
```
vhnatyk, line 55: btw - keeping o files in same dir as sources isn't cool cause Rust checks if the dir was changed and rebuilds. We can fine-tune the building later on, but maybe better to change to some dedicated temporary folder right away

vhnatyk (approved):
Generally looks good - didn't review it super thoroughly as it's a development branch and we need to move fast. Anyway, this will be re-reviewed later on.

icicle/CMakeLists.txt
vhnatyk: also `--prefix-symbols=` probably won't work at some point - as it prefixes all symbols - it also brakes some external imports from dynamic libraries - so it fails linking. And for sure it brakes global consts etc. As I tried - so it didn't work on linking the full library - hmm maybe it works ok on individual .cu files until global exports get into library. Instead, objcopy has a file argument and a command to modify only functions listed in that file. We will have to parse all our headers and add `extern` methods to that file. Of course we can add entries there manually - but probably manual editing is not a great idea overall.

icicle/appUtils/ntt/ntt.cu
jeremyfelder: ➕ Lets be consistent across the codebase, all snake case (our current) or all camel case (Google C++ style, CUDA Runtime style)

DmytroTym: Yea I wanted to make functions snake-case on CUDA side, as per google C++ style guide. Is everyone OK with that @jeremyfelder @ImmanuelSegol @ChickenLover

jeremyfelder: I'm fine either way but 

>Yea I wanted to make functions snake-case on CUDA side, as per google C++ style guide

I think google c++ style guide says function names are UpperCamelCase: https://google.github.io/styleguide/cppguide.html#Function_Names

If we want to have Google c++ style as a base we can add this to the clang-format configuration file: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#basedonstyle

PR #254, opened 2023-10-28
Two curve NTT correctness issue hotfix

DmytroTym:
This PR resolves the problem occurring in gnark groth16 circuits that use more than one curve simultaneously. More specifically, verification fails due to `krs2` MSM being incorrect, in turn caused by `computeH` function returning incorrect result. On closer inspection it turned out that `template_normalize_kernel` [here](https://github.com/ingonyama-zk/icicle/blob/main/icicle/appUtils/ntt/ntt.cuh#L333) just doesn't launch. It does so silently, without throwing any errors or visible memory leaks. It does seem that the problem has to do with compilation because simply using the same kernel with a different name helps as well as only using a single curve. The "fix" in this PR just replaces a second use-case of `template_normalize_kernel` with a more correct Montgomery treatment cherry-picked from our new API branch. I will try to get more info about this error in the following days while for now this fix should suffice.
PS: guess it's one more point for unified compilation for different wrapper languages - hopefully we'll have fewer strange hard-to-debug C++ compilation bugs that way lol...

### DIFF
--- a/icicle/primitives/field.cuh
+++ b/icicle/primitives/field.cuh
@@ -65,10 +65,6 @@ public:
 
   static constexpr HOST_DEVICE_INLINE Field modulus() { return Field{CONFIG::modulus}; }
 
-  static constexpr HOST_DEVICE_INLINE Field montgomery_r() { return Field{CONFIG::montgomery_r}; }
-
-  static constexpr HOST_DEVICE_INLINE Field montgomery_r_inv() { return Field{CONFIG::montgomery_r_inv}; }
-
   // private:
   typedef storage<TLC> ff_storage;
   typedef storage<2 * TLC> ff_wide_storage;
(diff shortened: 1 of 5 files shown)


icicle/primitives/field.cuh
```diff
@@ -714,7 +710,7 @@ public:
     hex_string << std::hex << std::setfill('0');
 
     for (int i = 0; i < TLC; i++) {
-      hex_string << std::setw(8) << xs.limbs_storage.limbs[i];
+      hex_string << std::setw(8) << xs.limbs_storage.limbs[TLC - i - 1];
```
ImmanuelSegol, line 713: just wondering why this change?

ImmanuelSegol (approved):
The fix makes sense to me.

Was just wondering if it would be possible perhaps for you to add some comments explaining this in the code.

jeremyfelder (dismissed):

DmytroTym:
@ImmanuelSegol yea I thought about commenting it somewhere but not sure where tbh. And I'm still not 100% sure when this happens too...
I will take some time in several days to try to see when the issue reproduces but for now I'm not really sure

ImmanuelSegol (approved):

jeremyfelder (approved):

PR #259, opened 2023-11-02
Feat/233 update api stas test

svpolonsky:
## Describe the changes

This PR adds CUDA error checking to MSM

## Linked Issues

Resolves # This is my first experience with PR, so I'd appreciate some help, please!

### DIFF
--- a/icicle/appUtils/msm/msm.cu
+++ b/icicle/appUtils/msm/msm.cu
@@ -386,6 +386,7 @@ void bucket_method_msm(
   unsigned NUM_THREADS = 1 << 10;
   unsigned NUM_BLOCKS = (nof_buckets + NUM_THREADS - 1) / NUM_THREADS;
   initialize_buckets_kernel<<<NUM_BLOCKS, NUM_THREADS, 0, stream>>>(buckets, nof_buckets);
+  CHECK_LAST_CUDA_ERROR();
 
   // accumulate ones
   P* ones_results; // fix whole division, in last run in kernel too
@@ -395,12 +396,14 @@ void bucket_method_msm(
   NUM_THREADS = min(1 << 8, nof_runs);
   NUM_BLOCKS = (nof_runs + NUM_THREADS - 1) / NUM_THREADS;
   add_ones_kernel<<<NUM_BLOCKS, NUM_THREADS, 0, stream>>>(d_points, d_scalars, ones_results, size, run_length);
+  CHECK_LAST_CUDA_ERROR();
 
   for (int s = nof_runs >> 1; s > 0; s >>= 1) {
     NUM_THREADS = min(MAX_TH, s);
     NUM_BLOCKS = (s + NUM_THREADS - 1) / NUM_THREADS;
     single_stage_multi_reduction_kernel<<<NUM_BLOCKS, NUM_THREADS, 0, stream>>>(
       ones_results, ones_results, s * 2, 0, 0, 0, s);
+    CHECK_LAST_CUDA_ERROR();  
   }
 
   unsigned* bucket_indices;
@@ -414,6 +417,7 @@ void bucket_method_msm(
   split_scalars_kernel<<<NUM_BLOCKS, NUM_THREADS, 0, stream>>>(
     bucket_indices + size, point_indices + size, d_scalars, size, msm_log_size, nof_bms, bm_bitsize, c);
   //+size - leaving the first bm free for the out of place sort later
+  CHECK_LAST_CUDA_ERROR();
 
   // sort indices - the indices are sorted from smallest to largest in order to group together the points that belong to
   // each bucket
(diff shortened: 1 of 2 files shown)


icicle/appUtils/msm/msm.cu
```diff
@@ -395,12 +396,14 @@ void bucket_method_msm(
   NUM_THREADS = min(1 << 8, nof_runs);
   NUM_BLOCKS = (nof_runs + NUM_THREADS - 1) / NUM_THREADS;
   add_ones_kernel<<<NUM_BLOCKS, NUM_THREADS, 0, stream>>>(d_points, d_scalars, ones_results, size, run_length);
+  CHECK_CUDA_ERROR();
```
ImmanuelSegol, line 399: `CHECK_CUDA_ERROR` should be used to wrap CUDA methods

`CHECK_CUDA_ERROR(<cuda_method>)`

I think u want to use 

`CHECK_LAST_CUDA_ERROR()`

This mehod calls `cudaGetLastError` [docs here](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__ERROR.html#group__CUDART__ERROR_1g3529f94cb530a83a76613616782bd233)

ImmanuelSegol (changes requested):

icicle/appUtils/msm/msm.cu
ImmanuelSegol: CHECK_SYNC_DEVICE_ERROR on the other hand is blocking and will wait for a computation to complete before returning error.

svpolonsky: I followed the best practice suggestion from [here](https://leimao.github.io/blog/Proper-CUDA-Error-Checking/):

**Check CUDA asynchronous error in development by synchronization and error checking after kernel launch calls and disable it in production.**

I have a `#DEFINE ERROR_MODE` in `error_handler.cuh` that selects either TEST mode (development) or PRODUCTION:

1. In TEST mode we want to localize errors even at expense of the speed. This is why I'd like to use CHECK_SYNC_DEVICE_ERROR.

In PRODUCTION mode we prioritize the speed over error localization, so  CHECK_LAST_CUDA_ERROR

Does this make sense? If yes, we can figure out how exactly we select between the TEST and PRODUCTION.

svpolonsky: There is another point, that I'd like put into separate comment. CUDA (and the article above) differentiate between recoverable (i.e. not enough memory) and non-recoverable (null pointer).

Do we want to take care of such errors? Personally, I'd just kill the process. K8s would take care of resilience by detecting the kill and restarting the app. For mining this shouldn't be a problem. Not sure about other apps.

ImmanuelSegol: @svpolonsky what u did is ok

Im just saying ur using the wrong method, if you look at [CHECK_CUDA_ERROR](https://github.com/ingonyama-zk/icicle/blob/2387086f4376372eb4f86b683697e094773b2d7d/icicle/utils/error_handler.cuh#L4) it requires a parameter input.

ImmanuelSegol: @svpolonsky use CHECK_LAST_CUDA_ERROR

icicle/appUtils/msm/msm.cu
```diff
@@ -386,6 +386,7 @@ void bucket_method_msm(
   unsigned NUM_THREADS = 1 << 10;
   unsigned NUM_BLOCKS = (nof_buckets + NUM_THREADS - 1) / NUM_THREADS;
   initialize_buckets_kernel<<<NUM_BLOCKS, NUM_THREADS, 0, stream>>>(buckets, nof_buckets);
+  CHECK_CUDA_ERROR();
```
ImmanuelSegol, line 389:
```suggestion
  CHECK_LAST_CUDA_ERROR();
```

it should be like this @svpolonsky

ImmanuelSegol (changes requested):

icicle/appUtils/msm/msm.cu
jeremyfelder: @ImmanuelSegol Stas is trying to use the defined names that were added [here](https://github.com/ingonyama-zk/icicle/pull/259/files#diff-4e31376046899777077b6965bf464226629e39527466bd8b85edf164988710e5R45-R48) so calling it depends on whether it is TEST or PRODUCTION.

Though @svpolonsky, I agree with Immanuel that shadowing the already defined [CHECK_CUDA_ERROR](https://github.com/ingonyama-zk/icicle/blob/2387086f4376372eb4f86b683697e094773b2d7d/icicle/utils/error_handler.cuh#L4) might not be the best. Can we change the TEST/PROD name to something else?

jeremyfelder:
@svpolonsky Any reason the base branch is `feat/233-update-api` and not `main`?

DmytroTym (changes requested):
First of all we should be checking errors from all CUDA API calls (like `cudaMalloc`, `cudaMemcpy` etc) and `cub` method calls, not just from kernel launches.
I personally prefer wrapping calls with macros rather than just doing stand-alone error checks, like matter-labs do, e.g. https://github.com/matter-labs/era-bellman-cuda/blob/main/src/msm.cu
1. It feels more logical to me because each error check is tied to a specific call which it checks for synchronous errors
2. With all the error checks for CUDA API calls, the code will end up seriously bloated with stand-alone checks

jeremyfelder:
Closing in favor of #317

PR #262, opened 2023-11-11
Update Rust apis

DmytroTym:
This PR aims to update Rust APIs to the minimal usable standard


icicle/appUtils/lde/lde.cuh
```diff
@@ -6,61 +6,57 @@
 
 /**
  * @namespace lde
- * LDE (stands for low degree extension) contains [NTT](@ref ntt)-based methods for translating between coefficient and
- * evaluation domains of polynomials. It also contains methods for element-wise manipulation of vectors, which is useful
- * for working with polynomials in evaluation domain.
+ * LDE (stands for low degree extension) contains [NTT](@ref ntt)-based methods for translating between coefficient and evaluation domains of polynomials.
+ * It also contains methods for element-wise manipulation of vectors, which is useful for working with polynomials in evaluation domain.
```
ChickenLover, line 10: But they are just general-purpose arithmetic. Wouldn't it be better to move these into some other file?

icicle/appUtils/lde/lde.cuh
```diff
@@ -6,61 +6,57 @@
 
 /**
  * @namespace lde
- * LDE (stands for low degree extension) contains [NTT](@ref ntt)-based methods for translating between coefficient and
- * evaluation domains of polynomials. It also contains methods for element-wise manipulation of vectors, which is useful
- * for working with polynomials in evaluation domain.
+ * LDE (stands for low degree extension) contains [NTT](@ref ntt)-based methods for translating between coefficient and evaluation domains of polynomials.
```
ChickenLover, line 9: isn't LDE a part of NTT and not vice versa?

icicle/appUtils/lde/lde.cuh
DmytroTym: I think you're right...
My initial logic was like this: in arkworks there [is](https://github.com/arkworks-rs/algebra/blob/master/poly/src/evaluations/univariate/mod.rs) this `Evaluations` struct which implements all the operations one might perform on polynomials in evaluation form, including element-wise add, sub and mult and I wanted to do something similar in lde. But we might also need vector operations elsewhere (e.g. in tree builders right?) so it's best to factor these out into a more general place
I think I'm going to remove lde for now then, and will (try to) implement a higher-level polynomial functionality in the next PR. Would that make sense?

icicle/appUtils/lde/lde.cuh
DmytroTym: Partially answered in the previous comment: I imagine future LDE as something like arkworks' [`Evaluation`](https://github.com/arkworks-rs/algebra/blob/master/poly/src/evaluations/univariate/mod.rs) or a mix of `Evaluation` and [`DensePolynomial`](https://github.com/arkworks-rs/algebra/blob/master/poly/src/polynomial/univariate/dense.rs). So, it's a higher-level wrapper over NTT. For example, to evaluate a polynomial of degree `N` on a domain of size `2N` you need 2 size `N` NTTs, or to multiply two polynomials you need some NTTs for translating between evaluation and coefficient forms

DmytroTym:
@vhnatyk a couple of questions I have about the current NTT design in this branch:
1. I don't understand the purpose of [`Domain`](https://github.com/ingonyama-zk/icicle/blob/develop/dima/feat-233-update-rust-apis/wrappers/rust/icicle-curves/icicle-bn254/src/ntt/domain.rs#L14) currently: it's just a wrapper around `Config`. Arkworks-style [domain](https://github.com/arkworks-rs/algebra/blob/master/poly/src/domain/radix2/mod.rs) is (roughly) size, twiddles and coset. So no `inout` or `batch_size`. I think for us `Domain` might be a useful abstraction too, but IMO we should either add it later on after this PR or spend some time designing it.
2. Here are my other general preferences though I'm not insisting and will accept whatever consensus the team develops:
- I would take `inout`, `size` and `is_inverse` out from config and put these fields into the NTT function separately. My intuition is that fields that have a natural default should be inside cfg but something that you almost definitely need to set manually shouldn't. Also imo `ntt(data, true, cfg)` is more readable than `ntt(cfg)`.
- Maybe it's better to have just one `on_device` flag instead of two (`inputs_on_device` and `outputs_on_device`) in `NTTConfig`? It just feels safer and more intuitive to me to not change where the pointer `inout` is pointing to (which will inevitably happen if inputs are on device and outputs on host let's say) in the function which is supposedly "in-place". From my experience, it's almost always either both inputs and outputs on device, or both on host in real-life NTTs
- I completely agree that we need convenience wrappers for NTT functions, just thought maybe we could leave a minimal number of expressive calls for now and later on grow the wrappers organically. I just don't really believe in our ability to thoughtfully design good quality wrappers right now without delaying the PR.
@jeremyfelder @ChickenLover please feel free to comment

icicle/appUtils/msm/msm.cu
```diff
@@ -27,6 +28,12 @@ namespace msm {
     // #define BIG_TRIANGLE
     // #define SSM_SUM  //WIP
 
+    template <typename S>
+    int get_optimal_c(int bitsize)
+    {
+      return ceil(log2(bitsize)) - 4;
```
jeremyfelder, line 34: Should this be able to be negative?

icicle/appUtils/ntt/ntt.cu
```diff
@@ -349,30 +346,44 @@ namespace ntt {
...
-  cudaError_t NTT(E* input, int size, bool is_inverse, NTTConfig<S> config)
+  cudaError_t NTT(NTTConfig<E, S>* config)
   {
-    uint32_t logn = uint32_t(log(size) / log(2));
-    uint32_t n_twiddles = size; // n_twiddles is set to 4096 as BLS12_381::scalar_t::omega() is of that order.
-    size_t input_size = size * config.batch_size * sizeof(E);
-    bool is_on_device = config.are_inputs_on_device;
-    bool generate_twiddles = (config.twiddles == nullptr);
-    cudaStream_t stream = config.ctx.stream;
+    CHECK_LAST_CUDA_ERROR();
```
jeremyfelder, line 351: Is this just for debugging now?
If not, `cudaGetLastError` resets to the latest error to `cudaSuccess`, should we use `cudaPeekAtLastError`?

icicle/CMakeLists.txt
jeremyfelder, line 1: Do we want to add defining G2 here or will we have users pass it via `-DG2_DEFINED`?

icicle/appUtils/msm/msm.cu
```diff
@@ -132,32 +139,14 @@ namespace msm {
...
-      }
-      const unsigned start_index = tid * run_length;
-      P sum = P::zero();
-      for (int i = start_index; i < min(start_index + run_length, msm_size); i++) {
-        if (scalars[i] == S::one()) sum = sum + points[i];
-      }
-      results[tid] = sum;
-    }
-
+    template <typename S>
```
jeremyfelder, line 149: Do we want to use force_inline instead of unused typename params?
Same for `find_max_size` below

icicle/appUtils/ntt/ntt.cu
```diff
@@ -61,7 +58,7 @@ namespace ntt {
       int number_of_threads = MAX_THREADS_BATCH;
       int number_of_blocks = (n * batch_size + number_of_threads - 1) / number_of_threads;
       reverse_order_kernel<<<number_of_blocks, number_of_threads, 0, stream>>>(arr, arr_reversed, n, logn, batch_size);
-      cudaMemcpyAsync(arr, arr_reversed, n * batch_size * sizeof(E), cudaMemcpyDeviceToDevice, stream);
+      cudaMemcpyAsync(arr, arr_reversed, n * batch_size * sizeof(E), cudaMemcpyDefault, stream);
```
jeremyfelder, line 61: cudaMemcpyDefault is only on devices that support unified virtual addressing...what devices are these? We should update docs for supported devices

wrappers/rust/icicle-curves/icicle-bn254/build.rs
```diff
@@ -0,0 +1,30 @@
...
+    let cargo_dir = var("CARGO_MANIFEST_DIR").unwrap();
+    let profile = var("PROFILE").unwrap();
+
+    let target_output_dir = format!("{}/../../target/{}", cargo_dir, profile);
+
+    Config::new("./icicle")
+                .define("BUILD_TESTS", "OFF") //TODO: feature
+                // .define("CURVE", "bls12_381")
+                .define("CURVE", "bn254")
+                // .define("ECNTT_DEFINED", "") //TODO: feature
```
jeremyfelder, line 17:
```suggestion
                // .define("ECNTT_DEFINED", "") //TODO: feature
                // .define("G2_DEFINED", "")
```

icicle/appUtils/msm/msm.cu
```diff
@@ -907,26 +904,59 @@ namespace msm {
...
+        config.are_scalars_montgomery_form, config.are_points_on_device, config.are_points_montgomery_form,
+        config.are_results_on_device, config.is_big_triangle, config.large_bucket_factor, config.is_async,
+        config.ctx.stream);
     else
       batched_bucket_method_msm(
-        config.bitsize, config.c, scalars, points, config.batch_size, msm_size, results, config.are_scalars_on_device,
-        config.ctx.stream);
+        bitsize, (config.c == 0) ? get_optimal_c<S>(bitsize) : config.c, scalars, points, config.batch_size, msm_size,
+        results, config.are_scalars_on_device, config.ctx.stream);
     return cudaSuccess;
```
jeremyfelder, line 949: I don't see us using `config->point_size` even though the comments on MSMConfig say it is used instead of msm_size in certain cases. This parameter in the config is a bit confusing, can't we always just use msm_size and batch_size to get the correct values?

jeremyfelder (commented):
> 1. I don't understand the purpose of [`Domain`](https://github.com/ingonyama-zk/icicle/blob/develop/dima/feat-233-update-rust-apis/wrappers/rust/icicle-curves/icicle-bn254/src/ntt/domain.rs#L14) ......

I agree

> 2. Here are my other general preferences....:
> 
> * I would take `inout`, `size` and `is_inverse` out from config and put these fields into the NTT function separately. My intuition is that fields that have a natural default should be inside cfg but something that you almost definitely need to set manually shouldn't. Also imo `ntt(data, true, cfg)` is more readable than `ntt(cfg)`.

I agree about `inout` and `size`. I'm not sure why `is_inverse` would be different than `is_coset` though?

> * Maybe it's better to have just one `on_device` flag instead of two (`inputs_on_device` and `outputs_on_device`) in `NTTConfig`? It just feels safer and more intuitive to me to not change where the pointer `inout` is pointing to (which will inevitably happen if inputs are on device and outputs on host let's say) in the function which is supposedly "in-place". From my experience, it's almost always either both inputs and outputs on device, or both on host in real-life NTTs

I agree that it will be less error prone but this will force a user to manually move the initial data and resulting data to/from device if they want to use `on_device=true` which might also be error prone.

> * I completely agree that we need convenience wrappers for NTT functions, just thought maybe we could leave a minimal number of expressive calls for now and later on grow the wrappers organically. I just don't really believe in our ability to thoughtfully design good quality wrappers right now without delaying the PR.

👍🏻 

> * We currently have three similar fields in `NTTConfig`: `Ordering`, `Decimation` and `Butterfly`. They are not independent, for example `Decimation::kDIT` and `Butterfly::kCooleyTukey` are basically the same thing. I think there's an issue in case for example `Decimation::kDIF` and `Butterfly::kCooleyTukey` are passed, in which case one has to be ignored. The reason why we have `Decimation` and `Butterfly` in the first place is for compatibility with other codebases (e.g. gnark uses DIT and DIF in their [codebase](https://github.com/celer-network/gnark/blob/master/internal/backend/bn254/groth16/prove.go#L337)) but imo it'd best to just document which decimation/butterfly options correspond to which `Ordering` options and remove `Decimation` and `Butterfly` fields altogether.

Yes, if multiple options are essentially the same outcome but different names, we should give one option and document the alternative names for that option.

icicle/appUtils/msm/msm.cu
DmytroTym: You're right, it's currently unused, I tried to mark this fact with the following comment: https://github.com/ingonyama-zk/icicle/pull/262/files#diff-cdd3b67200a359e20797b18b99ca7176f36d4c9d48d9843d22500c5424fe1c7dR106. The idea of this parameter is to control whether you want to re-use the same set of bases in all MSMs or have custom bases for each one. Current default is the latter, although it's really easy to make this parameter functional and from my understanding this will help you in halo2 where you could reduce memory footprint from bases by sharing bases between MSMs, right?

icicle/appUtils/ntt/ntt.cu
DmytroTym: @vhnatyk I also don't understand the purpose of this tbh

icicle/appUtils/ntt/ntt.cu
DmytroTym: We don't really have proper error checking unfortunately, I guess this was just for debugging here
I would consider error checking out of scope for this PR tbh

PR #263, opened 2023-11-12
fix typo ntt.rs

tudorpintea999:
fix typo


DmytroTym (changes requested):
Looking at https://en.wikipedia.org/wiki/Cooley–Tukey_FFT_algorithm, I guess "in-place" is correct

tudorpintea999:
hello, yes you re right. I tought it s just a logger 🤝

PR #289, opened 2023-11-29
Improved modulo multiplier

DmytroTym:
I tried to document inner workings of methods that were changed in `field.cuh`, but the bottom line is: Karatsuba implemented for wide multiplication; lsb and msb multipliers optimised as much as possible plus some extra tweaks.

### Benchmarks

All the benchmarks are from the same RTX 3090Ti machine.

Multiplication modulo base modulus of BLS12-377, throughput, 10^9 ops/s:

- Current `main` - 14.1
- This PR - 18.8
- Matter Labs' [z-prize submission](https://github.com/matter-labs/z-prize-msm-gpu/blob/main/bellman-cuda-rust/bellman-cuda-sys/native/ff_dispatch_st.cuh) - 18.1

Multiplication modulo base modulus of BN254, throughput, 10^9 ops/s:

- Current `main` - 29.6
- This PR - 38.9
- Matter Labs' [bellman-cuda](https://github.com/matter-labs/era-bellman-cuda/blob/main/src/ff_dispatch_st.cuh) - 39.5

Among some higher-level applications, we have a Poseidon tree based on Supranational's [PC2 implementation](https://github.com/supranational/pc2_cuda/blob/main/src/tree_builder_device.cu). The time for building depth 30 tree for BLS12-381 in seconds:

- Current `main` - 17.2
- This PR - 14.2
- Supranational's original version - 13.6

Despite being on-par in terms of multiplier, our tree is still a bit slower, probably due to squaring (sppark has a separate implementation and we just use multiplication for now).
Next, MSM of size 2^22 on BN254 curve (without data transfer), ms.:

- Current `main` - 88.8
- This PR - 70.6
- Matter Labs' [bellman-cuda](https://github.com/matter-labs/era-bellman-cuda/blob/main/src/msm.cu) - 44.1

UPD: one important reason for our slowdown compared to Matter Labs and limited acceleration compared to `main` is that 256-bit adder that we're using is not accelerated as much as expected. Here are the throughputs of different versions of projective BN254 adder:

- Current `main` - 2.49
- This PR - 2.91
- Matter Labs' [bellman-cuda](https://github.com/matter-labs/era-bellman-cuda/blob/main/src/ec.cuh) - 3.37

Meanwhile, 384-bit adder is accelerated as expected. E.g. for projective addition on BLS12-377:

- Current `main` - 1.11
- This PR - 1.56
- Matter Labs' [z-prize submission](https://github.com/matter-labs/z-prize-msm-gpu/blob/main/bellman-cuda-rust/bellman-cuda-sys/native/ec.cuh) - 1.5

It seems that the reason is that EC addition is basically register bound, and one advantage of Montgomery multiplication implemented by Matter Labs and Supranational is the smaller register footprint. This might not affect the performance at all, like with 384-bit adder but does affect 256-bit one, at least on 3090Ti. Solving this issue is future work.

@vhnatyk if you have NTT benchmarks, would be interesting to see them too.


jeremyfelder (approved):

vhnatyk (approved):
tested performance and correctness for bls12_381 - in my custom tests that is simple multiply in a for loop for different grid sizes on laptop 3050ti - I get ~9000 mult / microsecond = 9Gops versus 7800 in Sppark. That is decent advantage over current main ~7Gops  👍

PR #293, opened 2023-12-12
New API

jeremyfelder:
## Describe the changes

This PR revamps the API


DmytroTym (approved):
I think we're going to have another PR with breaking changes soon so this is not a stable version, but it's a stepping stone to more final interfaces. See https://github.com/ingonyama-zk/icicle/pull/262 and https://github.com/ingonyama-zk/icicle/pull/292 for the discussion of changes in this PR

PR #299, opened 2023-12-14
Feat/roman/msm ntt generics

ChickenLover:
## Describe the changes

This PR adds generics for curve definitions and MSM / NTT methods


wrappers/rust/icicle-core/src/curve.rs
```diff
@@ -12,6 +12,11 @@ pub trait CurveConfig: PartialEq + Copy + Clone {
     fn eq_proj(point1: *const c_void, point2: *const c_void) -> c_uint;
```
DmytroTym, line 12: Since you introduced associated types to the trait (which is probably a good idea), `c_void` can be changed to `Projective` I think

wrappers/rust/icicle-core/src/curve.rs
```diff
@@ -191,3 +196,161 @@ where
...
+        pub(crate) fn generate_random_affine_points(size: usize) -> Vec<G1Affine> {
+            let mut res = vec![G1Affine::zero(); size];
+            unsafe { GenerateAffinePoints(&mut res[..] as *mut _ as *mut G1Affine, size) };
+            res
+        }
+    };
+}
+
+#[macro_export]
+macro_rules! impl_curve_tests {
```
DmytroTym, line 285: A general note about all tests: earlier we discussed that it'd be nice to have all tests implemented in generic fashion outside of macros. I see at least one issue with this in functions `generate_random_scalars` and `generate_random_*_points` which are themselves implemented inside a macro. Is this the only blocker or is there anything else?

wrappers/rust/icicle-core/src/curve.rs
```diff
@@ -12,6 +12,11 @@ pub trait CurveConfig: PartialEq + Copy + Clone {
     fn eq_proj(point1: *const c_void, point2: *const c_void) -> c_uint;
     fn to_affine(point: *const c_void, point_aff: *mut c_void);
 
+    type BaseField;
```
DmytroTym, line 15: Wondering if it's possible to constrain these associated types, after all we know that for example `Projective = Projective<BaseField, Self>`

wrappers/rust/icicle-curves/icicle-bw6-761/src/curve.rs
```diff
@@ -0,0 +1,22 @@
+#[cfg(feature = "arkworks")]
+use ark_bw6_761::{g1::Config as ArkG1Config, Fq, Fr};
+use icicle_core::curve::{Affine, CurveConfig, Projective};
+use icicle_core::field::{Field, FieldConfig};
+use std::ffi::{c_uint, c_void};
+
+impl_curve!(12, 24,);
```
DmytroTym, line 7: For BW6-761 curve it might be better to import BLS12-377 base field instead of default macro implementation of the scalar field (see arkworks [version](https://github.com/arkworks-rs/algebra/blob/master/curves/bw6_761/src/fields/fr.rs)). I imagine that any codebase that uses this curve will benefit from compile-time knowledge that the two fields are the same. I think this also warrants splitting `imps_curve!` macro into two: `impl_field!` and `impl_curve!` which I think is a good idea on its own: a macro for implementing stand-alone fields will be useful for STARK fields. Also, consider how G2 works in BW6-761: it's a curve over the same base field as G1 (but a different curve nonetheless). I think to accommodate this it's nice to have a separate macro for fields. Also maybe one idea for macros would be not relying on names available in the scope but passing them as parameters (though I guess it's a tradeoff and you can argue that having concrete names inside a macro is more readable).

wrappers/rust/icicle-core/src/curve.rs
ChickenLover: I tried that initially. But this is an unstable feature

DmytroTym: Right. Well, perhaps you could've just used `Projective<Self::BaseField, Self>` everywhere instead of `Self::Projective` though that's a bit wordy

wrappers/rust/icicle-core/src/ntt/mod.rs
```diff
@@ -65,3 +67,210 @@ pub struct NTTConfig<'a, S> {
 //      *  generated by `S::one()` is used). */
 //     pub coset_table: Option<&'a [T]>,
 // }
+
+pub trait NTT<C: CurveConfig> {
```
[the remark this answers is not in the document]
ChickenLover: solved in ff2bf3e76ffeea86ec60824d2c5cdce3757d0f8f

wrappers/rust/icicle-core/src/curve.rs
ChickenLover: solved in ff2bf3e76ffeea86ec60824d2c5cdce3757d0f8f

wrappers/rust/icicle-curves/icicle-bw6-761/src/curve.rs
ChickenLover: Separated field macros from curve macros in 98675f766e432b373cf024a04cf5bb6fd32befa4

ChickenLover: Implementing bw6-761 with bls12-377 base field seems impossible in the current design due to this line in NTT macro
`impl NTT<$field> for $field_config {`

The base field config for bls12-377 is defined in another crate, so It's impossible to define new trait implementation here

DmytroTym: > The base field config for bls12-377 is defined in another crate, so It's impossible to define new trait implementation here

Yes sure, we should implement NTT on the base field of bls12-377 in bls12-377 crate IMO

DmytroTym: Oh wait, I don't think the second option works. After all, bls12-377 create will be compiled first so linker won't be able to find the extern NTT function which only gets created before bw6-761 crate is built...

wrappers/rust/icicle-core/src/curve.rs
DmytroTym: Let's leave it as TODO for this PR maybe?

wrappers/rust/icicle-core/src/curve.rs
```diff
@@ -1,142 +1,112 @@
...
 
 pub trait CurveConfig: PartialEq + Copy + Clone {
-    fn eq_proj(point1: *const c_void, point2: *const c_void) -> c_uint;
-    fn to_affine(point: *const c_void, point_aff: *mut c_void);
-
-    type BaseField;
-    type ScalarField;
+    type BaseField: FieldImpl;
+    type ScalarField: FieldImpl;
     type Projective;
```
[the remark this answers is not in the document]
ChickenLover: Removed in c2e4b46eaad32087a082d0ea606c539453ef547c

wrappers/rust/icicle-curves/icicle-bw6-761/src/curve.rs
ChickenLover: Moved this to an issue #307

DmytroTym (approved):
lgtm

PR #305, opened 2023-12-18
Rust NTT updates

DmytroTym:
This PR aims to update NTT interfaces to a state where they can be conveniently used from C++ and Rust without breaking changes in the near future, including when we ship our improved NTT algorithm.


DmytroTym:
Left to do:

1. Merge https://github.com/ingonyama-zk/icicle/pull/299 into dev and merge it into the current branch
2. Add some more Rust NTT tests, in particular with different device-related options like on-device memory and async execution using non-blocking streams

DmytroTym:
An overview of this PR:

1. On the CUDA side, it's mostly fixes for issues found during testing and in our latest integration, including many alignment errors. Also, enabling some important functionality in NTT and MSM. On the NTT side, cosets are now fully functional, and in MSM, batch/non-batch functionality is unified and `cfg.points_size` enabled.
2. In Rust, NTT and MSM tests are added and some refactoring done. I used `#[doc(hidden)]` a lot to denote functionality of `icicle-core` that needs to be public in order for curves to be implemented, but doesn't need to be exposed to normal users of icicle. Some of this functionality comes from complications with writing idiomatic generic code that needs to rely on inherently non-generic extern functions. Another example is random scalar/point generation which needs to be public for tests and benchmarks but probably won't be used by most integrations (also I think that it needs to accept an rng parameter so needs a breaking change in its current form so I don't want to make it fully public API). If you have better ideas for not-so-public functionality pls let me know.
3. For the end user on the Rust side, the intended usage is as follows: import the needed curve that implements our master-trait [SNARKCurve](https://github.com/ingonyama-zk/icicle/pull/305/files#diff-77a3b398a7c011998cb468384de6fd4bca78dde246ce04afaa84e9eabb88389fR11) which should provide everything an integration needs. The functions like `msm`, `ntt` and `get_default_*_config` are generic free-floating functions. So, something like:
```
from icicle_bn254 import CurveCfg;
from icicle_core import {msm::{msm, get_default_msm_config}, ntt::{ntt, get_default_ntt_config}};
....
ntt::<CurveCfg::ScalarField>(....);
msm::<CurveCfg>(....);
....
```
It's the simplest API I could come up with. At the same time, we have a more flexible system of traits and types that will hopefully allow us to implement sumcheck-oriented protocols or SNARKs, or maybe build some higher-level abstractions like polynomial operations with minimal changes (hopefully).

jeremyfelder (approved):

wrappers/rust/icicle-core/src/traits.rs
```diff
@@ -25,14 +33,14 @@ pub trait ArkConvertible {
     fn from_ark(ark: Self::ArkEquivalent) -> Self;
 }
 
-pub trait ResultWrap<T, TError>{
-    fn wrap(self) -> Result<T, TError>;
-    fn wrap_value(self, value: T) -> Result<T, TError>;
-    fn wrap_maybe_uninit(self, value: MaybeUninit<T>) -> Result<T, TError>;
-}
+// pub trait ResultWrap<T, TError>{
```
vhnatyk, line 36: to be removed

README.md
```diff
@@ -43,151 +43,32 @@ ICICLE is a CUDA implementation of general functions widely used in ZKP. ICICLE
...
-- follow [these instructions](https://github.com/ingonyama-zk/icicle/tree/main/icicle#prerequisites-on-ubuntu)
-- Any Nvidia GPU
+- [CUDA Toolkit](https://developer.nvidia.com/cuda-downloads) version 12.0 or newer.
+- [CMake]((https://cmake.org/files/)), version 3.18 and above. Latest version is recommended.
+- [GCC](https://gcc.gnu.org/install/download.html) version 9, latest version is recommended.
+- Any Nvidia GPU (which supports CUDA Toolkit version 12.0 or above).
 
-If you don't have access to a Nvidia GPU check out [google-colab](#google-colab). If you require more compute power and are looking to build or do research with ICICLE refer to our [grant program][GRANT_PROGRAM].
+> [!NOTE] 
+> It is possible to use CUDA 11 for cards which dont support CUDA 12, however we dont officially support this version and in the future there may be issues.
```
vhnatyk, line 52: maybe change to "_It is possible to use CUDA 11.0+, however..._"

vhnatyk:
some comments imo worth addressing in main

PR #317, opened 2023-12-21
213 error handling

### LINKED ISSUE #213: [FEAT] Implement error codes

vhnatyk:
## Describe the changes

draft of error handling on C side

the template would be to 

- start every function with `CHECK_LAST_CUDA_ERROR();` to detect prior and reset non-sticky errors.
- Then to wrap every API call returning cudaError_t in `CHECK_CUDA_ERROR(..)` to detect errors asap
- and on the function end  `return CHECK_LAST_IS_STICKY_ERROR();`

## Linked Issues

Resolves #213


icicle/utils/error_handler.cuh
```diff
@@ -34,4 +36,27 @@ void inline syncDevice(const char* const file, const int line)
...
+#define CHECK_LAST_IS_STICKY_ERROR() checkLastStickyError(__FILE__, __LINE__)
+cudaError_t inline checkLastStickyError(const char* const file, const int line)
+{
+  cudaError_t err{cudaGetLastError()};
+  if (err != cudaSuccess) {
+    std::cerr << "CUDA Runtime Error at: " << file << ":" << line << std::endl;
+    std::cerr << cudaGetErrorString(err) << std::endl;
+    // check for sticky (unrecoverable) error where context is corrupted
+    // and the only option is to restart process
+    err = cudaDeviceSynchronize(); // TODO: only one call to cudaDeviceSynchronize seems sufficent according to
```
jeremyfelder, line 48: This can be a pretty heavy operation. Is calling `cudaGetLastError()` twice in a row and seeing both have the same error enough for detecting stickiness since the first one won't reset the error if it is sticky

vhnatyk: if I understand you correctly - according to[ this thread](https://forums.developer.nvidia.com/t/cuda-errors-determine-sticky-ness/271625) - and comment by [Robert Crovella](https://developer.nvidia.com/blog/author/bob-crovella/)  
_It seems sufficient simply to **call `cudaDeviceSynchronize()` twice** (or even just once). If the same error (or any error) is indicated, its a pretty good indicator_  . 

Just calling `cudaGetLastError()` won't be enough to determine stickiness - the first call resets the error. In the current draft - the rationale is if we have some error, it's worth checking "stickiness" at the expense of  `cudaDeviceSynchronize()` - but once, assuming it's sufficient. Anyway, this will be tested once we settle on the approach.

jeremyfelder: > Just calling cudaGetLastError() twice won't be enough to determine stickiness - the first call resets the error.

Robert had an answer [here](https://stackoverflow.com/questions/43659314/how-can-i-reset-the-cuda-error-to-success-with-driver-api-after-a-trap-instructi) that says calling `cudaGetLastError()` **will not reset** a sticky error. I don't see why that isn't a sufficient test for stickiness

icicle/appUtils/ntt/ntt.cu
```diff
@@ -11,8 +11,8 @@ namespace ntt {
 
   namespace {
 
-    const uint32_t MAX_NUM_THREADS = 512;
-    const uint32_t MAX_THREADS_BATCH = 512;          // TODO: allows 100% occupancy for scalar NTT for sm_86..sm_89
+    const uint32_t MAX_NUM_THREADS = 512;   // TODO: hotfix - should be 1024, currently limits shared memory size
```
ImmanuelSegol, line 14: please open issues for these todos.

vhnatyk:
closing in favor of #322 , comments will be addressed in subsequent pr's to dev and on merge to main

PR #326, opened 2024-01-02
New API for Version 1.0.0

jeremyfelder:

PR #329, opened 2024-01-05
Conditional PIC

vhnatyk:
## Describe the changes

This PR enables [PIC](https://en.wikipedia.org/wiki/Position-independent_code) by setting ICICLE_PIC env or cmake var

Given that supposed performance hit doesn't affect modern CPUs anymore, maybe worth enabling it unconditionally or we can reconsider ICICLE being static library. I recall choice of static was due to the possibility of deployment without requirement for extra libs and supposed performance benefit (that also probably is not quite actual?)

## Linked Issues

Resolves #326


yshekel:
@vhnatyk 
I suggest making PIC default to avoid this bug (where it is linked to a shared object, which should always be pic). Otherwise users must be aware of this flag, and may encounter the bug as well.
This way, disabling PIC must be explicit and this makes the library more robust.
Note that the disadvantage of PIC is slightly less efficient code but I believe this is negligible.

vhnatyk:
PIC is conditional - set default to enabled

DmytroTym (approved):

PR #330, opened 2024-01-06
NTT acceleration

sss1h:
## Describe the changes

This PR reduces NTT execution time by parallelizing the independent butterfly operations in NTT.

### DIFF
--- a/icicle/appUtils/ntt/ntt.cuh
+++ b/icicle/appUtils/ntt/ntt.cuh
@@ -5,9 +5,11 @@
 #include "../../utils/sharedmem.cuh"
 #include "../vector_manipulation/ve_mod_mult.cuh"
 
-const uint32_t MAX_NUM_THREADS = 512;
-const uint32_t MAX_THREADS_BATCH = 512;          // TODO: allows 100% occupancy for scalar NTT for sm_86..sm_89
-const uint32_t MAX_SHARED_MEM_ELEMENT_SIZE = 32; // TODO: occupancy calculator, hardcoded for sm_86..sm_89
+//const uint32_t MAX_NUM_THREADS = 512;
+const uint32_t MAX_NUM_THREADS = 1024;
+// const uint32_t MAX_THREADS_BATCH = 512;          // TODO: allows 100% occupancy for scalar NTT for sm_86..sm_89
+const uint32_t MAX_THREADS_BATCH = 1024;          // TODO: allows 100% occupancy for scalar NTT for sm_86..sm_89
+const uint32_t MAX_SHARED_MEM_ELEMENT_SIZE = 63; // TODO: occupancy calculator, hardcoded for sm_86..sm_89
 const uint32_t MAX_SHARED_MEM = MAX_SHARED_MEM_ELEMENT_SIZE * MAX_NUM_THREADS;
 
 /**
@@ -184,12 +186,16 @@ __global__ void ntt_template_kernel_shared(
   uint32_t logn)
 {
   SharedMemory<E> smem;
+  SharedMemory<S> smem_s;
   E* arr = smem.getPointer();
+  S* twiddles = smem_s.getPointer();
 
   uint32_t task = blockIdx.x;
   uint32_t loop_limit = blockDim.x;
   uint32_t chunks = n / (loop_limit * 2);
   uint32_t offset = (task / chunks) * n;
+  uint32_t elements = 1 << 2; // elements in a group
+  uint32_t groups = n / elements; // group number
   if (task < max_task) {
     // flattened loop allows parallel processing
     uint32_t l = threadIdx.x;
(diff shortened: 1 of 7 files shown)


ImmanuelSegol:
Hey @sss1h,
Thanks for this PR. mind contacting me on telegram (https://t.me/OxMonia) ?

icicle/appUtils/ntt/ntt.cuh
```diff
@@ -314,17 +355,47 @@ void ntt_inplace_batch_template(
...
+        <<<num_blocks / 2, num_threads / 2, 0, stream>>>(d_inout, n / 2, d_twiddles, n / 2, total_tasks / 2, s, false);
+      ntt_template_kernel<E, S>
+        <<<num_blocks / 2, num_threads / 2, 0, stream>>>(d_inout + n / 2, n / 2, d_twiddles + n / 2, n / 2, total_tasks / 2, s, false);
+      cudaError_t cudaError = cudaGetLastError();
+      if(cudaError != cudaSuccess) {
+        printf("\nCUDA error: %s\n", cudaGetErrorString(cudaError));       
+        exit(1);
+      }
     }
+    ntt_template_kernel<E, S>
```
DmytroTym, line 384: I think this line poses an issue for correctness in some cases: consider NTTs of size less than `1 << logn_shmem` - they are computed fully inside the `ntt_template_kernel_shared` kernel. However, this line introduces one extra call that breaks correctness, this can be checked by verifying that for sizes of 1 << 10 or less ntt and intt fail to cancel each other but if this call and it's analogue in the forward NTT are commented out, they start to cancel

icicle/appUtils/ntt/ntt.cuh
```diff
@@ -314,17 +355,47 @@ void ntt_inplace_batch_template(
...
+    // E* d_inout_original = new E[n];
+    // for (int i = 0; i < n; i++) {
+    //   d_inout_original[i] = d_inout[i];
+    // }
+    for (int s = logn_shmem; s < logn - 1; s++) // TODO: this loop also can be unrolled
     {
       ntt_template_kernel<E, S>
-        <<<num_blocks, num_threads, 0, stream>>>(d_inout, n, d_twiddles, n, total_tasks, s, false);
+        <<<num_blocks / 2, num_threads / 2, 0, stream>>>(d_inout, n / 2, d_twiddles, n / 2, total_tasks / 2, s, false);
+      ntt_template_kernel<E, S>
```
DmytroTym, line 376: As far as I understand, your idea here is to take all the stages of Cooley-Tukey algorithm except the very last one and somehow parallelise the two halves of each stage? I think correctness breaks here, one issue is that the total number of butterflies you launch in each kernel call is `num_blocks * num_threads / 4`, for two kernel calls this is `num_blocks * num_threads / 2` so two times fewer butterflies than we need to complete one stage of Cooley-Tukey. I think this two-fold reduction in the number of butterflies is the reason for the acceleration of your NTT as compared to the baseline.
I believe that splitting one kernel into two here is possible in principle; however this does not lead to more parallelism, but to less - two kernels get called one-by-one while one kernel call may exploit GPU resources as much as possible. So for small NTTs this approach of splitting kernels might result in worse runtimes whereas for large NTTs it seems to be basically the same when accounted for the correct number of butterflies.

icicle/appUtils/ntt/ntt.cuh
```diff
@@ -212,20 +218,55 @@ __global__ void ntt_template_kernel_shared(
...
         uint32_t k = oij + shift_s;
         S tw = r_twiddles[j * n_twiddles_div];
-
-        E u = s == 0 ? arr_g[offset + oij] : arr[oij];
-        E v = s == 0 ? arr_g[offset + k] : arr[k];
+        // twiddles[threadIdx.x] = r_twiddles[j * n_twiddles_div];
+        
+        // arr[threadIdx.x % groups * elements + threadIdx.x / groups * 2] = arr_g[offset + oij];
+        // arr[threadIdx.x % groups * elements + threadIdx.x / groups * 2 + 1] = arr_g[offset + k];
+        __shared__ E u, v;
```
DmytroTym, line 225: Why do `u` and `v` need to be shared variables here? They can be thread local IMO

DmytroTym (commented):
Hello @sss1h
Thanks a lot for contributing to ICICLE!
I see that Rust tests are failing and I tried to highlight places where in my opinion correctness is broken.
We're really curious about your use-case: do you need a fast large size NTT on bn254?

icicle/appUtils/ntt/ntt.cuh
sss1h: Thank you for your detailed review. This approach does indeed break correctness. Could you please give me more idea on how to split a kernel into two here to get more parallelism?  In my use case, I want a fast 2<sup>26</sup> size NTT on bn254. Thank you very much.

DmytroTym: Could you please write to us at [hi@ingonyama.com](mailto:hi@ingonyama.com) so we can discuss in more detail? :)

PR #333, opened 2024-01-08
Smart pointers and documentation

DmytroTym:
## Describe the changes

This PR introduces a pointer that can live on host or device and adds doc comments for Rust APIs


wrappers/rust/icicle-cuda-runtime/src/memory.rs
```diff
@@ -2,37 +2,67 @@ use crate::bindings::{cudaFree, cudaMalloc, cudaMallocAsync, cudaMemcpy, cudaMem
...
-            .is_empty()
+        match self {
+            Self::Device(s) => s.is_empty(),
+            Self::Host(v) => v.is_empty(),
+        }
     }
 
-    pub fn as_slice(&mut self) -> &mut [T] {
-        self.0
+    pub fn is_on_device(&self) -> bool {
```
jeremyfelder, line 29: Do we want to align this with the `are_on_device` in the NTT and MSM configs?

DmytroTym: If I understand you correctly, this is implemented. On device flags in configs are private and are set depending on on-deviceness of data, see for example https://github.com/ingonyama-zk/icicle/pull/333/files#diff-8ca95f9dad1c6679d35f1eacde6413f991f6a34e7bd2249c12d610a121abf10fR104

jeremyfelder: Oo i just meant the naming of the variables so the example you showed would be 

local_cfg.are_inputs_on_device = input.**are**_on_device();

DmytroTym: Oh. We can I guess, I just named it like that because "slice" is singular

PR #334, opened 2024-01-08
refactor: generate curve-specific function names with macro

yshekel:
For exposed functions that are curve-specific, the expected symbol is based on the curve name.
For example MSMCuda becomes bn254MSMCuda for CURVE=bn254.

currently it is implemented via objcopy by redefining symbols after compilation.

This PR modifies the function names at preprocessing time instead.
For shared objects, objcopy doesn't work (since it cannot modify the dynsym section).


vhnatyk:
perhaps needs recreation after recent rebase of dev by @jeremyfelder

PR #337, opened 2024-01-09
refactor: generate curve-specific function names with macro

yshekel:
For exposed functions that are curve-specific, the expected symbol is based on the curve name.
For example MSMCuda becomes bn254MSMCuda for CURVE=bn254.

currently it is implemented via objcopy by redefining symbols after compilation.

This PR modifies the function names at preprocessing time instead.
For shared objects, objcopy doesn't work (since it cannot modify the dynsym section).

notes:

1. maybe utils/utils.h is not great because any repo probably has such a file. maybe icicle_utils?
2. the code is a little more cluttered because of this macro. 
example:   'extern "C" cudaError_t MSMCuda' becomes  'extern "C" cudaError_t CONCAT_EXPAND(CURVE, MSMCuda)'

### DIFF
--- a/icicle/utils/utils.h
+++ b/icicle/utils/utils.h
@@ -1,0 +1,8 @@
+#pragma once
+#ifndef ICICLE_UTILS_H
+#define ICICLE_UTILS_H
+
+#define CONCAT_DIRECT(a, b) a##b
+#define CONCAT_EXPAND(a, b) CONCAT_DIRECT(a, b) // expand a,b before concatenation
+
+#endif // ICICLE_UTILS_H
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -4,9 +4,11 @@ on:
   pull_request:
     branches:
       - main
+      - dev
   push:
     branches:
       - main
+      - dev
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/main-format.yml
+++ b/.github/workflows/main-format.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
       - main
+      - dev
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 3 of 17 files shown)


icicle/utils/utils.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
```
DmytroTym, line 1: Upon googling about using `.cpp` and `.h` versus `.cu` and `.cuh`, [here's](https://stackoverflow.com/a/15234351) one upvoted answer from stack overflow:

> h, cpp, c, hpp, inc - files that don't contain CUDA C code (e.g. __device__ and other keywords, kernel calls, etc.) and do not make any cuda runtime calls (cuda... functions). It is perfectly fine to call CUDA driver API (cu...) functions from these files. Note that it is possible to compile these files with compilers other then NVCC.
> cu - CUDA C source files. These files are passed to the NVCC compiler to be compiled into linkable (host/device) objects.
> cuh, cuinc - files that are included in .cu files. These files can have CUDA C keywords and/or call CUDA runtime functions.

I assume you were using similar logic? Maybe we should adopt this as a convention cos currently everything is `.cu` and `.cuh`

DmytroTym (approved):
These changes break the C++-side MSM test but I guess we should either delete it or bring it under the googletest umbrella. Also, maybe worth having the change in how we pass a curve from CMake implemented here too, shouldn't be too much code right?
In general, looks great to me :)

icicle/utils/utils.h
yshekel: AFAIK the suffix of included files doesn't matter since in C++ (in contrast to other languages) they are simply inlined by the preprocessor and only c/cpp/cu files are compiled.

So I think we should agree on a convention and stick with it. I actually used '.h' since I am used to it and did not think about that TBH.

vhnatyk:
> These changes break the C++-side MSM test but I guess we should either delete it or bring it under the googletest umbrella. Also, maybe worth having the change in how we pass a curve from CMake implemented here too, shouldn't be too much code right? In general, looks great to me :)

+1 for gtest, I'd say we need to refrain from practice where feature owners leave these clunky files messing around, quite commonly they are broken - either pls don't commit them or add as gtest 👍🏻

DmytroTym (approved):
lgtm

PR #341, opened 2024-01-12
G2 fix and BW6 scalar field on the Rust side

DmytroTym:
A couple of notes about the implementation:

- For bw6-761, the definition of its scalar field was moved to the bls12-377 crate, so foreign traits that needed to be implemented on it (like NTT) had to be implemented in bls12-377 crate. But because these traits need native functions to be linked earlier, I had to move cmake build of bw6-761 to bls12-377's `build.rs`. It was put under a feature to avoid excessive compilation in cases when bls12-377 is used stand-alone. One issue with this whole approach is that building the two curves is now slower because cargo cannot parallelise like where there were two `build.rs` scripts. We might do manual parallelisation of course if we think this is important.
- I used `mod`s a lot to distinguish functions with the same name in the same scope which started happening a lot when I implemented this PR. All of these functions one way or another appear inside macros so one alternative solution is modifying function names by something like the [paste](https://crates.io/crates/paste) crate, but I feel like module-based solution is more idiomatic and doesn't need any third-party crates so I went with it.


icicle/appUtils/ntt/ntt.cu
```diff
@@ -520,7 +520,7 @@ namespace ntt {
    *  - `S` is the [scalar field](@ref scalar_t) of the curve;
    */
   extern "C" cudaError_t
-  CONCAT_EXPAND(CURVE, InitializeDomain)(curve_config::scalar_t primitive_root, device_context::DeviceContext& ctx)
+    CONCAT_EXPAND(CURVE, InitializeDomain)(curve_config::scalar_t primitive_root, device_context::DeviceContext& ctx)
```
jeremyfelder, line 523: Formatting

wrappers/rust/icicle-curves/icicle-bn254/build.rs
```diff
@@ -4,12 +4,18 @@ fn main() {
     println!("cargo:rerun-if-env-changed=CXXFLAGS");
     println!("cargo:rerun-if-changed=../../../../icicle");
 
+    #[cfg(not(feature = "g2"))]
+    let g2_str = "";
+    #[cfg(feature = "g2")]
+    let g2_str = "ON";
+
```
jeremyfelder, line 11: Personal preference/nit:

```rust
    // Base config
    let mut config = Config::new("../../../../icicle");
    config
        .define("BUILD_TESTS", "OFF")
        .define("CURVE", "bn254")
        .define("CMAKE_BUILD_TYPE", "Release");

    // Optional Features
    #[cfg(feature = "g2")]
    config.define("G2_DEFINED", "ON");

    // Build
    let out_dir = config
        .build_target("icicle")
        .build();
```

wrappers/rust/icicle-curves/icicle-bw6-761/src/curve.rs
```diff
@@ -1,30 +1,57 @@
...
 impl_field!(BASE_LIMBS, BaseField, BaseCfg, Fq);
-impl_curve!("bw6_761", CurveCfg, ScalarField, BaseField);
+pub type ScalarField = bls12_377BaseField;
+impl_curve!(
+    "bw6_761",
+    bw6_761,
+    CurveCfg,
+    ScalarField,
+    BaseField,
+    ArkG1Config,
```
ChickenLover, line 24: What if arkworks feature is not enabled?

ChickenLover (commented):
Love the idea with `mod` for extern functions namespaces

wrappers/rust/icicle-curves/icicle-bw6-761/src/curve.rs
DmytroTym: Still compiles because `ArkG1Config` and other similar names are treated as meaningless unused idents if the feature is disabled

icicle/appUtils/ntt/ntt.cu
DmytroTym: Which version of clang-format are you/CI machine using?

DmytroTym:
Meh, bw6 tests ran out of memory. The card on France seems to have 2Gb of RAM. Not sure what's best solution, I can bring `-- --test-threads=1` back just for CI

icicle/appUtils/msm/msm.cu
```diff
@@ -967,7 +967,8 @@ namespace msm {
 
   } // namespace
 
-  extern "C" MSMConfig CONCAT_EXPAND(CURVE, DefaultMSMConfig)()
+  template <typename A>
```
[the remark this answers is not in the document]
ChickenLover: The only way I can think of is moving the definition into msm.cuh

icicle/appUtils/ntt/ntt.cu
jeremyfelder: locally I'm running 14.0.0.
the github hosted runners are running the software provided here https://github.com/actions/runner-images/blob/ubuntu22/20240107.1/images/ubuntu/Ubuntu2204-Readme.md#installed-software

icicle/appUtils/msm/msm.cu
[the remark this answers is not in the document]
DmytroTym: Doesn't really work as `msm.cu` `#include`s `msm.cuh`

jeremyfelder (approved):

PR #342, opened 2024-01-16
refactor: consolidate msm and batch-msm implementations to one function

yshekel:
refactor: consolidate msm and batch-msm implementations to one function  
   - now batch-msm support parallel BM accumulation in addition to large triangle accumulation

### DIFF
--- a/icicle/appUtils/msm/msm.cu
+++ b/icicle/appUtils/msm/msm.cu
@@ -302,7 +302,7 @@ namespace msm {
     __global__ void ssm_buckets_kernel(P* buckets, unsigned* single_bucket_indices, unsigned nof_buckets, unsigned c)
     {
       unsigned tid = (blockIdx.x * blockDim.x) + threadIdx.x;
-      if (tid > nof_buckets) return;
+      if (tid >= nof_buckets) return;
       unsigned bucket_index = single_bucket_indices[tid];
       S scalar_bucket_multiplier;
       scalar_bucket_multiplier = {
@@ -314,20 +314,22 @@ namespace msm {
     __global__ void last_pass_kernel(const P* final_buckets, P* final_sums, unsigned num_sums)
     {
       unsigned tid = (blockIdx.x * blockDim.x) + threadIdx.x;
-      if (tid > num_sums) return;
+      if (tid >= num_sums) return;
       final_sums[tid] = final_buckets[2 * tid + 1];
     }
 
     // this kernel computes the final result using the double and add algorithm
     // it is done by a single thread
     template <typename P, typename S>
-    __global__ void
-    final_accumulation_kernel(const P* final_sums, P* final_results, unsigned nof_msms, unsigned nof_bms, unsigned c)
+    __global__ void final_accumulation_kernel(
+      const P* final_sums, P* final_results, unsigned nof_msms, unsigned nof_bms, unsigned nof_empty_bms, unsigned c)
     {
       unsigned tid = (blockIdx.x * blockDim.x) + threadIdx.x;
-      if (tid > nof_msms) return;
+      if (tid >= nof_msms) return;
       P final_result = P::zero();
-      for (unsigned i = nof_bms; i > 1; i--) {
+      // Note: in some cases accumulation of bm is implemented such that some bms are known to be empty. Therefore
+      // skiping them.
+      for (unsigned i = nof_bms - nof_empty_bms; i > 1; i--) {
         final_result = final_result + final_sums[i - 1 + tid * nof_bms]; // add
         for (unsigned j = 0; j < c; j++)                                 // double
         {
@@ -344,9 +346,10 @@ namespace msm {
       unsigned c,
       S* scalars,
       A* points,
-      unsigned batch_size,
-      unsigned msm_size,
-      unsigned nof_points,
+      unsigned batch_size,      // number of MSMs to compute
+      unsigned single_msm_size, // number of elements per MSM (a.k.a N)
+      unsigned nof_points,      // numer of EC points in 'points' array. Must be either (1) single_msm_size if MSMs are
+                                // sharing points or (2) single_msm_size*batch_size otherwise
       P* final_result,
       bool are_scalars_on_device,
       bool are_scalars_montgomery_form,
@@ -360,7 +363,13 @@ namespace msm {
     {
       CHK_INIT_IF_RETURN();
 
-      const unsigned nof_scalars = batch_size * msm_size; // assuming scalars not shared between batch elements
+      const unsigned nof_scalars = batch_size * single_msm_size; // assuming scalars not shared between batch elements
+      const bool is_nof_points_valid = (nof_points == single_msm_size) || (nof_points == single_msm_size * batch_size);
+      if (!is_nof_points_valid) {
+        THROW_ICICLE_ERR(
+          IcicleError_t::InvalidArgument, "bucket_method_msm: #points must be either (1) single_msm_size if sharing "
+                                          "points or (2) single_msm_size*batch_size");
+      }
 
       S* d_scalars;
       A* d_points;
@@ -406,14 +415,15 @@ namespace msm {
 
       P* buckets;
       // compute number of bucket modules and number of buckets in each module
-      unsigned nof_bms = (bitsize + c - 1) / c;
-      unsigned bm_bitsize = (unsigned)ceil(log2(nof_bms));
+      unsigned nof_bms_per_msm = (bitsize + c - 1) / c;
+      unsigned bm_bitsize = (unsigned)ceil(log2(nof_bms_per_msm));
+      unsigned nof_bms_in_batch = nof_bms_per_msm * batch_size;
 #ifdef SIGNED_DIG
-      unsigned nof_buckets = nof_bms * ((1 << (c - 1)) + 1); // signed digits
+      const unsigned nof_buckets = nof_bms_per_msm * ((1 << (c - 1)) + 1); // signed digits
 #else
-      unsigned nof_buckets = nof_bms << c;
+      const unsigned nof_buckets = nof_bms_per_msm << c;
 #endif
-      unsigned total_nof_buckets = nof_buckets * batch_size;
+      const unsigned total_nof_buckets = nof_buckets * batch_size;
       CHK_IF_RETURN(cudaMallocAsync(&buckets, sizeof(P) * total_nof_buckets, stream));
 
       // launch the bucket initialization kernel with maximum threads
@@ -423,15 +433,15 @@ namespace msm {
 
       unsigned* bucket_indices;
       unsigned* point_indices;
-      CHK_IF_RETURN(cudaMallocAsync(&bucket_indices, sizeof(unsigned) * nof_scalars * (nof_bms + 1), stream));
-      CHK_IF_RETURN(cudaMallocAsync(&point_indices, sizeof(unsigned) * nof_scalars * (nof_bms + 1), stream));
+      CHK_IF_RETURN(cudaMallocAsync(&bucket_indices, sizeof(unsigned) * nof_scalars * (nof_bms_per_msm + 1), stream));
+      CHK_IF_RETURN(cudaMallocAsync(&point_indices, sizeof(unsigned) * nof_scalars * (nof_bms_per_msm + 1), stream));
 
       // split scalars into digits
       NUM_THREADS = 1 << 10;
       NUM_BLOCKS = (nof_scalars + NUM_THREADS - 1) / NUM_THREADS;
       split_scalars_kernel<<<NUM_BLOCKS, NUM_THREADS, 0, stream>>>(
-        bucket_indices + nof_scalars, point_indices + nof_scalars, d_scalars, nof_scalars, nof_points, msm_size,
-        nof_bms, bm_bitsize, c); //+nof_scalars - leaving the first bm free for the out of place sort later
+        bucket_indices + nof_scalars, point_indices + nof_scalars, d_scalars, nof_scalars, nof_points, single_msm_size,
+        nof_bms_per_msm, bm_bitsize, c); //+nof_scalars - leaving the first bm free for the out of place sort later
 
       // sort indices - the indices are sorted from smallest to largest in order to group together the points that
       // belong to each bucket
@@ -444,7 +454,7 @@ namespace msm {
         sort_indices_temp_storage, sort_indices_temp_storage_bytes, bucket_indices + nof_scalars, bucket_indices,
         point_indices + nof_scalars, point_indices, nof_scalars, 0, sizeof(unsigned) * 8, stream));
       CHK_IF_RETURN(cudaMallocAsync(&sort_indices_temp_storage, sort_indices_temp_storage_bytes, stream));
-      for (unsigned i = 0; i < nof_bms; i++) {
+      for (unsigned i = 0; i < nof_bms_per_msm; i++) {
         unsigned offset_out = i * nof_scalars;
         unsigned offset_in = offset_out + nof_scalars;
         // The second to last parameter is the default value supplied explicitly to allow passing the stream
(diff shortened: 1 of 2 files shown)


icicle/appUtils/msm/msm.cu
```diff
@@ -338,59 +340,63 @@ namespace msm {
...
     template <typename S, typename P, typename A>
     cudaError_t bucket_method_msm(
-      int bitsize,
-      int c,
+      unsigned bitsize,
+      unsigned c,
       S* scalars,
       A* points,
-      int size,
+      unsigned batch_size,
```
HadarIngonyama, line 347: The definition of these 3 variables is unclear, please add a description. Also, it seems that only 2 out of 3 are needed no?

icicle/appUtils/msm/msm.cu
```diff
@@ -530,7 +542,7 @@ namespace msm {
       CHK_IF_RETURN(cudaFreeAsync(sort_single_temp_storage, stream));
 
       // find large buckets
-      unsigned avarage_size = size / (1 << c);
+      unsigned avarage_size = msm_size / (1 << c);
```
HadarIngonyama, line 545: making sure that this is the size of a single msm

icicle/appUtils/msm/msm.cu
```diff
@@ -636,66 +649,73 @@ namespace msm {
...
 
       P* d_final_result;
-      if (!is_result_on_device) CHK_IF_RETURN(cudaMallocAsync(&d_final_result, sizeof(P), stream));
+      if (!are_results_on_device) CHK_IF_RETURN(cudaMallocAsync(&d_final_result, sizeof(P) * batch_size, stream));
 
       P* final_results;
-      if (is_big_triangle) {
-        CHK_IF_RETURN(cudaMallocAsync(&final_results, sizeof(P) * nof_bms, stream));
+      if (is_big_triangle || c == 1) {
+        CHK_IF_RETURN(cudaMallocAsync(&final_results, sizeof(P) * nof_bms * batch_size, stream));
```
HadarIngonyama, line 656: this seems to be used a lot, maybe define total_nof_bms = nof_bms*batch_size

icicle/appUtils/msm/msm.cu
```diff
@@ -636,66 +649,73 @@ namespace msm {
...
+                is_first_iter ? source_buckets : temp_buckets2, is_last_iter ? target_buckets : temp_buckets2,
+                1 << (target_bits_count - j), is_last_iter ? 1 << target_bits_count : 0, 1 /*=write_phase*/,
+                0 /*=padding*/, nof_threads);
             }
           }
           if (target_bits_count == 1) {
-            nof_bms = bitsize;
-            CHK_IF_RETURN(cudaMallocAsync(&final_results, sizeof(P) * nof_bms, stream));
+            // Note: the reduction ends up with 'target_windows_count' windows per batch element. for batch=1 we can
+            // only sum the first 'bitsize' elements to avoid a few ECADD which add zeros
```
HadarIngonyama, line 711: why isn't this true for every batch size?

icicle/appUtils/msm/msm.cu
```diff
@@ -717,18 +737,21 @@ namespace msm {
         }
       }
 
-      // launch the double and add kernel, a single thread
-      final_accumulation_kernel<P, S>
-        <<<1, 1, 0, stream>>>(final_results, is_result_on_device ? final_result : d_final_result, 1, nof_bms, c);
+      // launch the double and add kernel, a single thread per batch element
+      NUM_THREADS = 1;
```
HadarIngonyama, line 741: better use 32 threads min

icicle/appUtils/msm/msm.cu
```diff
@@ -992,21 +799,22 @@ namespace msm {
...
-        config.are_scalars_montgomery_form, config.are_points_on_device, config.are_points_montgomery_form,
-        config.are_results_on_device, config.is_big_triangle, config.large_bucket_factor, config.is_async, stream));
-    else
-      return CHK_STICKY(batched_bucket_method_msm(
-        bitsize, (config.c == 0) ? get_optimal_c<S>(msm_size) : config.c, scalars, points, config.batch_size, msm_size,
-        (config.points_size == 0) ? msm_size : config.points_size, results, config.are_scalars_on_device,
-        config.are_scalars_montgomery_form, config.are_points_on_device, config.are_points_montgomery_form,
-        config.are_results_on_device, config.is_async, stream));
+
+    unsigned c = config.batch_size > 1 ? ((config.c == 0) ? get_optimal_c(msm_size) : config.c) : 16;
```
HadarIngonyama, line 805: why is 16 the default for a single msm? it should also be the closest power of 2 to the optimal c

HadarIngonyama (changes requested):
Looks good to me apart from the comments. In addition - which tests and benches did you run? we need to make sure that batched and non batched are passing and that performance was not affected.

icicle/appUtils/msm/msm.cu
yshekel: yes it is. I will rename variables to make the code clearer.

icicle/appUtils/msm/msm.cu
yshekel: ok

icicle/appUtils/msm/msm.cu
yshekel: It is true for every batch size. 
The thing is that for fields like BLS12-377 where the scalar field is 253b we end up with >253 bms (e.g. 254 bms for c=2 with 1 iteration of reduction from 127 to 254 bms. For other cases can be a little more than that) where the most significant ones are guaranteed to be zero. 
The original code summed the bms based on bitsize so effectively summing only the non-zero bms. It is slightly more efficient more summing all bms (254 in the example I gave). I wrote the code this way for batch=1 since I did not want to end up with worse performance (although maybe negligible).

generally for batch>1 we must account for all bms so that the other batch elements are summing the right bms. If setting bm=253, the second batch element is summing wrong bms and all results end up wrong except for the first batch element.

Do you think I should always set bm based on the final number of windows? it is slightly less efficient but more drastic.

icicle/appUtils/msm/msm.cu
yshekel: ok

icicle/appUtils/msm/msm.cu
yshekel: It was like that for batch=1 and so I did not want to end up with different behavior.

Should I use the closest power of 2 instead? closest from below? I mean does 12 goes to 8 to 16?

yshekel:
> Looks good to me apart from the comments. In addition - which tests and benches did you run? we need to make sure that batched and non batched are passing and that performance was not affected.

Thanks for the comments, I will fix and push an additional commit.

The PR is testing all the curves, 15 tests each, including 3 MSM tests: msm, batch-msm, skewed-msm. They all Pass. It's the "Test / Test Rust on Linux" check.

Regarding performance: @DmytroTym (please correct me if wrong) profiled batched-msm with this PR: (a) batch=16 N=2^16 and also (b) batch=3 N=1000. For case (a) he measured ~60.7[ms] compared to ~413.8[ms] for the old implementation that used the triangle accumulation. For case (b) he measured ~12.7[ms] compared to the old ~8.9[ms]. In that small case the triangle accumulation is slightly better but of course it is configureable, and maybe a better choice of c can also improve the performance.

I think it will be very useful to have a benchmark to measure such modifications but AFAIK there isn't one. Is that right?

icicle/appUtils/msm/msm.cu
HadarIngonyama: why not have each msm sum only the non zero buckets? (and make sure that the buckets are the correct ones - I'm pretty sure you can use bm=253 and correct the kernel to sum the correct buckets) let's go for best performance

icicle/appUtils/msm/msm.cu
HadarIngonyama: I don't remember.. use the value that gives better performance:)

HadarIngonyama:
> > Looks good to me apart from the comments. In addition - which tests and benches did you run? we need to make sure that batched and non batched are passing and that performance was not affected.
> 
> Thanks for the comments, I will fix and push an additional commit.
> 
> The PR is testing all the curves, 15 tests each, including 3 MSM tests: msm, batch-msm, skewed-msm. They all Pass. It's the "Test / Test Rust on Linux" check.
> 
> Regarding performance: @DmytroTym (please correct me if wrong) profiled batched-msm with this PR: (a) batch=16 N=2^16 and also (b) batch=3 N=1000. For case (a) he measured ~60.7[ms] compared to ~413.8[ms] for the old implementation that used the triangle accumulation. For case (b) he measured ~12.7[ms] compared to the old ~8.9[ms]. In that small case the triangle accumulation is slightly better but of course it is configureable, and maybe a better choice of c can also improve the performance.
> 
> I think it will be very useful to have a benchmark to measure such modifications but AFAIK there isn't one. Is that right?

I believe there is a "benches" folder isn't that so?
also - best to have the msm use the "big triangle" automatically only for the parameters where it improves performance - have it as a default in the code.
another check that needs to be done is to check that the performance of a single msm (batch size=1) has not degraded as a result of the new changes.

icicle/appUtils/msm/msm.cu
yshekel: Yes I think you are right. I will do that.

icicle/appUtils/msm/msm.cu
yshekel: I think the optimum is log(N) (meaning the number of ECADD is minimized) but implementation-wise power of 2 work better from what I read. In addition memory grows exponential in c since you have 2^c buckets per BM so practically c=16 is the largest power of two that is reasonable.

Having said that, the code says:
    **unsigned get_optimal_c(int bitsize) { return max((unsigned)ceil(log2(bitsize)) - 4, 1U); }**

I don't understand why '-4'? It is quite odd since I then round it back to a power of two.

Anyway I will try to play with it a little and figure out what value are working better, but I think this is an optimization we can dedicate another PR and task for since I think it's a heuristic and not really part of this PR.

yshekel:
> > > Looks good to me apart from the comments. In addition - which tests and benches did you run? we need to make sure that batched and non batched are passing and that performance was not affected.
> > 
> > 
> > Thanks for the comments, I will fix and push an additional commit.
> > The PR is testing all the curves, 15 tests each, including 3 MSM tests: msm, batch-msm, skewed-msm. They all Pass. It's the "Test / Test Rust on Linux" check.
> > Regarding performance: @DmytroTym (please correct me if wrong) profiled batched-msm with this PR: (a) batch=16 N=2^16 and also (b) batch=3 N=1000. For case (a) he measured ~60.7[ms] compared to ~413.8[ms] for the old implementation that used the triangle accumulation. For case (b) he measured ~12.7[ms] compared to the old ~8.9[ms]. In that small case the triangle accumulation is slightly better but of course it is configureable, and maybe a better choice of c can also improve the performance.
> > I think it will be very useful to have a benchmark to measure such modifications but AFAIK there isn't one. Is that right?
> 
> I believe there is a "benches" folder isn't that so? also - best to have the msm use the "big triangle" automatically only for the parameters where it improves performance - have it as a default in the code. another check that needs to be done is to check that the performance of a single msm (batch size=1) has not degraded as a result of the new changes.

I did not see a 'benches' directory. Maybe it was removed.
I will try to see where big-triangle is more efficient in terms of N and batch-size or maybe N*batch_size and default to that.
Regarding batch=1 performance, I will verify it did not degrade but basically this is the same implementation exactly so other than multiplying by 1 in a few places (on host), I don't think it should be different at all.

icicle/appUtils/msm/msm.cu
HadarIngonyama: yes, for now just change 16 to get_optimal_c. (which will then be rounded to the closest power of 2)
the logn-4 is the value that minimizes the amount of ec additions for most sizes, but it's true that with the actual implementation there are other considerations too. Ideally we will have a dictionary.

icicle/appUtils/msm/msm.cu
```diff
@@ -992,21 +799,22 @@ namespace msm {
...
+    // reduce c to closest power of two if not using big_triangle reduction logic
+    // TODO: support arbitrary values of c
+    if (!config.is_big_triangle) {
+      while ((c & (c - 1)) != 0)
```
DmytroTym, line 809: Can't this procedure just hang for some values of `c`? Like, if `c=b101`, then `c & (c - 1)` doesn't add any extra bits so the process is never terminated. Also since the idea is rounding up I'm afraid any MSM of size bigger than 2^20 will end up with `c=32` which takes way too much memory

icicle/appUtils/msm/msm.cu
```diff
@@ -338,59 +340,63 @@ namespace msm {
     // this function computes msm using the bucket method
     template <typename S, typename P, typename A>
     cudaError_t bucket_method_msm(
-      int bitsize,
-      int c,
+      unsigned bitsize,
```
DmytroTym, line 343: Just a style related note: google guide advices to basically use `int` everywhere. I know that we're still not following this rule fully, but we at least tried to do it in higher-level and public functions. Do you think it's a good idea to stick to it?

DmytroTym (changes requested):
Looks good with just one change request related to rounding `c` to a power of 2
@HadarIngonyama in v1.0.0 we removed benches for now (but we should return them asap). But I indeed profiled some batched MSMs and saw improvements (mostly), the perf of singular MSM is unchanged

icicle/appUtils/msm/msm.cu
HadarIngonyama: I know someone who is very experienced with CUDA and he uses uint32_t wherever he can:)

DmytroTym: Well limbs obviously shouldn't be `int`. I'll just quote google guide to be more precise:

> We use int very often, for integers we know are not going to be too big, e.g., loop counters. Use plain old int for such things.
> ...
> You should not use the unsigned integer types such as uint32_t, unless there is a valid reason such as representing a bit pattern rather than a number, or you need defined overflow modulo 2^N. In particular, do not use unsigned types to say a number will never be negative. Instead, use assertions for this.

icicle/appUtils/msm/msm.cu
yshekel: I don't think 2 variables are enough:
- msm_size is the number of elements per msm sum
- batch_size is how many msms to compute in the batch
- nof_points can either be msm_size (if sharing points) or msm_size*batch_size (if not sharing points).

maybe I can replace nof_points by a boolean that tells if sharing points between batch elements or not. It would still be 3 variables and require modification of 'MSMConfig' which would break the API so I prefer avoiding it.

DmytroTym: I would leave three variables. By design it should be possible to have different points in the first and second MSMs and then re-use points from the first for the third and then points from the second for the fourth and so on...

icicle/appUtils/msm/msm.cu
DmytroTym: We discussed in the daily that fine-tuning the value of `c` is probably out of the scope of this PR..
Imo it'll more appropriate to tabulate best values of `c` and optimal accumulation method (triangle vs. non-triangle) when we establish proper benchmarking and run out of options to significantly speed up the MSM (potentially altering choices that we hard-coded). I think for now it's ok to leave reasonable defaults and give users an option to alter them via config

yshekel: I will leave it as is for now then, meaning the behavior remains as it is today.

icicle/appUtils/msm/msm.cu
yshekel: it will never hang and it rounds down, not up:
- for powers of two exactly 1 bit is 1'b1 --> meaning 'c & (c-1)' is zero. for example for c=4'b0100, c-1 = 4'b0011.
- otherwise each iteration is masking one bit to 1'b0, making the number smaller and therefore it must end in at most 31 iterations.
- note that masking c with any number cannot result in a number larger than c, meaning it is rounding down.

in your example c=b101 therefore c-1=b100 --> c& (c-1) = b100. Note that the first non zero bit is masked to 1'b0 and next iteration would mask the next non zero bit.

icicle/appUtils/msm/msm.cu
yshekel: We discussed it in the daily meeting.
I personally prefer unsigned values (and actually uint32_t as Hadar mentioned since it's more portable and can avoid nasty bugs since sizeof(int) is not guaranteed to be 4 in all systems) when the value should not be negative because it tells the reader that the value should not be negative.

Two notes:
(1) google style suggests asserting the values are not negative instead of using unsigned int. I am not sure it is applicable to kernel code.
(2) when using std containers like std::vector, their size() method returns an unsigned value and therefore the compiler will warn about comparison if the loop counter is signed. Iterators are probably the better design pattern in those cases but this is not really the point here.

icicle/appUtils/msm/msm.cu
DmytroTym: oh sorry you're right

DmytroTym (approved):

PR #345, opened 2024-01-22
Fix/windows build failing

jeremyfelder:
## Describe the changes

This PR fixes the rust build for windows by:
- Making the curve config's ntt size conditionally accessed when creating fields
- Removing the extern C function that returned a UDT containing non-POD types and replacing it with a `default_config` function on the Rust bindings side

### DIFF
--- a/icicle/appUtils/ntt/ntt.cu
+++ b/icicle/appUtils/ntt/ntt.cu
@@ -386,7 +386,8 @@ namespace ntt {
     // but it's a singleton that is supposed to be initialized once per program lifetime
     if (!Domain<S>::twiddles) {
       S omega = primitive_root;
-      for (int i = 0; i < S::TWO_ADICITY; i++)
+      unsigned omegas_count = S::get_omegas_count();
+      for (int i = 0; i < omegas_count; i++)
         omega = S::sqr(omega);
       if (omega != S::one()) {
         std::cerr << "Primitive root provided to the InitDomain function is not in the subgroup" << '\n';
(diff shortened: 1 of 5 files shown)


icicle/appUtils/ntt/ntt.cu
```diff
@@ -386,7 +386,8 @@ namespace ntt {
     // but it's a singleton that is supposed to be initialized once per program lifetime
     if (!Domain<S>::twiddles) {
       S omega = primitive_root;
-      for (int i = 0; i < S::TWO_ADICITY; i++)
+      unsigned omegas_count = S::get_omegas_count();
```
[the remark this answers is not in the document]
jeremyfelder: This solves the first issue of there not being a `omegas_count` member on an `fq_config`

icicle/primitives/field.cuh
```diff
@@ -85,6 +84,26 @@ public:
...
+      return CONFIG::omegas_count;
+    } else {
+      return 0;
+    }
+  }
+
+  template <typename T>
+  static constexpr bool has_member_name()
+  {
+    // if (std::is_same<unsigned, decltype(T::omegas_count)>) {
```
yshekel, line 99: please remove commented out code

icicle/primitives/field.cuh
```diff
@@ -85,6 +84,26 @@ public:
...
+    if (sizeof(T::omegas_count) > 0) {
```
yshekel, line 100: minor comment but I would rewrite as 'return sizeof(T::omegas_count) > 0' and avoid the if-else.

yshekel (approved):
Looks good to me. I wrote minor comments.

PR #356, opened 2024-01-30
multi card support

### LINKED ISSUE #135: [FEAT]: multi card support
## Description

Is it possible to detect, and if relevant parallelize over, multiple cards?

## Motivation

For gnark for example, we know that from a certain size of a circuit - multiple cards can divide the workload between them and offer close to linear improvement in performance

vhnatyk:
## Describe the changes

This PR enables multi-gpu support

## Linked Issues

Resolves #135


wrappers/rust/icicle-core/Cargo.toml
```diff
@@ -2,21 +2,24 @@
...
 
 icicle-cuda-runtime = { path = "../icicle-cuda-runtime" }
 
 ark-ff = { version = "0.4.0", optional = true }
-ark-ec = { version = "0.4.0", optional = true, features = [ "parallel" ] }
+ark-ec = { version = "0.4.0", optional = true, features = ["parallel"] }
 ark-poly = { version = "0.4.0", optional = true }
 ark-std = { version = "0.4.0", optional = true }
 
+#[dev-dependencies] # TODO: add as dev dependency?
```
jeremyfelder, line 20: Whats stopping us from making this a dev-dependency now?

jeremyfelder:
We should add documentation about how to use multiple devices. It wasn't clear to me at first how a Device slice could be set on anything but the default device

wrappers/rust/icicle-core/Cargo.toml
vhnatyk: I guess since test files are importing files containing `check_...` functions - these are available not only in `#[test]` - maybe this for examples? I personally would add rayon not only to tests - since it's imo the defacto standard for (non-async) multithreading anyway. but maybe we would like go async or tokio way etc

wrappers/rust/icicle-core/src/ntt/mod.rs
```diff
@@ -72,6 +72,18 @@ impl<'a, S: FieldImpl> NTTConfig<'a, S> {
             is_async: false,
         }
     }
+    pub fn default_config_for_device(device_id: usize) -> Self {
```
ChickenLover, line 75: Maybe it's better to call `default_config` and replace the ctx? Just to avoid code duplication

wrappers/rust/icicle-cuda-runtime/src/device_context.rs
```diff
@@ -1,27 +1,39 @@
...
-    /// Mempool to use. Default value: 0.
+    /// Mempool to use. Default value: 0. //TODO: multiple mempools per device ?
     pub mempool: CudaMemPool, // Assuming the type is provided by a CUDA binding crate
 }
 
 pub fn get_default_device_context() -> DeviceContext<'static> {
+    get_default_context_for_device(DEFAULT_DEVICE_ID)
+}
+
+// TODO: CudaResult
```
ChickenLover, line 25: It would be nice to add the doc-comment saying this will call set_device

wrappers/rust/icicle-cuda-runtime/src/memory.rs
```diff
@@ -1,67 +1,115 @@
...
+const UNDEFINED_DEVICE: i32 = -1;
+
 pub enum HostOrDeviceSlice<'a, T> {
     Host(Vec<T>),
-    Device(&'a mut [T]),
+    Device(&'a mut [T], i32),
 }
 
 impl<'a, T> HostOrDeviceSlice<'a, T> {
+    pub fn set_device_id(&mut self, new_id: i32) {
```
ChickenLover, line 20: What's this function for? Can't find a usage anywhere in the code

wrappers/rust/icicle-core/src/field.rs
```diff
@@ -15,6 +15,9 @@ pub struct Field<const NUM_LIMBS: usize, F: FieldConfig> {
     p: PhantomData<F>,
 }
 
+unsafe impl<const NUM_LIMBS: usize, F: FieldConfig> Send for Field<NUM_LIMBS, F> {}
+unsafe impl<const NUM_LIMBS: usize, F: FieldConfig> Sync for Field<NUM_LIMBS, F> {}
```
ChickenLover, line 19: What are these for? Field should derive Send and Sync automatically, doesn't it?

ChickenLover: I just tested it on my machine - it works without send and sync

ChickenLover (changes requested):

wrappers/rust/icicle-core/src/field.rs
vhnatyk: umm right as we discussed on a meeting - as long as you don't pass Field across the threads - all is fine, but if you do generate scalars on one thread and reuse across other threads as unmodified params for example etc - then you can't without these - automatic derive is prevented. I didn't add tests with that scenario because current tests cover that and make it a bit redundant but I didn't  want to modify existing tests as well  - again imo it's a bit redundant as they look like twins except inputs reusing across threads - and basic tutorial for `Arc(Mutex())`

ChickenLover: I see. Ok then

wrappers/rust/icicle-core/src/msm/tests.rs
```diff
@@ -1,8 +1,11 @@
 use super::{get_default_msm_config, msm, MSM};
 use crate::curve::{Affine, Curve, Projective};
 use crate::traits::{FieldImpl, GenerateRandom};
+use icicle_cuda_runtime::device::{get_device_count, set_device};
```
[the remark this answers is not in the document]
vhnatyk: yes, int for the actual number of cards

wrappers/rust/icicle-cuda-runtime/src/memory.rs
vhnatyk: setter for device_id of slice - I set it during slice on-device allocation - so indeed not used

wrappers/rust/icicle-core/src/ntt/mod.rs
vhnatyk: hmm create pool and stream and then abandon them? the calls are probably inexpensive, though - let's discuss

wrappers/rust/icicle-cuda-runtime/src/memory.rs
vhnatyk: removed as discussed

wrappers/rust/icicle-core/Cargo.toml
vhnatyk: just double checked - it builds if rayon added as dev-dependency, probably rust-analyser issue

wrappers/rust/icicle-core/src/ntt/mod.rs
vhnatyk: I just updated the design doc - after rethinking a bit - I guess it's a bad idea to call set_device implicitly - even if documented

DmytroTym, line 75: Maybe let's leave one function? Either `default_config_for_device` deleting the existing `default_config` or without this function, just setting the device to 0 by default

DmytroTym:
Maybe because setting a new device in the context automatically invalidates the context and setting a different stream invalidates the context if this stream was created for a different device, we should re-design or at least encapsulate the context somehow? Though I guess it's a breaking change technically

wrappers/rust/icicle-core/src/ntt/mod.rs
vhnatyk: removing old is a breaking change and we need new function anyway

wrappers/rust/icicle-cuda-runtime/src/device_context.rs
vhnatyk: trying to remove it so we only set_device explicitly

ChickenLover (dismissed):

icicle/appUtils/ntt/ntt.cu
```diff
@@ -361,13 +361,13 @@ namespace ntt {
   template <typename S>
   class Domain
   {
-    static inline int max_size = 0;
-    static inline int max_log_size = 0;
-    static inline S* twiddles = nullptr;
-    static inline std::unordered_map<S, int> coset_index = {};
+    static inline std::unordered_map<int, int> max_size_for_device = {};
```
yshekel, line 364: I think it would be cleaner to make the fields non static and then have a static map from int to Domain.

wrappers/rust/icicle-core/src/ntt/mod.rs
```diff
@@ -87,6 +100,7 @@ pub trait NTT<F: FieldImpl> {
     ) -> IcicleResult<()>;
     fn initialize_domain(primitive_root: F, ctx: &DeviceContext) -> IcicleResult<()>;
     fn get_default_ntt_config() -> NTTConfig<'static, F>;
+    fn get_default_ntt_config_for_device(device_id: usize) -> NTTConfig<'static, F>;
```
yshekel, line 103: consider adding device_id param to get_default_ntt_config() with default value to avoid having two methods

wrappers/rust/icicle-core/src/ntt/tests.rs
```diff
@@ -1,23 +1,29 @@
...
+    ntt::{get_default_ntt_config, get_default_ntt_config_for_device, initialize_domain, ntt, NTTDir, Ordering},
     traits::{ArkConvertible, FieldImpl, GenerateRandom},
 };
 
+use crate::ntt::get_default_context_for_device;
+
 use super::NTT;
 
-pub fn init_domain<F: FieldImpl + ArkConvertible>(max_size: u64)
+pub fn init_domain<F: FieldImpl + ArkConvertible>(max_size: u64, device_id: usize)
```
yshekel, line 21: consider adding default value for device_id to avoid passing 0 if having 1 GPU

vhnatyk: Rust doesn't support default values if I understand you correctly

wrappers/rust/rust-toolchain
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.70.0"
+channel = "1.72.0"
```
ChickenLover, line 2: Consider removing the whole file

ChickenLover: or downgrading it to 1.69.0

wrappers/rust/icicle-core/src/ntt/mod.rs
DmytroTym: Yea I don't like the idea of removing the old function either. But why letting the user explicitly set device here? Especially considering that we want there to be one device per one CPU thread so no explicit device changes besides setting it for the duration of a thread right? I think you mentioned offline that that was the idea. And proper device id can always be accessed by [`cudaGetDevice`](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g80861db2ce7c29b6e8055af8ae01bc78) too...
But if you insist on this scheme - let's be consistent. You created an extra config creation function that accepts a `device_id` for NTT but not for MSM, I think it should be either both of them or neither.

wrappers/rust/icicle-core/src/ntt/mod.rs
```diff
@@ -216,31 +243,30 @@ macro_rules! impl_ntt_tests {
...
         #[test]
         fn test_ntt_batch() {
-            INIT.get_or_init(move || init_domain::<$field>(MAX_SIZE));
+            INIT.get_or_init(move || init_domain::<$field>(MAX_SIZE, DEFAULT_DEVICE_ID));
             check_ntt_batch::<$field>()
         }
 
         #[test]
         fn test_ntt_device_async() {
-            INIT.get_or_init(move || init_domain::<$field>(MAX_SIZE));
```
DmytroTym, line 243: No init here?

wrappers/rust/icicle-core/src/msm/tests.rs
```diff
@@ -19,61 +22,73 @@ where
...
 
-        let msm_result_ark: ark_ec::models::short_weierstrass::Projective<C::ArkSWConfig> =
-            VariableBaseMSM::msm(&points_ark, &scalars_ark).unwrap();
-        let msm_res_affine: ark_ec::short_weierstrass::Affine<C::ArkSWConfig> = msm_host_result[0]
-            .to_ark()
-            .into();
-        assert!(msm_res_affine.is_on_curve());
-        assert_eq!(msm_host_result[0].to_ark(), msm_result_ark);
-    }
+                let mut cfg = get_default_msm_config::<C>();
```
DmytroTym, line 59: Related to the previous comment: doesn't this create config on the first device only?

wrappers/rust/icicle-cuda-runtime/src/device_context.rs
```diff
@@ -1,27 +1,42 @@
...
+pub const DEFAULT_DEVICE_ID: usize = 0;
+
+use crate::device::get_device;
+
+/// Properties of the device used in Icicle functions.
 #[repr(C)]
 #[derive(Debug, Clone)]
 pub struct DeviceContext<'a> {
-    /// Stream to use. Default value: 0.
+    /// Stream to use. Default value: 0. //TODO: multiple streams per device ?
```
DmytroTym, line 12: wdym by multiple streams per device?

DmytroTym (commented):
Really like the idea of one CPU thread = one device. Maybe we'll have some inconveniences integrating into third party applications that already have their own multithreading but stand-alone it seems very elegant and natural to me.

The only thing I don't really like is adding new functions for context and config creation that accept `device_id`. I feel like we can always just infer device id using `cudaGetDevice` especially considering that we want device id to be set once for the entire duration of the CPU thread. More generally, I feel the need to push back against all non-essential free floating functions being added to the public API. As discussed with @ChickenLover at some point, it feels like once there are too many of them - we should create a struct, a trait etc.

DmytroTym:
@vhnatyk okay last counter-proposal. Recently we moved `get_default_ntt_config` into Rust, this allowed us to actually implement it [generically](https://github.com/ingonyama-zk/icicle/blob/main/wrappers/rust/icicle-core/src/ntt/mod.rs#L65). Using this function seems to me much better than the free-floating functions as we had to before. Can we from now on use `NTTConfig::default_config` and `NTTConfig::default_config_for_device`, not add `get_default_ntt_config_for_device` and schedule `get_default_ntt_config` for deprecation? Maybe the same can be done for the `DeviceContext` struct, not sure why would I put a floating function there, for consistency with configs I guess, but now since we added another initialiser I would definitely prefer `DeviceContext::default_context` and `DeviceContext::default_context_for_device`.

wrappers/rust/rust-toolchain
yshekel: I already removed it so just rebase/merge.

wrappers/rust/icicle-core/src/ntt/mod.rs
vhnatyk: added comment

wrappers/rust/icicle-cuda-runtime/src/device_context.rs
vhnatyk: array of streams vs array of DeviceContext

icicle/appUtils/ntt/ntt.cu
vhnatyk: changed to array

examples/rust/msm/src/main.rs
```diff
@@ -91,8 +91,8 @@ fn main() {
         let mut g2_msm_results: HostOrDeviceSlice<'_, G2Projective> = HostOrDeviceSlice::cuda_malloc(1).unwrap();
         let stream = CudaStream::create().unwrap();
         let g2_stream = CudaStream::create().unwrap();
-        let mut cfg = msm::get_default_msm_config::<CurveCfg>();
-        let mut g2_cfg = msm::get_default_msm_config::<G2CurveCfg>();
+        let mut cfg = msm::MSMConfig::default();
+        let mut g2_cfg = msm::MSMConfig::default();
```
DmytroTym, line 95: Unimportant, just fyi: I think one config can be used for both msm

wrappers/rust/icicle-core/src/msm/tests.rs
```diff
@@ -1,5 +1,6 @@
-use super::{get_default_msm_config, msm, MSM};
+use super::{msm, MSM};
 use crate::curve::{Affine, Curve, Projective};
+use crate::msm::MSMConfig;
```
DmytroTym, line 3: Maybe unify imports here to either
```rust
use super::{msm, MSM, MSMConfig};
```
or
```rust
use crate::msm::{msm, MSM, MSMConfig};
```

vhnatyk: fixed

wrappers/rust/icicle-core/src/msm/tests.rs
vhnatyk: fixed

DmytroTym (approved):
lgtm

wrappers/rust/icicle-core/src/field.rs
```diff
@@ -15,6 +15,9 @@ pub struct Field<const NUM_LIMBS: usize, F: FieldConfig> {
     p: PhantomData<F>,
 }
 
+unsafe impl<const NUM_LIMBS: usize, F: FieldConfig> Send for Field<NUM_LIMBS, F> {}
```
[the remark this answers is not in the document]
vhnatyk: just tested - it covers both via field impl, point an scalars passed to other thread with this and fails compile without

yshekel (approved):

PR #357, opened 2024-01-31
Release v1.1.0

jeremyfelder:
Release v1.1.0:
- Updated examples to use the new API
- [c++] Curve specific functions using macros
- [c++] Consolidate MSM and Batch MSM to single function
- [CI] Add codespell in CI
- [FIX] Windows rust build
- [FIX] G2 on rust bindings
- [FIX] Bw6 using bls12377


examples/rust/msm/src/main.rs
```diff
@@ -1,7 +1,9 @@
 use icicle_bn254::curve::{
```
ImmanuelSegol, line 1: maybe add a batch MSM example ?
What about an example running in BM and Large triangle accumulation?

Maybe these should be under benchmarks tho not sure.

Please ignore this comment if you dont think we need this.

ImmanuelSegol, line 1: maybe add a batch MSM example ?
What about an example running in BM and Large triangle accumulation?

Maybe these should be under benchmarks tho not sure.

Please ignore this comment if you dont think we need this.

DmytroTym (approved):

examples/rust/msm/src/main.rs
svpolonsky: Is there a channel to ask our users what examples they are looking for?

PR #358, opened 2024-02-05
Fixed overflow in large coset NTTs

DmytroTym:
If domain has size 2^17, NTT on size 2^16 coset generated by `-1` fails. This happens due to index in `BatchMulKernel` overflowing, fixed by using `long` instead of `int`.


icicle/utils/utils_kernels.cuh
```diff
@@ -34,7 +34,7 @@ namespace utils_internal {
   {
     int tid = blockDim.x * blockIdx.x + threadIdx.x;
     if (tid < n_elements * batch_size) {
-      int scalar_id = tid % n_elements;
+      long scalar_id = tid % n_elements;
```
yshekel, line 37: I would be more specific and use uint32_t or uint64_t.

DmytroTym: Changed to int64_t as we still need negative indices for inverse NTT here

yshekel (approved):

PR #367, opened 2024-02-08
Mixed-radix NTT batch support

yshekel:


icicle/appUtils/ntt/ntt.cu
```diff
@@ -501,11 +501,11 @@ namespace ntt {
       h_coset.clear();
     }
 
-    const bool is_small_ntt = logn < 16;                  // cutoff point where mixed-radix is faster than radix-2
+    // (heuristic) cutoff point where mixed-radix is faster than radix-2
+    const bool is_small_ntt = (logn < 16) && ((size_t)size * batch_size < (1 << 20));
```
HadarIngonyama, line 505: why not have a table? It's only 16 lines

HadarIngonyama (changes requested):

icicle/appUtils/ntt/ntt.cu
yshekel: I will do that and later extend the table to account for orderings too.

yshekel: update: I will do it in the PR where adding orderings instead of doing it twice.

icicle/appUtils/ntt/kernel_ntt.cu
```diff
@@ -34,26 +34,30 @@ namespace ntt {
...
       next_element = dig_rev(next_element, log_size, dit);
       if (next_element < idx) return; // not handling this group
       if (next_element == idx) break; // calculated whole group
-      group[i++] = next_element;
+      group[i++] = next_element + size * batch_idx;
     }
 
-    if (i == 1) { // single element in group --> nothing to do (except maybe normalize for INTT)
-      if (is_normalize) { arr[idx] = arr[idx] * inverse_N; }
+    if (i == 1) { // single element in group --> nothing to do (except normalize for INTT)
```
DmytroTym, line 59: I think this `if` can be removed and the remainder of the function will deal with `i == 1` case correctly. This might even improve performance a tiny bit due to decreased thread divergence

icicle/appUtils/ntt/kernel_ntt.cu
```diff
@@ -462,81 +482,89 @@ namespace ntt {
     }
 
     if (log_size == 4) {
+      const int NOF_BLOCKS = (2 * batch_size + 64 - 1) / 64;
+      const int NOF_THREADS = min(64, 2 * batch_size);
```
DmytroTym, line 486: Small remark: imo
```c++
const int NOF_THREADS = min(64, 2 * batch_size);
const int NOF_BLOCKS = (2 * batch_size + NOF_THREADS - 1) / NOF_THREADS;
```
is better as it allows to change `NOF_THREADS` without accidentally breaking `NOF_BLOCKS` in the process. Repeats in several other places too.

icicle/appUtils/ntt/kernel_ntt.cu
yshekel: you are right

icicle/appUtils/ntt/kernel_ntt.cu
yshekel: fixed

DmytroTym (approved):
lgtm

HadarIngonyama (approved):

PR #368, opened 2024-02-08
Mixed radix NTT coset support

yshekel:


DmytroTym (approved):
Coset tests in Rust currently don't check cosets but once ordering PR is merged they will (as far as I understood) so all is good imo

PR #371, opened 2024-02-11
Mixed-radix NTT support all orderings

yshekel:


wrappers/rust/icicle-core/src/ntt/mod.rs
```diff
@@ -38,6 +49,22 @@ pub enum Ordering {
     kNR,
     kRN,
     kRR,
+    kNM,
+    kMN,
+}
+
+///Which NTT algorithm to use. options are:
+///- Auto: implementation selects autoamtically based on heuristic
```
DmytroTym, line 57: autoamtically -> automatically

DmytroTym: Also maybe worth mentioning that Auto is a good default choice and should be used in most cases?

wrappers/rust/icicle-core/src/ntt/mod.rs
```diff
@@ -30,6 +30,17 @@ pub enum NTTDir {
 /// a_4, a_2, a_6, a_1, a_5, a_3, a_7`.
 /// - kRN: inputs are bit-reversed-order and outputs are natural-order.
 /// - kRR: inputs and outputs are bit-reversed-order.
+///
+/// Mixed-Radix NTT: digit-reversal is a generalization of bit-reversal where the latter is a private case with 1b
```
DmytroTym, line 34: private case? partial case probably?

DmytroTym: typos are the same in CUDA-side doc comment

icicle/appUtils/ntt/tests/verification.cu
```diff
@@ -134,7 +137,7 @@ int main(int argc, char** argv)
   };
 
   CHK_IF_RETURN(benchmark(false /*=print*/, 1)); // warmup
-  int count = INPLACE ? 1 : 10;
+  int count = INPLACE ? 1 : 1;
```
DmytroTym, line 140: 1 or 1 :)

icicle/appUtils/ntt/ntt.cu
```diff
@@ -514,6 +534,7 @@ namespace ntt {
         reverse_input = true;
         break;
       case Ordering::kNR:
+      case Ordering::kNM:
```
DmytroTym, line 537: Don't see `kMN` among the cases. Also kind of weird that ordering is done here for radix 2 but one function deeper for mixed radix, maybe we can do some unification here? (speaking of, `dit` in mixed radix version is the same as `ct_butterfly` in radix2, just difference in lingo)

DmytroTym (commented):
One issue I have with this PR is that it's possible to force old radix-2 algorithm with an ordering like `NM` or `MN`. It might seem like `test_ntt_batch` checks correctness of it but it only compares batch and non-batch outputs of the same algorithm. While the outputs for radix-2 are consistent, they are not correct. Correct me if I'm wrong here. In terms of how to deal with `NM` and `MN` orderings in radix-2 NTT - I'm fine with a runtime error tbh cos it doesn't make a lot of sense.

About Rust NTT tests - they now take a pretty long time. One idea might be to put looping over the two algorithms as deeply as possible instead of at the very beginning of the test. This should reduce the CPU side work which is where the vast majority of test time goes into. Also it should make diffs much smaller and nicer to look at.

wrappers/rust/icicle-core/src/ntt/mod.rs
HadarIngonyama: I guess it should be special case

HadarIngonyama (commented):
all in all looks fine, just wanted to ask - did you make sure that the changes to the reordering kernels didn't harm the performance of the simple cases (I don't think the should, but just to verify)

yshekel:
> One issue I have with this PR is that it's possible to force old radix-2 algorithm with an ordering like `NM` or `MN`. It might seem like `test_ntt_batch` checks correctness of it but it only compares batch and non-batch outputs of the same algorithm. While the outputs for radix-2 are consistent, they are not correct. Correct me if I'm wrong here. In terms of how to deal with `NM` and `MN` orderings in radix-2 NTT - I'm fine with a runtime error tbh cos it doesn't make a lot of sense.
> 
> About Rust NTT tests - they now take a pretty long time. One idea might be to put looping over the two algorithms as deeply as possible instead of at the very beginning of the test. This should reduce the CPU side work which is where the vast majority of test time goes into. Also it should make diffs much smaller and nicer to look at.



> One issue I have with this PR is that it's possible to force old radix-2 algorithm with an ordering like `NM` or `MN`. It might seem like `test_ntt_batch` checks correctness of it but it only compares batch and non-batch outputs of the same algorithm. While the outputs for radix-2 are consistent, they are not correct. Correct me if I'm wrong here. In terms of how to deal with `NM` and `MN` orderings in radix-2 NTT - I'm fine with a runtime error tbh cos it doesn't make a lot of sense.
> 
> About Rust NTT tests - they now take a pretty long time. One idea might be to put looping over the two algorithms as deeply as possible instead of at the very beginning of the test. This should reduce the CPU side work which is where the vast majority of test time goes into. Also it should make diffs much smaller and nicer to look at.

Regarding NM, MN for radix-2: The way I am thinking about it is as follows: M (=Mixed) is **by definition the order of the elements that a DIF NTT/INTT would produce given Natural inputs**. DIT NTT/INTT of the same size expects this same order to output the Natural order back. Different NTT sizes would mix the elements differently but **consistently per size**. For element-wise ops like multiplication this is useful but again the only thing guaranteed here is consistency per size. 
Therefore I am thinking of radix-2 R as a special case of M which means that NR an NM are the same for radix-2 (and same for RN and MN). I don't see a reason why this is invalid to do or should throw an assertion. Technically a radix2 NTT can be plugged to the mixed-radix algorithm and the top-level table would be all ones in the logn size, such that digit-reversal is exactly bit-reversal and so we have logn stages, each composed of N/2 NTTS of size 2.

You are right that the batch test is only testing that batch computation is equivalent to a simple loop over the elements with btach=1 for each.

User who want to specifically get R should specify R, not M and this is a stronger request. For radix-2 it's free but for mixed-radix alg it's more expensive and therefore if not required (like for poly multiplication) then M is good enough.

Regarding rust tests, TBH I did not think about it too much and thought that most of the time is building but yes you are right I will try to nest the loop and avoid reallocation of arrays etc. The small downside of that is that in-place computations (when/if we add them) may fail or behave different then expected although since we randomize inputs it's maybe not really a problem.

wrappers/rust/icicle-core/src/ntt/tests.rs
```diff
@@ -50,34 +50,38 @@ where
 {
     let test_sizes = [1 << 4, 1 << 17];
     for test_size in test_sizes {
-        let ark_domain = GeneralEvaluationDomain::<F::ArkEquivalent>::new(test_size).unwrap();
+        for alg in [NttAlgorithm::Radix2, NttAlgorithm::MixedRadix] {
+            let ark_domain: GeneralEvaluationDomain<<F as ArkConvertible>::ArkEquivalent> =
```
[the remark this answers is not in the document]
yshekel: Probably rust-analyzer did it for me by mistake. Generally I think type annotations are a good thing but in this case I absolutely agree that it is redundant to annotate since the constructor is clearly describing the type.
I will remove it. Did not intend to add that.

wrappers/rust/icicle-core/src/ntt/mod.rs
yshekel: Thanks. Will fix typo and mention that.

wrappers/rust/icicle-core/src/ntt/mod.rs
yshekel: I commented about that already but I did mean that digit-reversal is the more general case and bit-reversal is a special (or private) case with 1b digits. Why is that wrong?

icicle/appUtils/ntt/tests/verification.cu
yshekel: oops

icicle/appUtils/ntt/ntt.cu
yshekel: kMN is not there for the exact same reason kRN isn't. The code was like that so I did not modify it but basically the default values of ct_butterfly and reverse_input are for those cases already. Probably cleaner to add those cases and not rely on the default values. I will do that.

About consolidation, there is asymmetry between the two algs where radix-2 is implemented in ntt.cu while mixed-radix is not. Ideally they are each in a different file and may share common logic but I avoided refactoring too much to avoid rebase problems for other PRs (like the multi-gpu that is also modifying ntt.cu).

Regarding this switch case specifically, I can try to consolidate but I am afraid it would cause more confusion than good and that's why I did not. Note that calculating the input/output/coset reorder and dit/dif bit and passing it to mixed-radix NTT seems worse to me compared to passing the ordering field and let the specific implementation decide how it emulates R etc. Also note that except for dit (which is ct_buttefrly ineed), radix2 doesn't care about output/coset reorder the way mixed-radix does and even input reorder is not always the same (mainly R emulation requires reorder for mixed-radix but not for radix2).
If you think it's important I will try to see how it can be cleanly done but my feeling is that this is not a good idea unless refactoring this entire kind of "strategy-pattern" more.

yshekel:
> all in all looks fine, just wanted to ask - did you make sure that the changes to the reordering kernels didn't harm the performance of the simple cases (I don't think the should, but just to verify)

@HadarIngonyama do you mean that NN for example is not slower now? I did not verify since I did not think there should be any difference at all but I will verify just to be 100% sure.

wrappers/rust/icicle-core/src/ntt/mod.rs
DmytroTym: Oh okay. Just unfamiliar usage of this word for me, assumed it was autocorrect or sth

yshekel: I guess I mistranslate from hebrew but special-case is the one I meant. I will fix.

icicle/appUtils/ntt/ntt.cu
yshekel: I did not consolidate the switch cases since I think it will do more bad than good but I did split the radix-2 implementation so it's more symmetric now.

DmytroTym: Definitely understand not wanting to refactor the new NTT and agree with _passing the ordering field and let the specific implementation decide how it emulates R etc_. Maybe then we can move reversals inside and rename ct_butterfly to dit in the old NTT?

DmytroTym (approved):

PR #372, opened 2024-02-11
MSM improvements

DmytroTym:
## Describe the changes

MSM can now handle zero base points. They are represented as affine points with `x` and `y` coordinates equal to zero which is (as far as I know) consistent with gnark and rapidsnark but not arkworks. Rust tests are changed accordingly.

A number of performance and memory improvements have been made:

1. Kernel launches, allocations and frees are moved around to minimise memory footprint and better parallelise copying bases from host to device and sorting scalars. Still, compute often has to wait for copying bases because it takes significantly more time than scalar sorting. The way to solve it in the future is probably doing what Matter Labs are doing - computing MSM in chunks that would allow masking uploading the next chunk of base points with bucket accumulation from the previous chunk.
2. Speaking of sorting scalars, instead of sorting indices for each bucket module individually, it's now done for all indices at once. While this requires a bit more memory and theoretically should take more operations that one-by-one approach, it turns out to be faster in practice. Not just for smaller MSM but for large ones as well. This also allows to automatically remove zero buckets.
3. Removing zero buckets allows us to avoid using them in reduction. Though this makes the code more convoluted and doesn't help too much for large MSMs, it boosts small MSMs quite a bit.
4. For large buckets, I increased the number of threads per bucket by a factor of `large_bucket_size` to make each thread in large bucket accumulation do the same amount of work as expected in normal accumulation. At the same time, this worsens potential memory bottleneck in large bucket accumulation because the amount of memory allocated [here](https://github.com/ingonyama-zk/icicle/compare/dev...develop/dima/msm_improvements?expand=1#diff-ae66125889f097df37478ec63588d85f130e079c8822d336bbcf14aa04cdb0a7L600) in the old version is proportional to the number of threads in the largest bucket times the number of large buckets. So if the largest bucket is really large and there are lots of much smaller large buckets, we might run out of memory. So I allocated only as much memory as necessary for each large bucket, depending on its size. This complicates the code but I think the speedup and memory savings are worth it.

## Benchmarks

Measurements are made on an RTX 3090Ti card for the bn254 curve, H2D memory operations not included.

In the first experiment, ~30% of scalars are equal to 1, there are also 10 random scalars each with frequency around 1%. The rest of scalars are chosen uniformly at random. In the second experiment, all scalars are chosen uniformly at random.
| \# | MSM size | Batch size | Old version, ms.  | New version, ms. |
| -- | --------- | ----------- | ----------------- | ----------------- |
| 1 | `2^22` | 1 | 33.8  | 29.8  |
| 1 | `2^22` | 3 | 68.2  | 56.3  |
| 2 | `2^24` | 1 | 151.5  | 149.8  |
| 2 | `2^14` | `2^8` | 79.5  | 73.1  |

## Failures and future work

I spent quite a bit of time trying to make reduction and scan methods from CUB and thrust work with our point types. I thought that it would be nice to let well-optimised CUDA libraries handle load balancing in bucket accumulation for us. As it turns out, there are many issues with this approach:

- It seems that CUB and thrust are optimised for minimising memory movement rather than compute but our EC addition is very much compute bound;
- Compile time grows a lot when trying to use our primitives inside CUB/thrust functions;
- I wasn't even able to get correctness, results seem to always be zero or random, no idea why.

Overall I think this is a dead end and it's not worth trying to swap our custom bucket accumulation to anything CUB or thrust provide nowadays.

In terms of future work, arbitrary choice of `c` is still not supported as of this PR (though I think I can implement it pretty quickly) and a bunch of other improvements like xyzz accumulation, signed digits/scalars and computing MSM in chunks as described earlier are still in TODO status.


wrappers/rust/icicle-core/src/curve.rs
```diff
@@ -147,13 +149,18 @@ where
...
     fn to_ark(&self) -> Self::ArkEquivalent {
-        let proj_x = self
-            .x
-            .to_ark();
-        let proj_y = self
-            .y
-            .to_ark();
-        Self::ArkEquivalent::new_unchecked(proj_x, proj_y)
+        if *self == Self::zero() {
+            Self::ArkEquivalent::zero()
```
yshekel, line 153: is it more efficient than the else case?

wrappers/rust/icicle-core/src/msm/tests.rs
```diff
@@ -24,19 +24,17 @@ where
     for test_size in test_sizes {
         let mut points = C::generate_random_affine_points(test_size);
         let scalars = <C::ScalarField as FieldImpl>::Config::generate_random(test_size);
-        let mut points_ark: Vec<_> = points
+        // sprinkle in some zero points
```
yshekel, line 27: same comment for the zeros

wrappers/rust/icicle-core/src/msm/tests.rs
```diff
@@ -168,14 +166,20 @@ where
         for batch_size in batch_sizes {
             let mut points = C::generate_random_affine_points(test_size * batch_size);
             let mut scalars = vec![C::ScalarField::zero(); test_size * batch_size];
+
+            // add some zero points
+            for _ in 0..100 {
```
yshekel, line 171: same comment for the zeros

icicle/appUtils/msm/msm.cu
```diff
@@ -31,30 +31,71 @@ namespace msm {
...
+      unsigned nof_large_buckets,
+      unsigned log_nof_large_buckets,
+      unsigned* bucket_indices)
+    {
+      const int tid = blockIdx.x * blockDim.x + threadIdx.x;
+      if (tid >= nof_large_buckets) { return; }
+      unsigned start = (sorted_bucket_sizes_sum[tid] + nof_pts_per_thread - 1) / nof_pts_per_thread + tid;
+      unsigned end = (sorted_bucket_sizes_sum[tid + 1] + nof_pts_per_thread - 1) / nof_pts_per_thread + tid + 1;
+      for (unsigned i = start; i < end; i++) {
+        bucket_indices[i] = tid | ((i - start) << log_nof_large_buckets);
```
yshekel, line 54: are you sure this is correct when nof_buckets is not a power of two?

wrappers/rust/icicle-core/src/curve.rs
DmytroTym: The else case just doesn't cover zero. `new_unchecked` assumes that inputs represent a valid non-zero point.

wrappers/rust/icicle-core/src/msm/tests.rs
DmytroTym: On the one hand we can
On the other, I don't think users of a function called `generate_random_affine_points` expect zero points being sprinkled in. We're not promising cryptographic rng here or anything and this function shouldn't be used to create secure randomness but still...

icicle/appUtils/msm/msm.cu
DmytroTym: It should be correct for non-powers-of-2 and in most cases `nof_large_buckets` is not a power of 2
What this line does is just packing two values into one number `bucket_indices[i]`. `tid` goes into the lowest `log_nof_large_buckets` bits and `i - start` goes into the rest. So `log_nof_large_buckets` in this case just means the number of bits needed to represent `tid` which varies from 0 to `nof_large_buckets`

yshekel: right, as long as log_nof_large_buckets = ceil(log2(nof_large_buckets)) this is correct. Since it's not verified inside the kernel, it could mix the two fields so that's why I suggested to make verify it at kernel launch.

wrappers/rust/icicle-core/src/msm/tests.rs
yshekel: Ok, I just thought that if you need it more than once it's worth writing it once. You could wrap this function is a test util that accepts the probability of each point to be zeroed but you should decide if it makes sense to you. If not that's fine too.

DmytroTym: Yes, creating a separate test util function makes sense, will do

icicle/appUtils/msm/msm.cu
DmytroTym: In principle we can just take log inside the kernel, I just wanted to avoid all threads doing identical work. Verification is definitely cheaper though, can do it inside the kernel if you want

yshekel: My motivation is to avoid debugging cases like nof_buckets=100, (int)log2(nof_buckets)=6.

I would personally add a comment about this assumption next to the kernel param and make sure the host is calling correctly. If you prefer assert inside the kernel it's fine too.

DmytroTym: Added a comment though maybe I should've done a more full doc comment for each kernel (but they are internal so I don't feel the need to spend too much time documenting them tbh)

icicle/appUtils/msm/msm.cu
```diff
@@ -450,74 +441,105 @@ namespace msm {
...
+      CHK_IF_RETURN(cudaMemcpyAsync(
+        &smallest_bucket_index, single_bucket_indices, sizeof(unsigned), cudaMemcpyDeviceToHost, stream));
+      unsigned zero_bucket_offset = (smallest_bucket_index == 0) ? 1 : 0;
+
       // sort by bucket sizes
       unsigned h_nof_buckets_to_compute;
       CHK_IF_RETURN(cudaMemcpyAsync(
         &h_nof_buckets_to_compute, nof_buckets_to_compute, sizeof(unsigned), cudaMemcpyDeviceToHost, stream));
-
-      // if all points are 0 just return point 0
```
DmytroTym, line 504: There was a question about removing this block. Turned out it was there for a good reason for this block to be here. If all the scalars are zero, there's a division by zero error inside. I will push a more elegant fix for this tomorrow.

icicle/appUtils/msm/msm.cu
```diff
@@ -226,43 +253,38 @@ namespace msm {
...
+      unsigned* __restrict__ large_bucket_thread_indices,
+      unsigned* __restrict__ point_indices,
       A* __restrict__ points,
-      const unsigned nof_buckets,
       const unsigned nof_buckets_to_compute,
-      const unsigned msm_idx_shift,
       const unsigned c,
-      const unsigned threads_per_bucket,
-      const unsigned max_run_length)
+      const int points_per_thread,
```
LeonHibnik, line 270: why not unsigned?

yshekel (approved):
honestly I don't fully understand the details of the large bucket accumulation, but overall looks good to me.
I approve but you may want Hadar to review too.

LeonHibnik (approved):
lgtm

PR #374, opened 2024-02-14
fix: verify NTT size is a power of two

yshekel:


icicle/appUtils/ntt/ntt.cu
```diff
@@ -526,16 +526,22 @@ namespace ntt {
   cudaError_t NTT(E* input, int size, NTTDir dir, NTTConfig<S>& config, E* output)
   {
     CHK_INIT_IF_RETURN();
+
     if (size > Domain<S>::max_size) {
-      std::cerr
-        << "NTT size is too large for the domain. Consider generating your domain with a higher order root of unity"
-        << '\n';
-      throw -1;
+      std::cerr << "[ERROR] NTT size=" << size
```
vhnatyk, line 531: use         THROW_ICICLE_ERR(
          IcicleError_t::InvalidArgument, "error message

icicle/appUtils/ntt/ntt.cu
```diff
@@ -526,16 +526,22 @@ namespace ntt {
...
+                << " is too large for the domain. Consider generating your domain with a higher order root of unity.";
+      return cudaErrorInvalidValue;
+    }
+
+    int logn = int(log2(size));
+    const bool is_size_power_of_two = size == (1 << logn);
+    if (!is_size_power_of_two) {
```
vhnatyk, line 538: same here

icicle/appUtils/ntt/ntt.cu
yshekel: I was not sure since we don't catch it and there is no real reason to crash user app.
Do you think it's better to throw exceptions?

DmytroTym (approved):

PR #382, opened 2024-02-18
feature: mixed-radix NTT fast twiddles mode

yshekel:
This mode is allocating 4N additional twiddle-factors (where N is max NTT size) for faster compute kernels.


icicle/appUtils/ntt/kernel_ntt.cu
```diff
@@ -116,7 +123,7 @@ namespace ntt {
     int tid = blockDim.x * blockIdx.x + threadIdx.x;
     if (tid >= n_elements * batch_size) return;
     int64_t scalar_id = tid % n_elements;
-    if (rev_type != eRevType::None) scalar_id = generalized_rev(tid, logn, dit, rev_type);
+    if (rev_type != eRevType::None) scalar_id = generalized_rev(tid, logn, dit, false, rev_type);
```
HadarIngonyama, line 126: shouldn't we be passing fast_tw as a parameter here?

yshekel: it's the kernel that handles cosets but it doesn't work with the fast-twiddles. That's why I removed the parameter and pass false. Otherwise it would calculate incorrectly.

icicle/appUtils/ntt/ntt.cuh
```diff
@@ -32,10 +32,13 @@ namespace ntt {
    * @param primitive_root Primitive root in field `S` of order \f$ 2^s \f$. This should be the smallest power-of-2
    * order that's large enough to support any NTT you might want to perform.
    * @param ctx Details related to the device such as its id and stream id.
+   * @param fast_twiddles_mode A mode where more memory is allocated for twiddle factors in exchange for faster compute.
+   * In this mode need additional 4N memory when N is the largest NTT size to be supported (which is derived by the
+   * primitive_root.
```
DmytroTym, line 37: bracket not closed here

icicle/appUtils/ntt/kernel_ntt.cu
```diff
@@ -501,6 +562,100 @@ namespace ntt {
     return CHK_LAST();
   }
 
+  // Fast-twiddles: 2N twiddles for forward, 2N for inverse
+  template <typename S>
+  __global__ void generate_basic_twiddles_fast_twiddles_mode(S basic_root, S* basic_twiddles)
+  {
+    S w0 = basic_root * basic_root;
+    S w1 = (basic_root + w0 * basic_root) * S::inv_log_size(1);
```
DmytroTym, line 570: I don't know if the runtime of this function is at all important but I think this line can be changed to
```c++
S w1 = S::div2(basic_root + w0 * basic_root);
```
which is also more readable imo. The same applies to the next line and the `generate_basic_twiddles_generic` method.

icicle/appUtils/ntt/thread_ntt.cu
```diff
@@ -13,19 +13,33 @@ struct stage_metadata {
...
+      {6, 6, 6, 6, 6},                                                                                                 \
+  }
+uint32_t constexpr STAGE_SIZES_HOST[31][5] = STAGE_SIZES_DATA;
+__device__ constexpr uint32_t STAGE_SIZES_DEVICE[31][5] = STAGE_SIZES_DATA;
+
+// construction for fast-twiddles
+uint32_t constexpr STAGE_PREV_SIZES[31] = {0,  0,  0,  0,  0,  0,  0,  0,  4,  5,  5,  6,  6,  9,  9, 10,
+                                           11, 11, 12, 15, 15, 16, 16, 18, 18, 20, 21, 21, 22, 23, 24};
+
+#define STAGE_SIZES_DATA_FAST_TW                                                                                       \
```
DmytroTym, line 32: Is it important that some stages are different for fast and normal twiddles?

DmytroTym (approved):
Overall, looks good. Minor comments, can be ignored

icicle/appUtils/ntt/thread_ntt.cu
HadarIngonyama: very important:) for fast twiddles it's important that all the sizes are part of the same tree. for regular this isn't important so the symmetric reordering prevails

icicle/appUtils/ntt/kernel_ntt.cu
HadarIngonyama: wasn't aware of that function, could be fine too

yshekel: are you sure about that? It computes incorrectly.

DmytroTym: Ohh sorry, `div2` only works for even inputs currently. I can make it work for odd inputs too ofc if it's important to make this function faster (though it seems that it only gets called a couple of times during twiddle generation). Otherwise let's leave it as it is.

PR #395, opened 2024-02-22
Devmode to Reduce compilation time (including G2 and ECNTT)

### LINKED ISSUE #316: Rust compilation time
### Discussed in https://github.com/ingonyama-zk/icicle/discussions/314

<div type='discussions-op-text'>

<sup>Originally posted by **sss1h** December 20, 2023</sup>
The Rust compilation time is about 15 minutes after changing the C++ functions.  Is there any way to speed up?</div>

vhnatyk:
## Describe the changes

Introduces 'develop' mode - with reduced -O0 optimizations instead of -O3 and disabled inlining and loop unrolling 
with slightly lower performance but much reduced build time (6 minutes  vs reported 163 minutes on i9 13900K? for all curves) 

## Linked Issues

Resolves #316


DmytroTym:
If we just disable the mixed-radix EC NTT and only use radix-2, would that give a further significant speedup?

vhnatyk:
umm - like still ~80%  compilation time it's msm.cu g2 - also weird C++ gtests fail but rust all passing, and didn't expect that `unroll` commenting to improve times - hoped `#pragma unroll 1` was supposed to but it even increases the time by 10%. Maybe enable g2 and new ecntt only on final merge to main

yshekel:
I don't think mixed-radix NTT is compiled for ECNTT here since it's not instantiated.
The following commit is instantiating it:
https://github.com/ingonyama-zk/icicle/commit/e1ff687ac31fc0507d08b57dffe3864b9ed8eba6

vhnatyk:
oh right - I forgot to also push merged e1ff687 I tested locally

icicle/appUtils/ntt/kernel_ntt.cu
```diff
@@ -1025,9 +1025,27 @@ namespace ntt {
...
     bool is_inverse,
     bool fast_tw,
-    Ordering ordering,
+    ntt::Ordering ordering,
     curve_config::scalar_t* arbitrary_coset,
     int coset_gen_index,
     cudaStream_t cuda_stream);
 
-} // namespace ntt
+#if defined(ECNTT_DEFINED)
```
LeonHibnik, line 1033: Comment this whole ifdef for now

vhnatyk: ok, btw - it compiles fast in .cu without Rust or Go

LeonHibnik: We agreed for the time being to remove mixed_radix from ecntt

icicle/appUtils/ntt/kernel_ntt.cu
```diff
@@ -1025,9 +1025,27 @@ namespace ntt {
...
     bool is_inverse,
     bool fast_tw,
-    Ordering ordering,
+    ntt::Ordering ordering,
     curve_config::scalar_t* arbitrary_coset,
     int coset_gen_index,
     cudaStream_t cuda_stream);
 
-} // namespace ntt
+  // #if defined(ECNTT_DEFINED)
```
yshekel, line 1033: remove commented code

icicle/appUtils/ntt/kernel_ntt.cu
```diff
@@ -2,9 +2,9 @@
 #include "appUtils/ntt/thread_ntt.cu"
 #include "curves/curve_config.cuh"
 #include "utils/sharedmem.cuh"
-#include "appUtils/ntt/ntt.cuh" // for Ordering
+#include "appUtils/ntt/ntt.cuh" // for ntt::Ordering
 
-namespace ntt {
+namespace mxntt {
```
yshekel, line 7: why different namespace?

icicle/appUtils/ntt/tests/verification.cu
```diff
@@ -12,8 +12,14 @@
 #include "ntt/ntt_impl.cuh"
 #include <memory>
 
+#ifdef ECNTT_DEFINED
```
yshekel, line 15: note that since you don't compile mixed-radix ECNTT, then I think you should remove it from here and the makefile too since it will compare radix-2 ECNTT to itself which is not very useful

icicle/common.cuh
```diff
@@ -0,0 +1,11 @@
+#if defined(DEVMODE) || defined(DEBUG)
+#define INLINE_MACRO
+#define UNROLL // #pragma unroll
```
yshekel, line 3: remove comment?

icicle/CMakeLists.txt
```diff
@@ -92,6 +92,10 @@ if (NOT IS_CURVE_SUPPORTED)
   message( FATAL_ERROR "The value of CURVE variable: ${CURVE} is not one of the supported curves: ${SUPPORTED_CURVES}" )
 endif ()
 
+if (DEVMODE STREQUAL "ON")
```
yshekel, line 95: not big deal but note that it is recommended to use cmake-option compared to strings since it's a boolean value. It's cleaner since it can take only two values and not any string but also because you declare options on top and it's easier to understand what options are there.

I can see other values like G2_DEFINED are the same but maybe worth making them all boolean options, unless the rust build.rs scripts cannot use that, which I guess is not the case but not sure about that.

LeonHibnik: @ChickenLover TODO in v2

icicle/appUtils/ntt/kernel_ntt.cu
vhnatyk: otherwise  it's linking issue similar to the one when annonymous namespaces

icicle/appUtils/ntt/kernel_ntt.cu
vhnatyk: left commented as requested by @LeonHibnik

icicle/appUtils/ntt/tests/verification.cu
vhnatyk: actually with devmode - the mixed radix ecntt compilation time becomes feasible - so we may reintroduce it due to performance reasons

icicle/appUtils/ntt/kernel_ntt.cu
yshekel: consider writing a comment explaining why it is commented out

yshekel (approved):
Approving based on the answers. Note that I wrote a minor comment about the commented mixed-radix ECTT. You can ignore it if you want.

LeonHibnik (approved):
TODO add devmode to golang

PR #400, opened 2024-02-26
[BREAKING] Change golang field representation to uint64

jeremyfelder:
## Describe the changes

This PR changes the field representation in Golang bindings to use uint64 limbs instead of uint32 for easier conversion with other libraries


wrappers/golang/internal/generator/main.go
```diff
@@ -46,33 +46,33 @@ var bn254 = CurveData{
 	PackageName:    "bn254",
 	Curve:          "bn254",
 	GnarkImport:    "bn254",
-	ScalarLimbsNum: 8,
-	BaseLimbsNum:   8,
-	G2BaseLimbsNum: 16,
+	ScalarLimbsNum: 4,
+	BaseLimbsNum:   4,
+	G2BaseLimbsNum: 8,
```
DmytroTym, line 51: Just wondering: can this `CurveData` struct accommodate curves without G2 (like grumpkin)?

DmytroTym:
Overall I'm not super good at golang or our golang bindings, but the PR looks pretty harmless to me and I can approve if needed.

wrappers/golang/internal/generator/main.go
jeremyfelder: Currently no. The rest of the generation script for the curves also doesn't handle this case.
I think the entire generation script here needs to be a bit more modular in general and especially to support curves that don't implement all the functionality

DmytroTym: Right, cos grumpkin and some other curves also don't implement NTT (Nova doesn't need it) and we'll get to small fields hopefully soon where you don't have curves at all. But all that is outside of the scope of this PR ofc, just wanted to mention

wrappers/golang/core/internal/curve.go
```diff
@@ -8,7 +8,7 @@ func (p MockProjective) Size() int {
 	return p.X.Size() * 3
 }
 
-func (p MockProjective) AsPointer() *uint32 {
```
ImmanuelSegol, line 11: looks good to me, this was a long awaited change

ImmanuelSegol (approved):

ImmanuelSegol (approved):
Only question I have is how do we roll this out since its going to be a breaking change for many integrations.

wrappers/golang/core/internal/field.go
```diff
@@ -6,30 +6,30 @@ import (
 )
 
 const (
-	BASE_LIMBS int8 = 8
+	BASE_LIMBS int = 4
```
bigsky77, line 9: Limbs sum to 256 bytes.

8 * 32 = 256 
and 
4 * 64 = 256

Is this correct?

bigsky77:
This shouldn't break any current integrations, all the changes make sense, and the tests pass.  Approve.

bigsky77 (approved):

wrappers/golang/core/internal/helpers_test.go
```diff
@@ -4,27 +4,27 @@ import (
...
-	baseBytes := []byte{1, 2, 3, 4}
+func generateBytesArray(size int) ([]byte, []uint64) {
+	baseBytes := []byte{1, 2, 3, 4, 5, 6, 7, 8}
 	var bytes []byte
-	var limbs []uint32
+	var limbs []uint64
 	for i := 0; i < size; i++ {
 		bytes = append(bytes, baseBytes...)
-		limbs = append(limbs, 67305985)
+		limbs = append(limbs, 578437695752307201)
```
nonam3e, line 27: what is this number for?

nonam3e (approved):

wrappers/golang/core/internal/helpers_test.go
jeremyfelder: Its the decimal representation of baseBytes interpreted as a uint64

PR #402, opened 2024-02-28
Vec ops docs

ImmanuelSegol:
## Describe the changes

This PR...

## Linked Issues

Resolves #


docs/docs/icicle/rust-bindings/vec-ops.md
```diff
@@ -0,0 +1,143 @@
...
+}
+```
+
+#### Methods
+
+- **`add`**: Computes the element-wise sum of two vectors.
+- **`sub`**: Computes the element-wise difference between two vectors.
+- **`mul`**: Performs element-wise multiplication of two vectors.
+
+### Argument Validation
```
ChickenLover, line 108: Remove this part:) `check_vec_ops_args` is just for internal use, and the user doesn't need to check anything before calling the methods

ChickenLover (changes requested):
waiting for rust examples

docs/docs/icicle/rust-bindings/vec-ops.md
ImmanuelSegol: I wanted to explain that we implemented this for them, so developers wont think they need to implement it themselves

bigsky77 (approved):

docs/docs/icicle/rust-bindings/vec-ops.md
ChickenLover: Inputs validation is the default thing and we do it in all of our primitives, so I don't think we should mention it in the docs

PR #412, opened 2024-03-01
[BREAKING] Rust representation of GPU memory re-worked

DmytroTym:


wrappers/rust/icicle-cuda-runtime/src/memory.rs
```diff
@@ -244,6 +264,16 @@ impl<T, const D_ID: usize> DeviceVec<T, D_ID> {
...
+
+    pub fn cuda_malloc_for_device(count: usize, device_id: usize) -> CudaResult<Self> {
+        check_device(device_id);
+        Self::cuda_malloc(count)
+    }
+
+    pub fn cuda_malloc_async_for_device(count: usize, stream: &CudaStream, device_id: usize) -> CudaResult<Self> {
+        check_device(device_id);
+        Self::cuda_malloc_async(count, stream)
+    }
```
jeremyfelder, line 276: Whats the use case here for passing a specific deviceId if we require it to match the current device's id?

wrappers/rust/icicle-cuda-runtime/src/memory.rs
```diff
@@ -205,9 +227,8 @@ impl<T, const D_ID: usize> DeviceSlice<T, D_ID> {
     }
 }
 
-impl<T, const D_ID: usize> DeviceVec<T, D_ID> {
+impl<T> DeviceVec<T> {
     pub fn cuda_malloc(count: usize) -> CudaResult<Self> {
```
jeremyfelder, line 231: Should this and `cuda_malloc_async` be private now that we have `cuda_malloc_for_device<_async>`?

wrappers/rust/icicle-cuda-runtime/src/memory.rs
DmytroTym: This version is preferred by @vhnatyk over implicit choice of device id in malloc.

wrappers/rust/icicle-cuda-runtime/src/memory.rs
DmytroTym: I don't know... If you're only using one card, there's no need in the `for_device` version of these methods. I personally would use `cuda_malloc_async` in multi-device settings too.

wrappers/rust/icicle-cuda-runtime/src/memory.rs
vhnatyk: tl;dr - not that it's preferred but seems aligned with everything else. Longer version: the check ensures we maintain correct device_id through our calls - since we don't have implicit `device_id` management elsewhere except here (by my legacy wip implementation and missed on review?). Maybe we can think of fully implicit or umm _automated_ `device_id `management inherent from the call to  `get_device()` on the current thread - but since we have the field in `DeviceContext` - that makes it redundant and anyway feels like can cause multiple bugs? I guess I had to provide basic description in the doc - but I somehow hoped it will emerge from everyone's edit, since it was so intensively discussed:) have to update doc with the current version of API and diagrams

wrappers/rust/icicle-core/src/msm/tests.rs
```diff
@@ -25,6 +24,7 @@ fn generate_random_affine_points_with_zeroes<C: Curve>(size: usize, num_zeroes:
     points
 }
 
+// TODO: this should be INCORRECT for more than 1 device!! correct it
```
vhnatyk, line 27: ?

DmytroTym: Sorry forgot to remove. It does pass on multi-device machines

vhnatyk (commented):
looks good , just one minor thing similar error strings to const, pretty big one 😄 - maybe worth  one more quick look

wrappers/rust/icicle-cuda-runtime/src/memory.rs
jeremyfelder: I think internal management of `device_id` by thread makes the most sense. `DeviceContext` can also get set this way by using `get_device()`. It might be more bug prone on our end but I think it will be less error prone on the user end which we don't have control over

alxiong:
I like the distinction made for `DeviceVec` and `DeviceSlice`, and the usage of `ManualDrop`. 

Would like to confirm one thing: 

the updated `msm()` API now accepts either `DeviceVec` and `DeviceSlice` now, just to make sure the following flow won't cause double free:
- assign a `DeviceVec` for some bases points (say 2^20) 
- get a `DeviceSlice` for a sub-slice of it (say 2^15 out of it)
- pass on this slice to the `msm()`, for some reason, this cuda operation panic and failed (say due to GPU out of memory)

(currently I'm using `mem::forget()` to forget the subslice without calling destructor to prevent double-freeing. but double-freeing would still be a risk if `msm()` failed in the middle before I got a chance to call `forget()` since currently there's no differences between subslice on device and original owned vec on device, so both have the same `cudaFree` logic in their destructor. 
so just want to double-check that changes introduced here would obviate my concern, since `DeviceSlice`'s destructor won't do anything, correct?)

DmytroTym:
@alxiong yea, the new functions do accept either `DeviceSlice`, `HostSlice` or `DeviceVec`. The latter is accepted by `Deref` to `DeviceSlice` (similar to how other Rust functions that accept slices can be used with vectors). And like normal Rust slices, `DeviceSlice` doesn't own any data so won't call `cudaFree`, only `DeviceVec` frees data :)

jeremyfelder:
Closing in favor of #443 which includes these changes

PR #416, opened 2024-03-04
Update overview.md

Otsar-Raikou:
Changing to "dependent repositories" badge instead of the "Machines using"

### DIFF
--- a/docs/docs/icicle/overview.md
+++ b/docs/docs/icicle/overview.md
@@ -2,7 +2,8 @@
 
 [![GitHub Release](https://img.shields.io/github/v/release/ingonyama-zk/icicle)](https://github.com/ingonyama-zk/icicle/releases)
 
-![Static Badge](https://img.shields.io/badge/Machines%20running%20ICICLE-544-blue)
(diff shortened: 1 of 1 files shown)


LeonHibnik (approved):

jeremyfelder (commented):
Are we sure we want to show this?
Its only capturing golang uses and is likely not accurate

Otsar-Raikou:
@jeremyfelder - i didnt understand this as only using golang uses, is this not icicle in general?

docs/docs/icicle/overview.md
```diff
@@ -2,7 +2,8 @@
 
 [![GitHub Release](https://img.shields.io/github/v/release/ingonyama-zk/icicle)](https://github.com/ingonyama-zk/icicle/releases)
 
-![Static Badge](https://img.shields.io/badge/Machines%20running%20ICICLE-544-blue)
+![Static Badge](https://img.shields.io/badge/Dependent%20Repositories-21-blue)
```
ImmanuelSegol, line 5: if its only for golang then add it to the golang main md file

ImmanuelSegol (changes requested):

Issue #264, opened 2023-11-14
[FEAT]: NTT acceleration


Iridescent-gcrace:
## Description
While reading this paper [GZKP](https://dl.acm.org/doi/10.1145/3575693.3575711), I found that the execution speed of NTT is faster than icicle. Are you guys interested in implementing it into icicle?

Looking forward to hearing your thoughts on this proposal. Thank you for your continuous efforts in improving Icicle.

Iridescent-gcrace:
Thank you for your answer! Your guys' work is excellent! May I ask when the accele

HadarIngonyama:
Hopefully in about 3 months.

cliff0412:
i dont have the account to download the paper. what optimisation do they conduct to improve NTT?

HadarIngonyama:
They make good use of shared memory in all the NTT stages and they have a good memory access pattern to global memory.
Our new implementation isn't based on this article, it's out own design, and it achieves even better results. Our main improvements are:
- Coalesced access to global memory
- Using shared memory for all the stages
- Fully utilizing register memory
- implementing the Winograd-8 algorithm

cliff0412:
> They make good use of shared memory in all the NTT stages and they have a good memory access pattern to global memory. Our new implementation isn't based on this article, it's out own design, and it achieves even better results. Our main improvements are:
> 
> * Coalesced access to global memory
> * Using shared memory for all the stages
> * Fully utilizing register memory
> * implementing the Winograd-8 algorithm

when will the new implementation be out? i think currently the shared memory is only used in certain stages.

HadarIngonyama:
The new implementation was released in release 1.3.0, it's called mixed-radix NTT. check it out - https://github.com/ingonyama-zk/icicle/blob/fc6badcb354dca03cac8e025b785d9d00b0942b8/icicle/appUtils/ntt/kernel_ntt.cu#L641

column-wise arranged inputs - I don't think this is currently supported. could you give an example for a use case? usually the NTT input is just a 1-D array, you want support for a column-wise 2-D array? I guess in that case it would depend on the column size.

cliff0412:
> The new implementation was released in release 1.3.0, it's called mixed-radix NTT. check it out -
> 
> https://github.com/ingonyama-zk/icicle/blob/fc6badcb354dca03cac8e025b785d9d00b0942b8/icicle/appUtils/ntt/kernel_ntt.cu#L641
> 
[...]

PR #424, opened 2024-03-06
NTT columns batch

HadarIngonyama:
This PR adds the columns batch feature - enabling batch NTT computation to be performed directly on the columns of a matrix without having to transpose it beforehand, as requested in issue #264. 

Also some small fixes to the reordering kernels were added and some unnecessary parameters were removes from functions interfaces.


icicle/appUtils/ntt/tests/verification.cu
```diff
@@ -19,29 +19,65 @@ typedef curve_config::scalar_t test_data;
 void random_samples(test_data* res, uint32_t count)
 {
   for (int i = 0; i < count; i++)
+    // res[i] = i==0 ? test_scalar::one() : test_scalar::zero();
```
yshekel, line 22: remove commented code. Also in other places

icicle/appUtils/ntt/tests/verification.cu
```diff
@@ -19,29 +19,65 @@ typedef curve_config::scalar_t test_data;
...
   }
 }
 
+__global__ void copy_batch(test_scalar* in, test_scalar* out, int row_size, int column_size){
+  int tid = blockDim.x * blockIdx.x + threadIdx.x;
+  if (tid >= row_size * column_size) return;
+  out[tid] = in[tid];
+}
+
+__global__ void copy_batch2(test_scalar* in, test_scalar* out, int row_size, int column_size){
```
yshekel, line 40: why need 3 such functions? same for transpose. If you need them, please give clear names

icicle/appUtils/ntt/tests/vri2.cu
```diff
@@ -0,0 +1,228 @@
+
```
yshekel: why is this test needed in addition to the other one?

icicle/appUtils/ntt/thread_ntt.cu
```diff
@@ -211,8 +212,20 @@ public:
     }
   }
 
+  __device__ __forceinline__ void loadGlobalDataBatched(
```
yshekel, line 215: is that only for column batch? if yes, please rename to make to code clearer

wrappers/rust/icicle-core/src/ntt/mod.rs
```diff
@@ -77,6 +77,8 @@ pub struct NTTConfig<'a, S> {
     pub coset_gen: S,
     /// The number of NTTs to compute. Default value: 1.
     pub batch_size: i32,
+    /// True if the NTTs are stored as the columns of a matrix and not as the rows.
```
yshekel, line 80: I would rephrase and say that when true, NTT is computed over columns.

icicle/appUtils/ntt/ntt.cu
yshekel, line 1: you need to make sure that if column batch is true then you choose the mixed-radix alg since radix2 does not support it, right?
Also note that it means that there are a few cases that are no supported (like 2^7) since mixed radix is not supported on those.

yshekel (changes requested):
overall looks good. Wrote some comments, mainly to make the code cleaner.
Note that you must make sure that radix2 is never selected for column batch and it means that some cases are not supported (like 2^7) so I think you should log a clear error.

I think that we must have tests in CI before merging it, including small NTTs and the cases that are not supported to make sure it behaves as expected.

icicle/appUtils/ntt/tests/verification.cu
```diff
@@ -141,22 +198,35 @@ int main(int argc, char** argv)
...
   CHK_IF_RETURN(benchmark(true /*=print*/, count));
 
+  if (COLUMNS_BATCH) {
+    transpose_batch<<<(NTT_SIZE * BATCH_SIZE + 256 - 1)/256,256>>>(GpuOutputNew, GpuScalarsTransposed, BATCH_SIZE, NTT_SIZE);
+    cudaDeviceSynchronize();
+    printf("trans err %d\n", cudaGetLastError());
+    CHK_IF_RETURN(
+      cudaMemcpy(GpuOutputNew, GpuScalarsTransposed, NTT_SIZE * BATCH_SIZE * sizeof(test_data), cudaMemcpyDeviceToDevice));
+    cudaDeviceSynchronize();
+    printf("copy err %d\n", cudaGetLastError());
```
jeremyfelder, line 212: Are these `cudaDeviceSynchronize` needed or for debug?

icicle/appUtils/ntt/tests/verification.cu
```diff
@@ -141,22 +198,35 @@ int main(int argc, char** argv)
...
+  }
+
   // verify
   CHK_IF_RETURN(
     cudaMemcpy(CpuOutputNew.get(), GpuOutputNew, NTT_SIZE * BATCH_SIZE * sizeof(test_data), cudaMemcpyDeviceToHost));
   CHK_IF_RETURN(
     cudaMemcpy(CpuOutputOld.get(), GpuOutputOld, NTT_SIZE * BATCH_SIZE * sizeof(test_data), cudaMemcpyDeviceToHost));
+   cudaDeviceSynchronize();
+    printf("copy err %d\n", cudaGetLastError());
```
jeremyfelder, line 221: Remove

ChickenLover:
@HadarIngonyama please change the branch to `main` as we don't use `main` anymore. Also, I think this will break golang bindings if we don't change the NTTConfig in golang accordingly

icicle/appUtils/ntt/tests/verification.cu
HadarIngonyama: done

icicle/appUtils/ntt/tests/verification.cu
HadarIngonyama: removed

icicle/appUtils/ntt/tests/vri2.cu
HadarIngonyama: removed

icicle/appUtils/ntt/thread_ntt.cu
HadarIngonyama: done

wrappers/rust/icicle-core/src/ntt/mod.rs
HadarIngonyama: done

icicle/appUtils/ntt/ntt.cu
HadarIngonyama: done

icicle/appUtils/ntt/tests/verification.cu
HadarIngonyama: removed

icicle/appUtils/ntt/tests/verification.cu
HadarIngonyama: removed

HadarIngonyama:
I've fixed all the above comments, please review

DmytroTym (approved):
Overall looks good. I can add columns batch mode into one of Rust tests as part of this PR (or else will do it later on). I also have two performance related questions:

1. When running `verification.cu` with different parameters, I see columns batch mode consistently outperforming normal batch mode. One example: NTT size is 2^20, batch size is 128, out of place, NN ordering. Columns batch takes ~70.2 ms. and normal batch ~94.7. The kernels are the same, most of the difference is in relative speed of the first two kernels: `batch_elementwise_mul_with_reorder_kernel` - 9.5 ms. vs. 14 ms. and especially `reorder_digits_inplace_and_normalize_kernel` - 10.4 ms. vs. 25.8 ms. Do you understand why this happens?
2. For smaller NTTs, radix2 is faster and while this might be offset by a larger batch size, when NTT size times batch size is ~2^16 or smaller, radix2 seems to be better. If such small sizes are relevant, I can add support for column batching for radix2. Though both @cliff0412 's use-case and danksharding are significantly larger.

HadarIngonyama:
> Overall looks good. I can add columns batch mode into one of Rust tests as part of this PR (or else will do it later on). I also have two performance related questions:
> 
> 1. When running `verification.cu` with different parameters, I see columns batch mode consistently outperforming normal batch mode. One example: NTT size is 2^20, batch size is 128, out of place, NN ordering. Columns batch takes ~70.2 ms. and normal batch ~94.7. The kernels are the same, most of the difference is in relative speed of the first two kernels: `batch_elementwise_mul_with_reorder_kernel` - 9.5 ms. vs. 14 ms. and especially `reorder_digits_inplace_and_normalize_kernel` - 10.4 ms. vs. 25.8 ms. Do you understand why this happens?
> 2. For smaller NTTs, radix2 is faster and while this might be offset by a larger batch size, when NTT size times batch size is ~2^16 or smaller, radix2 seems to be better. If such small sizes are relevant, I can add support for column batching for radix2. Though both @cliff0412 's use-case and danksharding are significantly larger.

1. Not sure how exactly it improves the reordering, I think in general there is room for optimizing all the reordering kernels.
2. Yes, it wood be great if you could add support for radix2 NTT also.
I think @LeonHibnik  is working on adding a rust test, please coordinate with him.

DmytroTym:
I wanted to add a new test initially as all our tests are really cluttered with logic. But `test_ntt_batched` is a really convenient place for this check. Also didn't have the time do the radix2 version so that's future work I guess.

LeonHibnik (approved):

Issue #147, opened 2023-08-16
[FEAT]: MSM - add precompute support


LeonHibnik:
## Description

Provide support for utilizing msm pre-computation 

## Motivation

Accelerate msm performance

PR #427, opened 2024-03-07
MSM pre-computation

DmytroTym:
## Brief description

This PR adds pre-computation to the MSM, for some theory see [this](https://youtu.be/KAWlySN7Hm8?si=XeR-htjbnK_ySbUo&t=1734) timecode of Niall Emmart's talk.
In terms of public APIs, one method is added. It does the pre-computation on-device leaving resulting data on-device as well. No extra structures are added, only `precompute_factor` from `MSMConfig` is now activated.

## Performance

While performance gains are for now often limited by our inflexibility in choice of `c` (for example, very large MSMs get basically no speedup from pre-compute because currently `c` cannot be larger than 16), there's still a number of MSM sizes which get noticeable improvement:

| Pre-computation factor | bn254 size `2^20` MSM, ms.  | bn254 size `2^12` MSM, size `2^10` batch, ms. | bls12-381 size `2^20` MSM, ms. | bls12-381 size `2^12` MSM, size `2^10` batch, ms. |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| 1  | 14.1  | 82.8  | 25.5  | 136.7  |
| 2  | 11.8  | 76.6  | 20.3  | 123.8  |
| 4  | 10.9  | 73.8  | 18.1  | 117.8  |
| 8  | 10.6  | 73.7  | 17.2  | 116.0  |

Here for example pre-computation factor = 4 means that alongside each original base point, we pre-compute and pass into the MSM 3 of its "shifted" versions. Pre-computation factor = 1 means no pre-computation. GPU used for benchmarks is a 3090Ti.

## TODOs and open questions

- Golang APIs are missing;
- I mentioned that to utilise pre-compute to its full potential we need arbitrary choice of `c`. One issue with this is that pre-compute will become dependent on `c`. For now this is not the case as `c` can only be a power of 2 and powers of 2 can always share the same pre-computation. So apparently we need to make `c` a parameter of the precompute function to future-proof it from a breaking change. This is pretty unnatural and counterintuitive as `c` is typically chosen in runtime after pre-compute is done but I don't really see another way, pls let me know if you do. UPD: `c` is added into pre-compute function, for now it's unused and it's documented how it will change in the future.

Resolves https://github.com/ingonyama-zk/icicle/issues/147
Co-authored with @ChickenLover


DmytroTym:
This is ready for review although I plan to push another commit with more documentation of internals and public interfaces.

ChickenLover (approved):

ChickenLover (approved):

LeonHibnik (approved):

PR #428, opened 2024-03-08
update go install source in Dockerfile

cyl19970726:
## Describe the changes

The go package resources have been updated in https://go.dev/dl/ and it leads to failure when executing `docker build -t icicle-demo .`


ImmanuelSegol (approved):
@jeremyfelder have a look to please

jeremyfelder (approved):
I haven't been able to reproduce a failure/error as the old URL redirects correctly but its a good idea to update anyway. Thanks for the contribution 🔥 🚀

PR #429, opened 2024-03-09
Update examles/rust/msm dockerfile to  fix bug 

cyl19970726:
## Describe the changes

There is a conflict between https://github.com/ingonyama-zk/icicle/blob/main/examples/rust/msm/Cargo.toml#L7 and  https://github.com/ingonyama-zk/icicle/blob/main/examples/rust/msm/.devcontainer/Dockerfile#L24, it leads to an error happen when executing `cargo run --release` at docker

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ ENV PATH="/root/.cargo/bin:${PATH}"
 
 # Install Golang
 ENV GOLANG_VERSION 1.21.1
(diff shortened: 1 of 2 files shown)


jeremyfelder:
@cyl19970726 Thanks for your contribution! 🚀 

What docker command are you running with this?

I think this will still produce an error if the context of the `docker build` is the `examples/rust/msm` directory since the COPY command won't be able to access outside the context

PR #430, opened 2024-03-11
Cleanup golang curves

jeremyfelder:
## Describe the changes

This PR moves non-curve specific functionality to the core package for MSM, NTT, and VecOps.

### DIFF
--- a/wrappers/golang/core/msm.go
+++ b/wrappers/golang/core/msm.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"fmt"
+	"unsafe"
 
 	cr "github.com/ingonyama-zk/icicle/wrappers/golang/cuda_runtime"
 )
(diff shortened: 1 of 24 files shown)


ImmanuelSegol (approved):
The design looks ok, no comments there.

I like the use of generics, my only concern is that golang generics come at performance costs in some cases. I have never benchmarked this and im not sure if this change will effect performance.


Also regarding `scalarsPointer = unsafe.Pointer(&scalars.(HostSlice[S])[0])` is this casting ? Im just wondering how this will effect Error messages and debugging incase of an issue.

jeremyfelder:
>I like the use of generics, my only concern is that golang generics come at performance costs in some cases. I have never benchmarked this and im not sure if this change will effect performance.

This is a good point, not sure what the performance cost is either. I'll measure it and see if it makes sense. I have other ideas on how to reduce the amount of templating that is needed but also not sure how those ideas affect perfomance.

>Also regarding scalarsPointer = unsafe.Pointer(&scalars.(HostSlice[S])[0]) is this casting ?

Technically its type assertion on an interface type and not casting but essentially its the same

jeremyfelder:
Superseded by #473 and #457

PR #435, opened 2024-03-14
Add Keccak hash function

ChickenLover:
This PR adds support for Keccak-256 and Keccak-512. It only adds them in c++. There is no way of adding rust or golang wrappers  rn as it requires having an `icicle-common` create / mod

### DIFF
--- a/icicle/appUtils/keccak/keccak.cu
+++ b/icicle/appUtils/keccak/keccak.cu
@@ -247,7 +247,7 @@ namespace keccak {
 
     int number_of_threads = 1024;
     int number_of_gpu_blocks = (number_of_blocks - 1) / number_of_threads + 1;
-    keccak_hash_blocks<C, D><<<number_of_threads, number_of_gpu_blocks, 0, stream>>>(
+    keccak_hash_blocks<C, D><<<number_of_gpu_blocks, number_of_threads, 0, stream>>>(
       input_device, input_block_size, number_of_blocks, output_device);
 
     if (!config.are_inputs_on_device) CHK_IF_RETURN(cudaFreeAsync(input_device, stream));
--- a/icicle/appUtils/keccak/keccak.cu
+++ b/icicle/appUtils/keccak/keccak.cu
@@ -1,0 +1,275 @@
+#include "keccak.cuh"
+
+namespace keccak {
+#define ROTL64(x, y) (((x) << (y)) | ((x) >> (64 - (y))))
+
+#define TH_ELT(t, c0, c1, c2, c3, c4, d0, d1, d2, d3, d4)                                                              \
+  {                                                                                                                    \
+    t = ROTL64((d0 ^ d1 ^ d2 ^ d3 ^ d4), 1) ^ (c0 ^ c1 ^ c2 ^ c3 ^ c4);                                                \
+  }
+
+#define THETA(                                                                                                         \
+  s00, s01, s02, s03, s04, s10, s11, s12, s13, s14, s20, s21, s22, s23, s24, s30, s31, s32, s33, s34, s40, s41, s42,   \
+  s43, s44)                                                                                                            \
+  {                                                                                                                    \
+    TH_ELT(t0, s40, s41, s42, s43, s44, s10, s11, s12, s13, s14);                                                      \
+    TH_ELT(t1, s00, s01, s02, s03, s04, s20, s21, s22, s23, s24);                                                      \
+    TH_ELT(t2, s10, s11, s12, s13, s14, s30, s31, s32, s33, s34);                                                      \
+    TH_ELT(t3, s20, s21, s22, s23, s24, s40, s41, s42, s43, s44);                                                      \
+    TH_ELT(t4, s30, s31, s32, s33, s34, s00, s01, s02, s03, s04);                                                      \
+    s00 ^= t0;                                                                                                         \
+    s01 ^= t0;                                                                                                         \
+    s02 ^= t0;                                                                                                         \
+    s03 ^= t0;                                                                                                         \
+    s04 ^= t0;                                                                                                         \
+                                                                                                                       \
+    s10 ^= t1;                                                                                                         \
+    s11 ^= t1;                                                                                                         \
+    s12 ^= t1;                                                                                                         \
+    s13 ^= t1;                                                                                                         \
+    s14 ^= t1;                                                                                                         \
+                                                                                                                       \
+    s20 ^= t2;                                                                                                         \
+    s21 ^= t2;                                                                                                         \
+    s22 ^= t2;                                                                                                         \
+    s23 ^= t2;                                                                                                         \
+    s24 ^= t2;                                                                                                         \
+                                                                                                                       \
+    s30 ^= t3;                                                                                                         \
+    s31 ^= t3;                                                                                                         \
+    s32 ^= t3;                                                                                                         \
+    s33 ^= t3;                                                                                                         \
+    s34 ^= t3;                                                                                                         \
+                                                                                                                       \
+    s40 ^= t4;                                                                                                         \
+    s41 ^= t4;                                                                                                         \
+    s42 ^= t4;                                                                                                         \
+    s43 ^= t4;                                                                                                         \
+    s44 ^= t4;                                                                                                         \
+  }
+
+#define RHOPI(                                                                                                         \
+  s00, s01, s02, s03, s04, s10, s11, s12, s13, s14, s20, s21, s22, s23, s24, s30, s31, s32, s33, s34, s40, s41, s42,   \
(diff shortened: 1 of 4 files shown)


icicle/appUtils/keccak/test.cu
HadarIngonyama, line 1: Is this the verification test? I can't see a comparison of the results to a reference so how do you know that the results is correct? also, what about checking non-default configurarions?

icicle/appUtils/keccak/keccak.cu
```diff
@@ -0,0 +1,275 @@
+#include "keccak.cuh"
+
+namespace keccak {
+#define ROTL64(x, y) (((x) << (y)) | ((x) >> (64 - (y))))
```
HadarIngonyama, line 4: what is the benefit of using #define for these functions and not just making them device functions?

icicle/appUtils/keccak/keccak.cu
```diff
@@ -0,0 +1,275 @@
...
+        s[0], s[5], s[10], s[15], s[20], s[1], s[6], s[11], s[16], s[21], s[2], s[7], s[12], s[17], s[22], s[3], s[8],
+        s[13], s[18], s[23], s[4], s[9], s[14], s[19], s[24]);
+      IOTA(s[0], RC[i]);
+    }
+  }
+
+  template <int C, int D>
+  __global__ void keccak_hash_blocks(uint8_t* input, int input_block_size, int number_of_blocks, uint8_t* output)
+  {
+    int bid = (blockIdx.x * blockDim.x) + threadIdx.x;
```
HadarIngonyama, line 175: bid? block id? this is the thread id

icicle/appUtils/keccak/keccak.cu
```diff
@@ -0,0 +1,275 @@
...
+    uint8_t* output_device;
+    if (config.are_outputs_on_device) {
+      output_device = output;
+    } else {
+      CHK_IF_RETURN(cudaMallocAsync(&output_device, number_of_blocks * (D / 8), stream));
+    }
+
+    int number_of_threads = 1024;
+    int number_of_gpu_blocks = (number_of_blocks - 1) / number_of_threads + 1;
+    keccak_hash_blocks<C, D><<<number_of_threads, number_of_gpu_blocks, 0, stream>>>(
```
HadarIngonyama, line 250: afaik the first argument is number of blocks and the second is number of threads... not sure how it's working for you

icicle/appUtils/keccak/keccak.cu
```diff
@@ -0,0 +1,275 @@
...
+
+    uint8_t* b_input = input + bid * input_block_size;
+    uint8_t* b_output = output + bid * d_bytes;
+    uint64_t state[25] = {}; // Initialize with zeroes
+
+    int input_len = input_block_size;
+
+    // absorb
+    while (input_len >= r_bytes) {
+      // #pragma unroll
```
HadarIngonyama, line 190: I guess when you unroll you run into program memory problems?

icicle/appUtils/keccak/keccak.cu
```diff
@@ -0,0 +1,275 @@
...
+    int bid = (blockIdx.x * blockDim.x) + threadIdx.x;
+    if (bid >= number_of_blocks) { return; }
+
+    const int r_bits = 1600 - C;
+    const int r_bytes = r_bits / 8;
+    const int d_bytes = D / 8;
+
+    uint8_t* b_input = input + bid * input_block_size;
+    uint8_t* b_output = output + bid * d_bytes;
+    uint64_t state[25] = {}; // Initialize with zeroes
```
HadarIngonyama, line 184: does this fit in register space?

icicle/appUtils/keccak/keccak.cu
```diff
@@ -0,0 +1,275 @@
...
+        s[13], s[18], s[23], s[4], s[9], s[14], s[19], s[24]);
+      KHI(
+        s[0], s[5], s[10], s[15], s[20], s[1], s[6], s[11], s[16], s[21], s[2], s[7], s[12], s[17], s[22], s[3], s[8],
+        s[13], s[18], s[23], s[4], s[9], s[14], s[19], s[24]);
+      IOTA(s[0], RC[i]);
+    }
+  }
+
+  template <int C, int D>
+  __global__ void keccak_hash_blocks(uint8_t* input, int input_block_size, int number_of_blocks, uint8_t* output)
```
HadarIngonyama, line 173: this function includes many loops, is there no way to parallelize it?

HadarIngonyama (commented):
I've added some comments and questions to the code. 2 important questions are:
1. How did you verify the correctness?
2. How does the performance compare to other implementations? and is the kernel fully utilizing memory/compute throughput?

icicle/appUtils/keccak/keccak.cu
```diff
@@ -0,0 +1,275 @@
...
+    s34 ^= t3;                                                                                                         \
+                                                                                                                       \
+    s40 ^= t4;                                                                                                         \
+    s41 ^= t4;                                                                                                         \
+    s42 ^= t4;                                                                                                         \
+    s43 ^= t4;                                                                                                         \
+    s44 ^= t4;                                                                                                         \
+  }
+
+#define RHOPI(                                                                                                         \
```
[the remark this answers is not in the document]
ChickenLover: Yes. https://en.wikipedia.org/wiki/SHA-3#The_block_permutation

icicle/appUtils/keccak/test.cu
ChickenLover: I actually checked them by comparing the results locally against other implementations (one of them was python sha3). I was going to have better tests in Rust, but then I found that it is impossible to include in Rust atm

icicle/appUtils/keccak/keccak.cu
ChickenLover: this is just for inlining

icicle/appUtils/keccak/keccak.cu
ChickenLover: by block id I mean the block of original input

icicle/appUtils/keccak/keccak.cu
ChickenLover: good catch. I was only testing on small inputs, so it didn't trigger any errors

icicle/appUtils/keccak/keccak.cu
ChickenLover: actually, I just need to test what gives better performance

icicle/appUtils/keccak/keccak.cu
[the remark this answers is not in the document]
HadarIngonyama: hmm ok but what is th_elt? and maybe rhopi should be rho_pi

icicle/appUtils/keccak/keccak.cu
HadarIngonyama: well it's not clear from the code. there should be a clear separation between input blocks notation and cuda blocks and threads notation

krakhit:
For correctness can you add a test that generate test vectors from https://keccak.team/2013/yes_this_is_keccak.html and check that it produces the same result as the standard implementation

icicle/appUtils/keccak/keccak.cu
ChickenLover: good question. Never actually calculated it, maybe we need to reduce the number of threads per block. One state is 200 bytes, and a common register space size is 256kB afaik. So it should just barely fit 1024 states

HadarIngonyama: the limit is 255 register per thread.
you can check the usage in nsight compute.

icicle/appUtils/keccak/keccak.cu
ChickenLover: There is no way and no need to parallelize them.
The absorb loop cannot be parallelized by design, as it is sequential, but if you have each input of size lesser than R then this loop is only 1 iteration
The last block is at most R/64 iterations
The padding loop is at most R/64 iterations, and it's just assigning zeroes
Last loop can be parallelized but I don't think it will give any performance boost tbh

icicle/appUtils/keccak/keccak.cu
ChickenLover: yeah, I agree. My vocabulary is just too poor to come up with better names:) And also, we don't have any cuda blocks and threads notation standard atm. Maybe it's a good idea to add some conventional naming in v2

LeonHibnik (approved):

DmytroTym (approved):

krakhit (approved):
For correctness please add a test withtest vectors from https://keccak.team/2013/yes_this_is_keccak.html and check that it produces the same result as the standard implementation

PR #443, opened 2024-03-21
v2 new design

ChickenLover:
### Curves supported:

- [x] bls12-377
- [x] bls12-381
- [x] bn254
- [x] grumpkin
- [x] bw6

### Primitives supported:

- [x] NTT (supported, but not working correctly)
- [x] ECNTT
- [x] MSM
- [x] Poseidon
- [x] Tests

### Hash supported:

- [x] Keccak

### To-Do before merge:

- [x] Change to use a common `curve_config.cuh` / `field_config.cuh`
- [x] Merge with current main
- [x] Update CI configs
- [x] Test if ECNTT works
- [x] Test if G2 MSM works
- [x] Merge device memory redesign
- [x] Add babybear field

### To-Do after merge:
- [ ] Update golang API
- [x] Add Rust crate for icicle_hash
- [ ] Update all the examples to use new API
- [ ] Add polynomial API
- [ ] Update Rust curve template
- [ ] Update docs
- [ ] Expose c_api.h as proposed by @yshekel


icicle/CMakeLists.txt
```diff
@@ -96,26 +34,8 @@ if (NOT BUILD_TESTS)
...
-    icicle
-    utils/vec_ops.cu
-    utils/mont.cu
-    primitives/field.cu
-    primitives/projective.cu
-    appUtils/msm/msm.cu
-    ${ICICLE_SOURCES}
-  )
-  set_target_properties(icicle PROPERTIES OUTPUT_NAME "ingo_${CURVE}")
+  add_library(icicle curves/${CURVE}/extern.cu)
```
yshekel, line 37: so your intention is that this extern would include everything and we compile it as a single file? note that this typically slows down the compilation

icicle/fields/bn254/bn254_params.cuh
```diff
@@ -173,27 +173,6 @@ namespace bn254 {
     // true if i^2 is negative
     static constexpr bool i_squared_is_negative = true;
   };
-
```
yshekel, line 176: where did you move those to? or removed? why?

icicle/fields/bn254/extern.cu
```diff
@@ -0,0 +1,8 @@
+#define FIELD bn254
```
yshekel, line 1: ideally we should compile those three files in parallel and not include source files. Only headers (cuh) should be included.
I know it's tricky for template libraries but we can explicitly instantiate where need to.
The way you chose is slower and I think should be avoided if possible.

icicle/fields/bn254/bn254_params.cuh
yshekel: oh I get it it's part of the curve, not the field. Please ignore.

icicle/curves/bn254/bn254.cuh
```diff
@@ -0,0 +1,45 @@
+#pragma once
```
yshekel, line 1: I find it confusing that both curves and fields have the same file names.

icicle/fields/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+  set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -DG2_DEFINED=ON")
+endif ()
+
+option(BUILD_TESTS "Build tests" OFF)
+
+if (NOT BUILD_TESTS)
+
+  message(STATUS "Building without tests.")
+
+  add_library(icicle ${FIELD}/extern.cu)
```
yshekel, line 37: something not clear to me: here you define icicle as one library but in another library you define it as well as another one.
I think you need to name the libraries better  (maybe here icicle_fields or fields or so) and have icicle link to all of them

yshekel: also note that if multiple libs would depend on fields (probably will happen) then we will end up with multiple symbols so the library would end up larger with duplicate code. I think this should not cause link issues but not ideal. 
For example both curves and polynomials would depend on fields so both would link to fields and we end up with fields twice in the final icicle library. Unless I misunderstood your intention.

One option is to use shared libs. Or think about the structure again.

yshekel (commented):
Wrote some comments.

About the description, what do you mean by "Freedom of implementation - everything can be replaced for a specific curve / field". Can you give an example that cannot be supported right now but will be with this redesign?

icicle/fields/bn254/extern.cu
ChickenLover: you are right. I'll change it

ChickenLover:
@yshekel RE: "freedom of implementation" example. We currently have a single NTT implementation that is being called for all the fields. If we want to have a different NTT algo for some specific field/fields (e.g. small fields twiddles generation), we will need to do smth like `#if CURVE == smth`

icicle/CMakeLists.txt
```diff
@@ -1,152 +1,29 @@
...
-    set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -DCURVE_ID=${I}")
-    set(IS_CURVE_SUPPORTED TRUE)
-  endif ()
-endforeach()
+option(BUILD_TESTS "Build tests" OFF)
 
-if (NOT IS_CURVE_SUPPORTED)
-  message( FATAL_ERROR "The value of CURVE variable: ${CURVE} is not one of the supported curves: ${SUPPORTED_CURVES}" )
+if((DEFINED CURVE) AND (DEFINED FIELD))
+  message( FATAL_ERROR "CURVE and FIELD cannot be defined at the same time" )
```
yshekel, line 13: why is that? Can't we now choose a subset of fields and curves and build them? 
BTW we are not building each curve to a lib anymore right? I mean we build all to one lib, unless I misunderstood

icicle/CMakeLists.txt
```diff
@@ -1,152 +1,29 @@
...
-    set(CMAKE_CUDA_ARCHITECTURES native) # do native
-  else()
-    # no GPUs found, like on Github CI runners
-    set(CMAKE_CUDA_ARCHITECTURES 50) # some safe value
-  endif()
-endif()
+set_env()
+set_gpu_env()
 
 project(icicle LANGUAGES CUDA CXX)
```
yshekel, line 8: I think you should define all options on top so it's clear which options are there and the default values, rather than looking for them in the code

icicle/CMakeLists.txt
```diff
@@ -1,152 +1,29 @@
...
-set(IS_CURVE_SUPPORTED FALSE)
-set(I 0)
-foreach (SUPPORTED_CURVE ${SUPPORTED_CURVES})
-  math(EXPR I "${I} + 1")
-  if (CURVE STREQUAL SUPPORTED_CURVE)
-    set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -DCURVE_ID=${I}")
-    set(IS_CURVE_SUPPORTED TRUE)
-  endif ()
-endforeach()
+option(BUILD_TESTS "Build tests" OFF)
```
yshekel, line 10: I think you should define all options on top so it's clear which options are there and the default values, rather than looking for them in the code

icicle/CMakeLists.txt
ChickenLover: I thought about it, but then we need to rethink our wrappers design. As for now - rust builds Icicle in each curve

icicle/Common.cmake
```diff
@@ -0,0 +1,65 @@
+function(set_env)
```
yshekel, line 1: consider putting this under a 'cmake' directory. Later we may add some other cmake scripts that do specific things and it will be easy to navigate

icicle/CMakeLists.txt
yshekel: Maybe also options like G2 and ECNTT can be defined here

icicle/include/fields/bn254/bn254.cuh
```diff
@@ -0,0 +1,21 @@
+#pragma once
+#ifndef BN254_FIELDS_H
```
yshekel, line 2: maybe worth renaming fields and/or curves fields.

icicle/include/fields/field_config.cuh
```diff
@@ -0,0 +1,58 @@
...
+ */
+namespace field_config {
+  /**
+   * Scalar field. Is always a prime field.
+   */
+  typedef Field<fp_config> scalar_t;
+
+#if defined(BASE_FIELD)
+  /**
+   * Base field of G1 curve. Is always a prime field.
```
yshekel, line 43: I find it weird that the field knows about the curve. I mean fields and curves are still somewhat coupled.
On the other hand we kind of discussed it and we go back and forth. I don't know  best option would be.

BTW is there one interesting extension field per field? I know that for some small fields the extension has 4 elements and wonder how it will fit here, and whether we will need more than one extension for a given field.

icicle/src/fields/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+list(APPEND FIELD_SOURCE ${SRC}/poseidon/poseidon.cu)
+list(APPEND FIELD_SOURCE ${SRC}/poseidon/tree/merkle.cu)
+
+if (NOT FIELD IN_LIST SUPPORTED_FIELDS_WITHOUT_NTT)
+    list(APPEND FIELD_SOURCE ${SRC}/ntt/extern.cu)
+    list(APPEND FIELD_SOURCE ${SRC}/ntt/kernel_ntt.cu)
+endif()
+
+add_library(${TARGET} STATIC ${FIELD_SOURCE})
+target_include_directories(${TARGET} PUBLIC ${CMAKE_SOURCE_DIR}/include/ ${CMAKE_SOURCE_DIR}/src/)
```
yshekel, line 42: if possible try to avoid the src in incdir

icicle/src/fields/mont.cu
```diff
@@ -0,0 +1,19 @@
+#include "fields/field_config.cuh"
+
+using namespace field_config;
+
+#include "utils/mont.cuh"
+#include "gpu-utils/device_context.cuh"
+#include "utils/utils.h"
+
+namespace mont {
+  extern "C" cudaError_t CONCAT_EXPAND(FIELD, ScalarConvertMontgomery)(
```
yshekel, line 10: I wonder if we should expose all the C-API declarations in header files. It's of course relevant to all exposed APIs. I realize that ffi is declaring it too but to make the exposed APIs clear maybe it's worth doing so.
What do you think?

icicle/src/msm/msm.cuh
```diff
@@ -4,12 +4,11 @@
 
 #include <cuda_runtime.h>
 
-#include "curves/curve_config.cuh"
-#include "primitives/affine.cuh"
-#include "primitives/field.cuh"
-#include "primitives/projective.cuh"
-#include "utils/device_context.cuh"
-#include "utils/error_handler.cuh"
+#include "curves/affine.cuh"
```
yshekel, line 7: why isn't this file it in include dir?

yshekel (commented):
Overall looks pretty good!
I think that fields and curves are not decoupled which is not ideal but again, we discussed it and I don't know if it's that bad.  I wrote a comment about that and some more in the fields section.
If we can avoid exposing the curve's fields separately from the curves, maybe it would help.

about hashes, are they exposed in all the libs currently? Is this the intention?

About polynomials, they should probably link to a field and/or a curve in case coeffs/image are EC points, plus NTT. How would it fit? Is it a library per curve/field? or part of each curve/field library?

icicle/include/fields/field_config.cuh
ChickenLover: I also find this curve/field binding very awkward.

Regarding the extension field, yes, I also thought that it should be more flexible. But I think it can be added afterwards, without breaking the structure

The curves' fields are not used on their own, that is true. Actually, that was my very first proposal - to not move their compilation into fields module, but rather compile them inside the according curve. But then, we started going back and forth with separating them into their own lib to utilize cmake caching (for grumpkin and bw).

Actually, I think I will once again try to move them into curves and only leave goldilocks/babybear/etc in `fields/CMakeLists.txt`

icicle/src/fields/mont.cu
ChickenLover: Yes, I think we can try it

icicle/include/fields/field_config.cuh
yshekel: I also think it's probably better but maybe let's discuss it with the team so that you don't do this change back and forth

icicle/src/msm/msm.cuh
ChickenLover: good catch, fixed

icicle/src/curves/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+
+if (G2)
+  set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -DG2 -DEXT_FIELD")
+endif ()
+
+if (BUILD_BASE_FIELD)
+  set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -DBASE_FIELD")
+endif ()
+
+set(TARGET icicle_curve_${CURVE})
```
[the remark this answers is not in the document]
ChickenLover: yeah, you are right. It's just the legacy of previous designs. Will remove

ChickenLover:
@yshekel thank you for your review

> about hashes, are they exposed in all the libs currently? Is this the intention?

Not sure what you meant. All the non-field hashes are compiled in `ingo_hash` lib. There are no Rust nor Golang wrappers atm. I am planning to add them later

> About polynomials, they should probably link to a field and/or a curve in case coeffs/image are EC points, plus NTT. How would it fit? Is it a library per curve/field? or part of each curve/field library?

I think the most simple solution would be to make Polynomials part of each curve/field library, just like NTT is instantiated as NTT in field libs and as NTT + optional ECNTT in curve libs

icicle/include/fields/field_config.cuh
ChickenLover: ok, I'll fix the other stuff then, while we wait for tomorrow's meeting

icicle/src/fields/mont.cu
yshekel: Maybe we could add a include/c_api dir and place all those headers there. It would be quite clear what is exposed without looking for it in src dirs. I think better have different files, for NTT, MSM etc. rather than one for all

yshekel:
> @yshekel thank you for your review
> 
> > about hashes, are they exposed in all the libs currently? Is this the intention?
> 
> Not sure what you meant. All the non-field hashes are compiled in `ingo_hash` lib. There are no Rust nor Golang wrappers atm. I am planning to add them later
> 
> > About polynomials, they should probably link to a field and/or a curve in case coeffs/image are EC points, plus NTT. How would it fit? Is it a library per curve/field? or part of each curve/field library?
> 
> I think the most simple solution would be to make Polynomials part of each curve/field library, just like NTT is instantiated as NTT in field libs and as NTT + optional ECNTT in curve libs

Makes sense yes. I probably missed the hash lib so yeah that's ok then.

icicle/src/ntt/tests/verification.cu
```diff
@@ -1,20 +1,19 @@
 
-#define CURVE_ID BLS12_381
+#define CURVE_ID 1
 
-#include "primitives/field.cuh"
-#include "primitives/projective.cuh"
+#include "curves/bn254/bn254.cuh"
```
yshekel, line 4: why directly include and not via curve_config? and BTW I think it's broken since there is no such file

icicle/src/ntt/tests/verification.cu
```diff
@@ -1,20 +1,19 @@
...
-#include "curves/curve_config.cuh"
 #include "ntt/ntt.cu"
-#include "ntt/ntt_impl.cuh"
+#include "kernel_ntt.cu"
 #include <memory>
 
-typedef curve_config::scalar_t test_scalar;
-typedef curve_config::scalar_t test_data;
-#include "kernel_ntt.cu"
+typedef bn254::scalar_t test_scalar;
```
yshekel, line 15: why not via field_config?

ChickenLover:
Ready to be merged. With the lack of examples and golang bindings, but everything else seems to work

icicle/tests/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
+cmake_minimum_required(VERSION 3.18)
+
+include(../cmake/Common.cmake)
+include(../cmake/CurvesCommon.cmake)
+include(../cmake/FieldsCommon.cmake)
+
+set_env()
+set_gpu_env()
+
+project(icicle_tests LANGUAGES CUDA CXX)
```
yshekel, line 10: I have a concern here: polynomial test in my branch links to icicle but I don't see how I will be able to do it this way. Same for benchmarks.
I think this is solved by not making tests a standalone project but call add_subdirectory() from the top level cmake so that the test target can link to icicle_curve or icicle_field.

icicle/src/fields/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+set(TARGET icicle_field)
+
+set(SRC ${CMAKE_SOURCE_DIR}/src)
+
+set(FIELD_SOURCE ${SRC}/fields/field.cu)
+list(APPEND FIELD_SOURCE ${SRC}/vec_ops/extern.cu)
+list(APPEND FIELD_SOURCE ${SRC}/fields/mont.cu)
+
+# TODO: impl poseidon for small fields. note that it needs to be defined for the extension field!
+if (DEFINED CURVE)
```
yshekel, line 16: would icicle_field not compiled without those?

icicle/src/fields/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+  list(APPEND FIELD_SOURCE ${SRC}/poseidon/poseidon.cu)
+  list(APPEND FIELD_SOURCE ${SRC}/poseidon/tree/merkle.cu)
+endif()
+
+if (NOT FIELD IN_LIST SUPPORTED_FIELDS_WITHOUT_NTT)
+  list(APPEND FIELD_SOURCE ${SRC}/ntt/extern.cu)
+  list(APPEND FIELD_SOURCE ${SRC}/ntt/kernel_ntt.cu)
+endif()
```
yshekel, line 24: The intention is to do the same for polynomials right?

yshekel (commented):
Very nice!

I wrote a few comments, mainly one concern about the tests and minor comments.

icicle/src/fields/CMakeLists.txt
ChickenLover: yes

icicle/tests/CMakeLists.txt
ChickenLover: sure, we can do that

icicle/src/fields/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+if (EXT_FIELD)
+  set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -DEXT_FIELD")
+endif ()
+
+SET(SUPPORTED_FIELDS_WITHOUT_NTT grumpkin)
+
+set(TARGET icicle_field)
+
+set(SRC ${CMAKE_SOURCE_DIR}/src)
```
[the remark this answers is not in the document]
ChickenLover: What you talking about is CMAKE_CURRENT_SOURCE_DIR. ([link](https://stackoverflow.com/questions/53195186/difference-between-cmake-current-source-dir-cmake-source-dir-variables#:~:text=CMAKE_SOURCE_DIR%20is%20where%20cmake%20was,CMAKE_SOURCE_DIR%20will%20remain%20the%20same.))
CMAKE_SOURCE_DIR is pointing to a top level cmake, which will be the same as PROJECT_SOURCE_DIR

icicle/src/fields/CMakeLists.txt
ChickenLover: poseidon rn is not implemented for babybear, so yeah, if we try building it, it will not compile

PR #444, opened 2024-03-24
fix: NTT input is const

yshekel:
Make NTT input const pointer so that polynomial API implementation can call it with internal memory pointer.


HadarIngonyama (approved):
seems fine, this doesn't affect the ability to perform inplace NTT right?

yshekel:
> seems fine, this doesn't affect the ability to perform inplace NTT right?

Right. In that case the output pointer is the same but passed mutable as well.

ChickenLover (approved):

PR #451, opened 2024-04-03
Groth16 test with verification

DmytroTym:
This PR adds a "dummy verifier" to Groth16 and a test which swaps G2 for the scalar field. This allows checking correctness without pairings. One change to the prover that was needed for everything to work is zero padding to the power of two without which verifier fails.
One small fix also added which corrects `Projective::from_affine` function for zero input.

### DIFF
--- a/icicle/tests/polynomial_test.cu
+++ b/icicle/tests/polynomial_test.cu
@@ -20,6 +20,27 @@ using curve_config::scalar_t;
 #include "appUtils/msm/msm.cuh"
 #include "utils/device_context.cuh"
 
+class dummy_g2_t: public scalar_t {
+public:
+  static constexpr __host__ __device__ dummy_g2_t to_affine(const dummy_g2_t& point) { return point; }
+
+  static constexpr __host__ __device__ dummy_g2_t from_affine(const dummy_g2_t& point) { return point; }
+
+  static constexpr __host__ __device__ dummy_g2_t generator() { return dummy_g2_t { scalar_t::one() }; }
+
+  static __host__ __device__ dummy_g2_t zero() { return dummy_g2_t { scalar_t::zero() }; }
+
+  friend __host__ __device__ dummy_g2_t operator*(const scalar_t& xs, const dummy_g2_t& ys) {
+    return dummy_g2_t { scalar_t::reduce(scalar_t::mul_wide(xs, ys)) };
+  }
+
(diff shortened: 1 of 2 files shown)


icicle/tests/polynomial_test.cu
```diff
@@ -38,6 +59,16 @@ namespace msm {
     return CONCAT_EXPAND(CURVE, G2MSMCuda)(scalars, points, msm_size, config, out);
   }
 
+  cudaError_t
+  _G2MSM(const scalar_t* scalars, const dummy_g2_t* points, int msm_size, MSMConfig& config, dummy_g2_t* out)
+  {
+    scalar_t* scalars_host = static_cast<scalar_t*>(malloc(msm_size * sizeof(scalar_t)));
```
yshekel, line 65: note that we don't release the memory

icicle/tests/polynomial_test.cu
```diff
@@ -905,6 +968,21 @@ TEST_F(PolynomialTest, Groth16)
...
+  // (1) construct R1CS and QAP for circuit with N inputs
+  Groth16Example<scalar_t, affine_t, projective_t, dummy_g2_t, dummy_g2_t> groth16_example(30 /*=N*/);
+
+  // (2) compute witness: randomize inputs and compute other entries [1,out,...N inputs..., ... intermediate values...]
+  auto witness = groth16_example.random_witness_inputs();
+  groth16_example.compute_witness(witness);
+
+  groth16_example.setup();
+  auto proof = groth16_example.prove(witness);
+  assert(groth16_example.dummy_verify(proof, witness));
```
yshekel, line 983: use gtest ASSERT_EQ instead. Note that assert() is not executing in release mode.

icicle/tests/polynomial_test.cu
DmytroTym: I think I have a habit of releasing CUDA memory by now, but not CPU memory lol

icicle/tests/polynomial_test.cu
```diff
@@ -20,6 +20,27 @@ using curve_config::scalar_t;
 #include "appUtils/msm/msm.cuh"
 #include "utils/device_context.cuh"
 
+class dummy_g2_t: public scalar_t {
```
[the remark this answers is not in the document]
DmytroTym: `to_affine`, `from_affine` and `generator`

PR #454, opened 2024-04-04
another approach

ChickenLover:


icicle/CMakeLists.txt
yshekel, line 1: is EXT_FIELD slowing the field compilation? if not maybe default to true? for fields at least. G2 for curve maybe not

icicle/src/fields/field.cu
```diff
@@ -1,6 +1,11 @@
+#ifdef CURVE_ID
```
yshekel, line 1: maybe duplicate the file and avoid the coupling? it's so small anyway and not sharing much

icicle/include/curve_or_field.cuh
```diff
@@ -0,0 +1,7 @@
+#ifdef CURVE_ID
```
yshekel, line 1: pragma once

yshekel (commented):
looks better to me

yshekel (approved):

PR #457, opened 2024-04-04
Baby bear

DmytroTym:


icicle/include/fields/field.cuh
```diff
@@ -247,10 +247,10 @@ public:
   add_sub_u32_device(const uint32_t* x, const uint32_t* y, uint32_t* r, size_t n = (TLC >> 1))
   {
     r[0] = SUBTRACT ? ptx::sub_cc(x[0], y[0]) : ptx::add_cc(x[0], y[0]);
-    for (unsigned i = 1; i < (CARRY_OUT ? n : n - 1); i++)
+    for (unsigned i = 1; i < n; i++)
       r[i] = SUBTRACT ? ptx::subc_cc(x[i], y[i]) : ptx::addc_cc(x[i], y[i]);
     if (!CARRY_OUT) {
-      r[n - 1] = SUBTRACT ? ptx::subc(x[n - 1], y[n - 1]) : ptx::addc(x[n - 1], y[n - 1]);
+      ptx::addc(0, 0);
```
ChickenLover, line 253: What is the meaning of doing that?

icicle/include/fields/stark_fields/baby_bear.cuh
```diff
@@ -0,0 +1,89 @@
...
+    // TODO: we're very confused by plonky3 and risc0 having different nonresidues: 11 and -11 respectively
+    static constexpr bool nonresidue_is_negative = true;
+  };
+} // namespace baby_bear
+
+/**
+ * This method multiplies `as` and `bs` and writes the (wide) result into `rs`.
+ */
+template<> __device__ __forceinline__ void
+Field<baby_bear::fp_config>::multiply_raw_device(const ff_storage& as, const ff_storage& bs, ff_wide_storage& rs)
```
ChickenLover, line 58: Is this implementation unique for baby bear or it can be used for other stark curves?

icicle/tests/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+if(DEFINED CURVE)
+  if(DEFINED FIELD)
+    message( FATAL_ERROR "CURVE and FIELD cannot be defined at the same time" )
+  endif ()
+  check_curve()
+elseif(DEFINED FIELD)
+  check_field()
+endif ()
+
+set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -DG2=ON")
```
ChickenLover, line 21: Maybe it's worth moving G2 to an option here?

icicle/tests/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
+cmake_minimum_required(VERSION 3.18)
+
+include(../cmake/Common.cmake)
+include(../cmake/CurvesCommon.cmake)
+include(../cmake/FieldsCommon.cmake)
+
+set_env()
+set_gpu_env()
+
+project(icicle_benchmarks LANGUAGES CUDA CXX)
```
ChickenLover, line 10: icicle_tests?

ChickenLover (approved):
Overall looks very good. I like what you did with unifying ids for curve/fields. Still need to update the CI though.

icicle/benchmarks/field_benchmarks.cu
```diff
@@ -0,0 +1,62 @@
+#include <benchmark/benchmark.h>
+#include "utils/test_functions.cuh"
+#include "fields/field_config.cuh"
+
+using namespace field_config;
+
+static void BM_FieldMul(benchmark::State& state) {
```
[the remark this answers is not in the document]
DmytroTym: Can add that as well, I guess for larger fields multiplication is expected to be the bottleneck by far but for smaller fields addition can be comparably expensive

icicle/include/fields/field.cuh
DmytroTym: It's just a way to accommodate add and sub for 1-limb numbers (such as baby bear)

icicle/include/fields/stark_fields/baby_bear.cuh
DmytroTym: For 1 limb curves, yes. For 64 bits (eg goldilocks), we'd need four products (`a[0]*b[0]`, `a[1]*b[0]`, `a[0]*b[1]`, `a[1]*b[1]`). Actually, I'll just move this implementation to `field.cuh` making it more generic so that no extra implementations are needed for correctness for small fields. Though performance-wise, optimal implementations will still need to be specialised

icicle/tests/CMakeLists.txt
DmytroTym: Yes that's very fair. Curve tests have been written before grumpkin was added and they just assume that every curve has G2 and import names like `g2_affine_t` and `g2_projective_t` without conditional compilation. So it was just easier to enable G2 in cmake, but really I should make G2 conditional in tests, will do that

icicle/include/curves/params/grumpkin.cuh
```diff
@@ -3,9 +3,10 @@
 #define GRUMPKIN_PARAMS_H
 
 #include "fields/storage.cuh"
-#include "fields/params/grumpkin_fields.cuh"
+#include "fields/snark_fields/bn254_scalar.cuh"
 
 namespace grumpkin {
+  typedef bn254::fp_config fq_config;
```
ChickenLover, line 9: What about 
`typedef bn254::fq_config fp_config;` ?

PR #461, opened 2024-04-07
Fix Golang TestNttDeviceAsync

LeonHibnik:
## Describe the changes

This PR fixes TestNttDeviceAsync by adding a missing call to initDomain

## Linked Issues

Resolves #

### DIFF
--- a/wrappers/golang/curves/bls12377/ntt_test.go
+++ b/wrappers/golang/curves/bls12377/ntt_test.go
@@ -120,6 +120,7 @@ func TestECNtt(t *testing.T) {
 
 func TestNttDeviceAsync(t *testing.T) {
 	cfg := GetDefaultNttConfig()
+	initDomain(largestTestSize, cfg)
 	scalars := GenerateScalars(1 << largestTestSize)
 
 	for _, size := range []int{1, 10, largestTestSize} {
--- a/wrappers/golang/curves/bls12377/ntt_test.go
+++ b/wrappers/golang/curves/bls12377/ntt_test.go
@@ -77,6 +77,12 @@ func TestNTTGetDefaultConfig(t *testing.T) {
 	assert.ElementsMatch(t, cosetGenField.GetLimbs(), actual.CosetGen)
 }
 
+func TestInitDomain(t *testing.T) {
+	t.Skip("Skipped because each test requires the domain to be initialized before running. We ensure this using the init() function")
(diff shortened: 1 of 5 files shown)


ChickenLover (approved):

wrappers/golang/curves/bls12377/ntt_test.go
```diff
@@ -120,6 +120,7 @@ func TestECNtt(t *testing.T) {
 
 func TestNttDeviceAsync(t *testing.T) {
 	cfg := GetDefaultNttConfig()
+	initDomain(largestTestSize, cfg)
```
jeremyfelder, line 123: Was this test consistently failing when running the entire test file or only when running this test?

LeonHibnik: Consistently on running only this test, randomly on CI/running all the tests

jeremyfelder (commented):
I think it might be a better idea to just call `initDomain` in an init function for the test file so that any individual test that is ran will not have this issue and when running the entire test file, we should ignore the InitDomain test since it it effectively being tested via other tests

wrappers/golang/curves/bls12377/ntt_test.go
jeremyfelder: > Consistently on running only this test

This makes sense. All the other tests as well that rely on initDomain will have the same issue.

> randomly on CI/running all the tests

I guess the `go test` engine doesn't have a deterministic order for calling tests... i'll look into it but I think my suggestion [here](https://github.com/ingonyama-zk/icicle/pull/461#pullrequestreview-1984884792) is likely the best solution for now

wrappers/golang/curves/bls12377/ntt_test.go
```diff
@@ -72,11 +77,6 @@ func TestNTTGetDefaultConfig(t *testing.T) {
 	assert.ElementsMatch(t, cosetGenField.GetLimbs(), actual.CosetGen)
 }
 
-func TestInitDomain(t *testing.T) {
-	cfg := GetDefaultNttConfig()
-	assert.NotPanics(t, func() { initDomain(largestTestSize, cfg) })
-}
-
```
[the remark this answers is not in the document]
LeonHibnik: Good idea

jeremyfelder (approved):

PR #463, opened 2024-04-08
[FEAT]: golang device slice ranges

jeremyfelder:
## Describe the changes

This PR adds the capability to slice a DeviceSlice, allowing portions of data that are already on the device to be reused.

Additionally, this PR removes the need for a HostSlice underlying type to implement a Size function and uses unsafe.Sizeof instead. This together with #407 will allow direct usage of gnark-crypto types with HostSlice without the need for converting to ICICLE types

### DIFF
--- a/wrappers/golang/core/slice.go
+++ b/wrappers/golang/core/slice.go
@@ -59,6 +59,10 @@ func (d *DeviceSlice) Range(start, end int, endInclusive bool) DeviceSlice {
 		panic("Cannot have negative or zero size slices")
 	}
 
+	if end >= d.length {
+		panic("Cannot increase slice size from Range")
+	}
+
 	var newSlice DeviceSlice
 	switch {
 	case start < 0:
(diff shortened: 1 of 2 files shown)


wrappers/golang/core/slice.go
```diff
@@ -54,6 +54,65 @@ func (d DeviceSlice) CheckDevice() {
...
+	newSlice.capacity = newSlice.length * sizeOfElement
+	newSlice.inner = d.inner
+	return newSlice
+}
+
+func (d *DeviceSlice) RangeFrom(start int) DeviceSlice {
+	if start >= d.length {
+		panic("Cannot have negative or zero size slices")
+	}
+
```
nonam3e, line 103: A check should be added here
```
if start < 0 {
    panic("Negative value for start is not supported")
}
```
Also checks inside `Range` from `RangeFrom` and `RangeTo` can be removed optionally

PR #465, opened 2024-04-08
[FEAT] ReleaseDomain API

LeonHibnik:
## Describe the changes

This PR adds a NTT ReleaseDomain API in Golang and Rust

## Linked Issues

Resolves #


wrappers/rust/icicle-core/src/ntt/tests.rs
```diff
@@ -27,6 +30,14 @@ where
     }
 }
 
+pub fn rel_domain<F: FieldImpl>(ctx: &DeviceContext)
+where
+    <F as FieldImpl>::Config: NTT<F>,
+{
+    // let ctx = DeviceContext::default_for_device(device_id);
```
ChickenLover, line 37: commented code

ChickenLover (approved):

wrappers/rust/icicle-core/src/ntt/tests.rs
LeonHibnik: Removed

yshekel (approved):
I just had a comment about how ReleaseDomain is tested. I guess you already know that so I think it's up to you to decide what to do.

wrappers/golang/curves/bls12377/ntt_test.go
```diff
@@ -201,6 +201,11 @@ func TestNttBatch(t *testing.T) {
 	}
 }
 
+func TestReleaseDomain(t *testing.T) {
+	cfg := GetDefaultNttConfig()
+	ReleaseDomain(cfg.Ctx)
```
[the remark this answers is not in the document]
jeremyfelder: >I think that ideally the test should release an allocated domain, call NTT and assert the error.

It should be possible to `assert.Panics(Ntt())` after releasing the domain.

wrappers/golang/curves/bls12377/ntt_test.go
```diff
@@ -201,6 +201,11 @@ func TestNttBatch(t *testing.T) {
 	}
 }
 
+func TestReleaseDomain(t *testing.T) {
```
jeremyfelder, line 204: We need to ensure that this tests is a teardown part of the tests otherwise other tests that run after it will fail having no domain allocated

wrappers/rust/icicle-core/src/ntt/tests.rs
```diff
@@ -27,6 +30,13 @@ where
     }
 }
 
+pub fn rel_domain<F: FieldImpl>(ctx: &DeviceContext)
```
yshekel, line 33: if this is the exposed API I would rename to release_domain

jeremyfelder (changes requested):
The tests for golang are not deterministic in run order. The ReleaseDomain test should be made into a teardown test to ensure it runs after all other tests that require it

wrappers/rust/icicle-core/src/ntt/tests.rs
jeremyfelder: Its a test function so I think the naming is fine but it doesn't need to be public since it is only being used in https://github.com/ingonyama-zk/icicle/pull/465/files#diff-96ad6a0ca42317faaf6524296cdaa92fcc3c634d36e2882c2b322d9b7287fd84R378

wrappers/golang/curves/bls12377/ntt_test.go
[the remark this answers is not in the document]
LeonHibnik: Fixed

jeremyfelder (approved):
Looks good

PR #466, opened 2024-04-09
Integrate polynomial API to V2

yshekel:
New polynomial API, implementing the design document we reviewed.
Includes:
(1) C++ template API that is backend (GPU, ZPU or other) agnostic (polynomials.h)
(2) concrete CUDA implementation (polynomial_cuda_backend.cu/cuh)
(3) fully tested (polynomial_test.cu)
(4) C API for FFI with documentation and simple test (polynomials_c_api.h)
(5) Groth16 example - setup and prove only (missing G12 and pairings). Currently under tests. Will be moved to be an example in next examples PR.
(6) icicle library is now built for tests as well. Polynomial tests are linked to icicle lib.


icicle/src/polynomials/cuda_backend/kernels.cuh
```diff
@@ -0,0 +1,104 @@
+#pragma once
+#include "stdint.h"
+#include "../../../src/vec_ops/vec_ops.cu" // TODO Yuval: avoid this
```
ChickenLover, line 3: Can we work around it?

icicle/src/polynomials/cuda_backend/polynomial_cuda_backend.cu
```diff
@@ -0,0 +1,873 @@
...
+      const bool is_memset_zeros = coefficients == nullptr;
+      allocate(nof_coefficients, State::Coefficients, is_memset_zeros);
+      if (coefficients) {
+        const bool is_ptr_on_host = is_host_ptr(coefficients);
+
+        CHK_STICKY(cudaMemcpyAsync(
+          m_storage, coefficients, nof_coefficients * sizeof(C),
+          is_ptr_on_host ? cudaMemcpyHostToDevice : cudaMemcpyDeviceToDevice, m_device_context.stream));
+        CHK_STICKY(
+          cudaStreamSynchronize(m_device_context.stream)); // protect against coefficients being released too soon
```
ChickenLover, line 151: wdym by that? Why can't we let the user to decide if he wants async here?

ChickenLover (approved):
Great job as always!

icicle/src/polynomials/cuda_backend/kernels.cuh
yshekel: Yes I can expose them like the add/sub/mul by decalring in vec_ops.cuh and intantiate in vec_ops/extern.cu. I was just being a little lazy.
Alternatively I can move them back to the cuda_backend/kernels.cuh like it was.
It is mainly a question of what we want to expose in vec_ops I think. What we don't I will have in the cuda_backend//kernels.cuh and what we do I can expose like add/sub/mul.

So I will do that I think.

icicle/src/polynomials/cuda_backend/polynomial_cuda_backend.cu
yshekel: The problem I want to avoid is that the host data is still valid when copying.
I could give the user an 'async' flag so that he is aware that he must keep the data alive. The thing I don't like about that is that he would later need to call some blocking sync on the polynomial and it must be device agnostic.
So something like 

Polynomial p = Polynomial::from_coefficients(host_coeffs_p, nof_coeffs, true /*=async*/);
...
...
p.sync(); // or sync(p) or something like that
// can now release host_coeffs_p

We could do that yes. What do you think? I can also add it later since a last param that defaults to false will not break any user code.

ChickenLover: From our experience with zkwasm integration - async is a crucial part of achieving maximal throughput in applications. Even if we talk about ZPU, I think async will still be relevant.

But yeah, maybe we can add it later.

PR #470, opened 2024-04-10
Tests and benches for extension fields

DmytroTym:
- C++ side tests and benchmarks refactored using templates;
- Extension field tests and benchmarks added;
- README added for benches;
- Some small additions to extension field API for convenience;

One issue that I'm having is the following error during `cargo test`:

> Invalid argument : by: NTT at: icicle/icicle/src/ntt/ntt.cu:624 error: NTT size=16 is too large for the domain. Consider generating your domain with a higher order root of unity.

Though it's also happening in the V2 branch for me, so maybe it was introduced by changes around release domain or EC NTT? Might be related to https://github.com/ingonyama-zk/icicle/pull/469


LeonHibnik:
@DmytroTym the issue you had was caused by the release_domain test, #469 fixes it by assuring that the domain release test is running last
Not sure why CI tests are failing here

ChickenLover (approved):

PR #473, opened 2024-04-12
remove generics

nonam3e:
## Describe the changes

This PR removes unnecessary generic types from golang binding functions


wrappers/golang/core/slice.go
```diff
@@ -11,6 +11,7 @@ type HostOrDeviceSlice interface {
 	Cap() int
 	IsEmpty() bool
 	IsOnDevice() bool
+	GetPointerSafe() unsafe.Pointer
```
jeremyfelder, line 14: Rename to `AsUnsafePointer`

wrappers/golang/core/slice.go
```diff
@@ -54,6 +55,11 @@ func (d DeviceSlice) CheckDevice() {
 	}
 }
 
+func (d DeviceSlice) GetPointerSafe() unsafe.Pointer {
+	d.CheckDevice()
```
jeremyfelder, line 59: I'm not sure this is needed here. Its possible to just use this function to pass the pointer around and not use it immediately.

wrappers/golang/core/slice.go
```diff
@@ -188,6 +194,14 @@ func (h HostSlice[T]) SizeOfElement() int {
 	return int(unsafe.Sizeof(h[0]))
 }
 
+func (h HostSlice[T]) AsPointer() unsafe.Pointer {
+	return unsafe.Pointer(&h[0])
+}
```
jeremyfelder, line 199: As discussed, only return a pointer to the first element and let the function below be used to get the unsafe.Pointer around it

jeremyfelder (changes requested):
Nice and clean 💪🏻 
Left a few small comments

wrappers/golang/core/slice.go
jeremyfelder: I think this should be removed.

PR #474, opened 2024-04-13
Feat/vlad/transpose go binding

vladfdp:
## Describe the changes

This PR adds a golang binding for the transpose matrix kernel in vec_ops.go.
There is also a test for it in vec_ops_test

## Linked Issues

Resolves #

### DIFF
--- a/wrappers/golang/curves/bls12377/vec_ops.go
+++ b/wrappers/golang/curves/bls12377/vec_ops.go
@@ -55,7 +55,7 @@ func VecOp(a, b, out core.HostOrDeviceSlice, config core.VecOpsConfig, op core.V
 }
 
 func TransposeMatrix(in, out core.HostOrDeviceSlice, columnSize, rowSize int, ctx cr.DeviceContext, onDevice, isAsync bool) (ret core.IcicleError) {
-
+	core.TransposeCheck(in, out, onDevice)
 	var inPointer, outPointer unsafe.Pointer
 	if onDevice {
 		inPointer = in.(core.DeviceSlice).AsPointer()
--- a/wrappers/golang/curves/bls12377/vec_ops.go
+++ b/wrappers/golang/curves/bls12377/vec_ops.go
@@ -53,3 +53,26 @@ func VecOp(a, b, out core.HostOrDeviceSlice, config core.VecOpsConfig, op core.V
 
 	return ret
 }
+
+func TransposeMatrix(in, out core.HostOrDeviceSlice, columnSize, rowSize int, ctx cr.DeviceContext, onDevice, isAsync bool) (ret core.IcicleError){
+
+	var inPointer, outPointer unsafe.Pointer
+	if onDevice {
+		inPointer = in.(core.DeviceSlice).AsPointer()
+		outPointer = out.(core.DeviceSlice).AsPointer()
+	} else {
+		inPointer = unsafe.Pointer(&in.(core.HostSlice[ScalarField])[0])
+		outPointer = unsafe.Pointer(&out.(core.HostSlice[ScalarField])[0])
+	}
+	cIn := (*C.scalar_t)(inPointer)
+	cOut := (*C.scalar_t)(outPointer)
+
+	cCtx := (*C.DeviceContext)(unsafe.Pointer(&ctx))
+	cRowSize := (C.int)(rowSize)
+	cColumnSize := (C.int)(columnSize)
+	cOnDevice := (C._Bool)(onDevice)
+	cIsAsync := (C._Bool)(isAsync)
+
+	err := (cr.CudaError)(C.bls12_377TransposeMatrix( cIn, cRowSize, cColumnSize, cOut, cCtx, cOnDevice, cIsAsync))
+	return core.FromCudaError(err)
+} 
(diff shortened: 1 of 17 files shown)


wrappers/golang/curves/bls12377/vec_ops.go
```diff
@@ -53,3 +53,26 @@ func VecOp(a, b, out core.HostOrDeviceSlice, config core.VecOpsConfig, op core.V
...
+func TransposeMatrix(in, out core.HostOrDeviceSlice, columnSize, rowSize int, ctx cr.DeviceContext, onDevice, isAsync bool) (ret core.IcicleError) {
+
+	var inPointer, outPointer unsafe.Pointer
+	if onDevice {
+		inPointer = in.(core.DeviceSlice).AsPointer()
+		outPointer = out.(core.DeviceSlice).AsPointer()
+	} else {
+		inPointer = unsafe.Pointer(&in.(core.HostSlice[ScalarField])[0])
+		outPointer = unsafe.Pointer(&out.(core.HostSlice[ScalarField])[0])
+	}
```
jeremyfelder, line 66: I would ensure that both in and out are indeed on the device using `IsOnDevice()` interface method and also adding `checkDevice` to ensure it is being used on the same device that the memory is located on. 

Maybe the above logic should be in a TransposeCheck function in the core package?

wrappers/golang/curves/bls12377/vec_ops_test.go
```diff
@@ -31,3 +32,37 @@ func TestVecOps(t *testing.T) {
...
+
+	TransposeMatrix(matrix, out, columnSize, rowSize, ctx, onDevice, isAsync)
+	TransposeMatrix(out, out2, rowSize, columnSize, ctx, onDevice, isAsync)
+
+	assert.Equal(t, matrix, out2)
+
+	var d_matrix, d_out, d_out2 core.DeviceSlice
+	onDevice = true
+
+	matrix.CopyToDeviceAsync(&d_matrix, *ctx.Stream, true)
```
jeremyfelder, line 58: We should create a new stream since the default stream is the "main" thread on the GPU.

wrappers/golang/curves/bls12377/vec_ops_test.go
```diff
@@ -31,3 +32,37 @@ func TestVecOps(t *testing.T) {
...
+
+	matrix := GenerateScalars(rowSize * columnSize)
+
+	out := make(core.HostSlice[ScalarField], rowSize*columnSize)
+	out2 := make(core.HostSlice[ScalarField], rowSize*columnSize)
+
+	ctx, _ := cr.GetDefaultDeviceContext()
+
+	TransposeMatrix(matrix, out, columnSize, rowSize, ctx, onDevice, isAsync)
+	TransposeMatrix(out, out2, rowSize, columnSize, ctx, onDevice, isAsync)
```
jeremyfelder, line 51: Since `isAsync` is `true`,  a `syncStream` should be called before performing anything else after the async calls to ensure that the operations have finished

wrappers/golang/curves/bls12377/vec_ops_test.go
```diff
@@ -31,3 +32,37 @@ func TestVecOps(t *testing.T) {
...
+	matrix.CopyToDeviceAsync(&d_matrix, *ctx.Stream, true)
+	d_out.MallocAsync(columnSize*rowSize*matrix.SizeOfElement(), matrix.SizeOfElement(), *ctx.Stream)
+	d_out2.MallocAsync(columnSize*rowSize*matrix.SizeOfElement(), matrix.SizeOfElement(), *ctx.Stream)
+
+	TransposeMatrix(d_matrix, d_out, columnSize, rowSize, ctx, onDevice, isAsync)
+	TransposeMatrix(d_out, d_out2, rowSize, columnSize, ctx, onDevice, isAsync)
+	output := make(core.HostSlice[ScalarField], rowSize*columnSize)
+	output.CopyFromDeviceAsync(&d_out2, *ctx.Stream)
+
+	assert.Equal(t, matrix, output)
```
jeremyfelder, line 67: Since `isAsync` is `true`,  a `syncStream` should be called before performing anything else after the async calls to ensure that the operations have finished.

wrappers/golang/curves/bls12377/vec_ops_test.go
vladfdp: Yes my bad I only set isAsync to true to see if it would break stuff

wrappers/golang/curves/bls12377/vec_ops.go
vladfdp: right so I should make a transposeCheck function in core that asserts that onDevice is consistent with IsOnDevice(), performs a checkDevice, and picks the right pointer type.
Do I just return the pointers and assign them to inPointer and outPointer ?

jeremyfelder: >Do I just return the pointers and assign them to inPointer and outPointer ?

For now we just perform the checks in `core.xxxCheck` and access the correct pointer type in the curve's function. @nonam3e has an open [PR](https://github.com/ingonyama-zk/icicle/pull/473) to change how we access the pointer, so we will move to that method once it is merged

wrappers/rust/Cargo.toml
```diff
@@ -14,7 +14,7 @@ exclude = [
 ]
 
 [workspace.package]
-version = "1.9.1"
+version = "1.10.0"
```
jeremyfelder, line 17: Revert

jeremyfelder (approved):
I changed the base branch to be v2 since we are all merging against there now.
Small change to revert but otherwise looks good

PR #476, opened 2024-04-14
Feat/roman/c api headers

ChickenLover:


icicle/include/api/babybear/field/api_ext.h
```diff
@@ -0,0 +1,21 @@
+// WARNING: This file is auto-generated by a script.
```
yshekel, line 1: why split ext?

icicle/include/api/babybear/field/api_ext.h
```diff
@@ -0,0 +1,21 @@
...
+// Any changes made to this file may be overwritten.
+// Please modify the code generation script instead.
+// Path to the code generation script: icicle/include/api/gen_c_api.py
```
yshekel, line 4: I would put it under icicle/scripts then

icicle/include/api/gen_c_api.py
```diff
@@ -0,0 +1,89 @@
...
+    "ecntt"
+}
+
+FIELDS_MODULES = {
+    "field",
+    "ntt",
+    "poseidon",
+}
+
+CURVES_CONFIG = {
```
yshekel, line 23: please explain what the dicts for curves and fields are. I can see it from the code below but would be clearer to have it explained here

icicle/include/api/gen_c_api.py
```diff
@@ -0,0 +1,89 @@
...
+    # Generate API for ingo_curve
+    for curve, skip in CURVES_CONFIG.items():
+        for module in CURVES_MODULES | FIELDS_MODULES - set(skip):
+            module_templates_path = TEMPLATES_PATH.joinpath(module)
+
+            module_path = API_PATH.joinpath(curve).joinpath(module)
+            module_path.mkdir(parents=True, exist_ok=True)
+
+            for header_template in module_templates_path.iterdir():
+                if "ext" in header_template.name and "ext" in skip:
```
yshekel, line 55: would be better to use a regex and find any of the skipped keywords in the filename instead of hardcoding them like this

icicle/include/api/gen_c_api.py
```diff
@@ -0,0 +1,89 @@
...
+                    continue
+
+                if "g2" in header_template.name and "g2" in skip:
+                    continue
+
+                contents = open(header_template).read()
```
yshekel, line 61: should use 'with open...' instead of open() directly, especially since you don't close the files.

ChickenLover: why?

icicle/include/api/gen_c_api.py
ChickenLover: not sure I got what you mean

yshekel (changes requested):
Overall I think it's good but I think we should consider having less files.
For example instead of api/bls12_377/curve/api.h and api/bls12_377/curve/api_g2.h and api/bls12_377/ntt/api.h and so on just have c_api/bls12_377.h which expose all relevant APIs including NTT, MSM and so on. Even if you declare APIs that are not defined, it's not a big deal since it's a linker error but the same is true if you include the api_g2.h without building with G2. I think it would make it simpler.

In termf of your script, I think it should not be more complex if you keep the same tempaltes and you conact instead of having a new file

icicle/include/api/babybear/field/api_ext.h
ChickenLover: Just trying to follow the same structure we have in the codebase. But I guess we can merge in a single file, fine

icicle/include/api/gen_c_api.py
jeremyfelder: `with open` auto closes the file once the with block ends.

icicle/include/api/gen_c_api.py
```diff
@@ -0,0 +1,89 @@
...
+                contents = contents.replace("${FIELD}", curve)
+                contents = contents.replace("${FIELD_UPPER}", curve.upper())
+                contents = contents.replace("${FIELD_HEADER}", f"fields/snark_fields/{curve}_scalar.cuh")
+                contents = contents.replace("${CURVE}", curve)
+                contents = contents.replace("${CURVE_UPPER}", curve.upper())
```
jeremyfelder, line 66: Using the strings.Template lib might be simpler to maintain and read:
https://docs.python.org/3/library/string.html#string.Template.safe_substitute

icicle/include/api/gen_c_api.py
ChickenLover: yeah, I know, but why do we need it here?:)

yshekel: It's generally recommended to use context-managers when handling resoruce such as files in Python. It's more robust since it closes the file even when an exception is raised. It's also clearer since you can see the scope where the file is used and you don't have to close it explicitly.
I think that calling open() directly should only be done when you cannot use a context manager.

yshekel (approved):
Looks good to me. One question that is not really related: why do we keep having both pragma once and the guard-defs?

ChickenLover:
@yshekel afaik it's because `#pragma once` is not supported in some (very very old) compilers. But I actually think that we might have more potential bugs with def-guards than it will every be helpful to anyone. Actually we did have a bug in main related to the improper def guard naming

jeremyfelder (approved):
Small update for golang tests in the download path which we spoke about. Otherwise looks 🔥 👍🏻

PR #477, opened 2024-04-14
Feat/roman/v2 examples

ChickenLover:
## Describe the changes

This PR...

## Linked Issues

Resolves #


examples/c++/msm/CMakeLists.txt
```diff
@@ -8,18 +8,17 @@ if (${CMAKE_VERSION} VERSION_LESS "3.24.0")
...
+
 add_executable(
   example
   example.cu
 )
-
-find_library(NVML_LIBRARY nvidia-ml PATHS /usr/local/cuda-12.0/targets/x86_64-linux/lib/stubs/ )
-target_link_libraries(example ${NVML_LIBRARY})
+target_include_directories(example PRIVATE "../../../icicle/include")
+target_link_libraries(example ${CMAKE_SOURCE_DIR}/../../../icicle/build/lib/libingo_curve_bn254.a)
```
yshekel, line 22: isn't it enough to link to the curve lib?

examples/c++/msm/CMakeLists.txt
```diff
@@ -8,18 +8,17 @@ if (${CMAKE_VERSION} VERSION_LESS "3.24.0")
...
-include_directories("../../../icicle")
+
 add_executable(
   example
   example.cu
 )
-
-find_library(NVML_LIBRARY nvidia-ml PATHS /usr/local/cuda-12.0/targets/x86_64-linux/lib/stubs/ )
-target_link_libraries(example ${NVML_LIBRARY})
+target_include_directories(example PRIVATE "../../../icicle/include")
```
yshekel, line 21: Can we avoid this relative path? Maybe use some env variable to tell where icicle is? Ideally should add_subdirectory()

examples/c++/ntt/compile.sh
```diff
@@ -3,9 +3,12 @@
 # Exit immediately on error
 set -e
 
-rm -rf build
 mkdir -p build
-cmake -S . -B build
-cmake --build build
 
+# Configure and build Icicle
```
yshekel, line 8: Maybe use another script that is building icicle and exports the build/include dir and source it from here

jeremyfelder: Isn't this way more realistic right now since each integration needs to build from source for their own needs? I don't imagine them having two separate build scripts

jeremyfelder (approved):

examples/c++/ntt/compile.sh
yshekel: I think there are two options: (1) build (or install with the include dir) in some directory and link to that which is how it is now or (2) do it directly via cmake (FetchContent).

examples/c++/msm/CMakeLists.txt
ChickenLover: The idea was to maybe have pre-built libraries in the future. So add_subdirectory will not necessarily be needed. Actually, there are multiple ways for the user to link with Icicle. But I agree with using variable for Icicle path. Just remember that this is just an example and not the actual app

examples/c++/msm/CMakeLists.txt
ChickenLover: yes, you are right

yshekel (approved):
Approved. Just wondering about the extension fields: will you add them later?

PR #481, opened 2024-04-17
[FEAT]: Babybear for golang

jeremyfelder:
## Describe the changes

This PR:
- Moves all curve/field agnostic checks to `core` package
- Adds babybear field
  - Changed field representation back to using 32 bit integers
- Adds grumpkin curve
- Changes the generation and structure of the golang bindings
  - There is a configuration file for each field and curve
  - Each additional functionality of the API is generated on its own per field or curve and is a separate package within the field or curve package. This allows integrations to only compile what they need and not all of our supported primitives.
- Generating files can be done using `go generate ./...` from any directory within the project


.github/workflows/golang.yml
```diff
@@ -33,13 +33,23 @@ jobs:
...
+          - name: bn254
+            build_args: -g2 -ecntt
+          - name: bls12_381
+            build_args: -g2 -ecntt
+          - name: bls12_377
+            build_args: -g2 -ecntt
+          - name: bw6_761
+            build_args: -g2 -ecntt
+          - name: grumpkin
+            build_args:
```
ChickenLover, line 52: TODO Roman: implement the same for our c++ workflow (rn it just doesn't test grumpkin at all)

wrappers/golang/build.sh
```diff
@@ -57,4 +91,17 @@ do
...
+# Needs to remove the CMakeCache.txt file to allow building fields after curves
+# have been built since CURVE and FIELD cannot both be defined
+rm -f "$BUILD_DIR/CMakeCache.txt"
+
+for FIELD in "${BUILD_FIELDS[@]}"
+do
+  echo "FIELD=${FIELD}" > build_config.txt
+  echo "DEVMODE=${DEVMODE}" >> build_config.txt
+  cmake -DCMAKE_CUDA_COMPILER=$CUDA_COMPILER_PATH -DFIELD=$FIELD -DEXT_FIELD=$EXT_FIELD -DDEVMODE=$DEVMODE -DCMAKE_BUILD_TYPE=Release -S . -B build
+  cmake --build build -j8 && rm build_config.txt
```
ChickenLover, line 105: `-j8` => `-j`

wrappers/golang/internal/generator/ecntt/templates/ecntt.h.tmpl
```diff
@@ -1,5 +1,4 @@
 #include <cuda_runtime.h>
-#include "../../../include/types.h"
```
ChickenLover, line 2: how go knows about NTTConfig and projective_t if there is nothing else included?

wrappers/golang/internal/generator/lib_linker/generate.go
```diff
@@ -0,0 +1,32 @@
+package libe_linker
```
ChickenLover, line 1: typo?

ChickenLover (approved):
Mostly reviewed the c++ part and the new code gen script. Configs looks nice.

wrappers/golang/internal/generator/ecntt/templates/ecntt.h.tmpl
jeremyfelder: They're partially declared lower in this file: https://github.com/ingonyama-zk/icicle/pull/481/files/2043ed13721067aba9e71ee381a41fd402a8958d#diff-ebd41d829ee4e442d0136366489b1f9d02faf670ac9553cdf7b7f21386bde13fR10-R11

wrappers/golang/internal/generator/lib_linker/generate.go
jeremyfelder: Yes, nice catch 👍🏻

wrappers/golang/build.sh
jeremyfelder: Why do we want to use the build tool's default number of jobs?

LeonHibnik: What's the default number here?
Maybe it's better to set it to use all the cpu cores

ChickenLover: I actually thought that -j will use num-cpus threads

jeremyfelder: https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-build-j. Lets leave as is for now unless we have a reason to change it

PR #485, opened 2024-04-21
[FEAT]: Poly API bindings for golang

jeremyfelder:
## Describe the changes

This PR add golang bindings for the Poly API.

In order to support running tests for both ntt and poly APis, tests have been moved to their own package under each curve/field. This allows us to run all the tests in between a curve/field specific setup and teardown.

TODO:
- [ ] Add DeviceSlice views into Polys


wrappers/golang/internal/generator/config/bls12377.go
```diff
@@ -5,7 +5,7 @@ func init() {
 		PackageName:         "bls12377",
 		Curve:               "bls12_377",
 		GnarkImport:         "bls12-377",
-		SupportsPoseidon:    true,
```
ChickenLover, line 8: I think you accidentally removed poseidon

wrappers/golang/internal/generator/config/grumpkin.go
```diff
@@ -5,6 +5,7 @@ func init() {
 		PackageName:         "grumpkin",
 		Curve:               "grumpkin",
 		GnarkImport:         "",
+		SupportsPoly:        false,
```
ChickenLover, line 8: but why?

wrappers/golang/internal/generator/ntt/generate.go
```diff
@@ -13,12 +13,6 @@ var nttTemplates = map[string]string{
 	"header":       "ntt/templates/ntt.h.tmpl",
 }
 
-var ecnttTemplates = map[string]string{
```
ChickenLover, line 16: What happened here? Are they not needed anymore?

jeremyfelder: it was a left over that wasn't removed. Ecntt generator was moved to its own folder

LeonHibnik (approved):

PR #496, opened 2024-04-24
Nonam3e/golang/keccak

nonam3e:
## Describe the changes

This PR adds keccak bindings + passes cfg as reference in keccak cuda functions

### DIFF
--- a/icicle/src/hash/keccak/keccak.cu
+++ b/icicle/src/hash/keccak/keccak.cu
@@ -245,7 +245,7 @@ namespace keccak {
       CHK_IF_RETURN(cudaMallocAsync(&output_device, number_of_blocks * (D / 8), stream));
     }
 
-    int number_of_threads = 1024;
+    int number_of_threads = 512;
     int number_of_gpu_blocks = (number_of_blocks - 1) / number_of_threads + 1;
     keccak_hash_blocks<C, D><<<number_of_gpu_blocks, number_of_threads, 0, stream>>>(
       input_device, input_block_size, number_of_blocks, output_device);
--- a/icicle/src/hash/keccak/keccak.cu
+++ b/icicle/src/hash/keccak/keccak.cu
@@ -224,7 +224,7 @@ namespace keccak {
 
   template <int C, int D>
   cudaError_t
-  keccak_hash(uint8_t* input, int input_block_size, int number_of_blocks, uint8_t* output, KeccakConfig config)
+  keccak_hash(uint8_t* input, int input_block_size, int number_of_blocks, uint8_t* output, KeccakConfig& config)
   {
     CHK_INIT_IF_RETURN();
     cudaStream_t& stream = config.ctx.stream;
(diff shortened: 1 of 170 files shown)


wrappers/golang/internal/generator/mock/generate.go
```diff
@@ -25,7 +25,7 @@ func Generate(baseDir, packageName, field, fieldPrefix string, isScalar bool, nu
 		isScalar,
 		numLimbs,
 	}
-	generator.GenerateFile(mockTemplates["field"], baseDir, strings.ToLower(field)+"_", "", fieldData)
+	generator.GenerateFile(mockTemplates["fields"], baseDir, strings.ToLower(field)+"_", "", fieldData)
```
jeremyfelder, line 28: Revert

icicle/src/hash/keccak/keccak.cu
```diff
@@ -245,7 +245,7 @@ namespace keccak {
       CHK_IF_RETURN(cudaMallocAsync(&output_device, number_of_blocks * (D / 8), stream));
     }
 
-    int number_of_threads = 1024;
+    int number_of_threads = 512;
```
jeremyfelder, line 248: Why is this being reduced?

wrappers/golang/hash/keccak/hasher.go
```diff
@@ -0,0 +1,70 @@
...
+	ctx, _ := cr.GetDefaultDeviceContext()
+	return KeccakConfig{
+		ctx,
+		false,
+		false,
+		false,
+	}
+}
+
+func KeccakCheck(input core.HostOrDeviceSlice, output core.HostOrDeviceSlice, cfg *KeccakConfig) (unsafe.Pointer, unsafe.Pointer, unsafe.Pointer) {
```
jeremyfelder, line 31: This can be a private method since it is only used by Keccak256 and Keccak512
```suggestion
func keccakCheck(input core.HostOrDeviceSlice, output core.HostOrDeviceSlice, cfg *KeccakConfig) (unsafe.Pointer, unsafe.Pointer, unsafe.Pointer) {
```

Also change the calls in Keccak512 and Keccak256

wrappers/golang/hash/keccak/hasher.go
```diff
@@ -0,0 +1,70 @@
...
+	cInput := (*C.uint8_t)(inputPointer)
+	cOutput := (*C.uint8_t)(outputPointer)
+	cInputBlockSize := (C.int)(inputBlockSize)
+	cNumberOfBlocks := (C.int)(numberOfBlocks)
+	cConfig := (*C.KeccakConfig)(cfgPointer)
+
+	__ret := C.keccak256_cuda(cInput, cInputBlockSize, cNumberOfBlocks, cOutput, cConfig)
+	err := (cr.CudaError)(__ret)
+	return core.FromCudaError(err)
+}
```
jeremyfelder, line 57: Since the only difference between Keccak512 and Keccak256 functions here is the C function name, does it make sense to have a single function with a switch case/if else similar to https://github.com/ingonyama-zk/icicle/blob/main/wrappers/golang/curves/bn254/vecOps/vec_ops.go#L23-L30

jeremyfelder (commented):
Some minor questions and comments

wrappers/golang/hash/keccak/hasher.go
ChickenLover: let's make a single internal function that will be called by Keccak256 and Keccak512

icicle/src/hash/keccak/keccak.cu
ChickenLover: Keccak kernel requires a lot of registers, so running with 1024 threads won't fit the GPU

jeremyfelder: Was this working before hand though?

ChickenLover (approved):

jeremyfelder (approved):

PR #497, opened 2024-04-24
fix: bug regarding MixedRadix coset NTT for NM/MN ordering

yshekel:
The bug was in how twiddles array is indexed when multiplied by a mixed (M) vector to implement (I)NTT on cosets.
The fix is to use the DIF-digit-reverse to know which element in the natural (N) vector, moved to index 'i' in the M vector. This is emulating a DIT-digit-reverse (which is mixing like a DIF-compute) reorder of the twiddles array and element-wise multiplication without reordering the twiddles memory.

### DIFF
--- a/icicle/include/ntt/ntt_impl.cuh
+++ b/icicle/include/ntt/ntt_impl.cuh
@@ -32,6 +32,7 @@ namespace mxntt {
     S* external_twiddles,
     S* internal_twiddles,
     S* basic_twiddles,
+    S* linear_twiddle, // twiddles organized as [1,w,w^2,...] for coset-eval in fast-tw mode
     int ntt_size,
     int max_logn,
     int batch_size,
--- a/icicle/src/ntt/kernel_ntt.cu
+++ b/icicle/src/ntt/kernel_ntt.cu
@@ -141,14 +141,9 @@ namespace mxntt {
     if (tid >= size * batch_size) return;
     int64_t scalar_id = (tid / columns_batch_size) % size;
     if (rev_type != eRevType::None) {
-      // Notes:
-      // (1) DIF (I)NTT is mixing digits (N->M) such that a DIF reverse complemets it back to N. Same for DIT where a
-      // DIT reverse complements the DIT compute. Obviously DIF<->DIT complement each other. Therefore is should be
-      // clear that DIF is mixing like a DIT reverse and vice versa.
-      // (2) to multiply a Mixed-order in_vec by scalars, need to mix the scalars corresponding to how the in_vec was
-      // mixed. Mixing N->M is done by DIF (I)NTT. Therefore for multiplication in M state always use DIT reversal.
+      const bool dif = rev_type == eRevType::NaturalToMixedRev;
       scalar_id =
-        generalized_rev((tid / columns_batch_size) & ((1 << log_size) - 1), log_size, true /*=dit*/, fast_tw, rev_type);
+        generalized_rev((tid / columns_batch_size) & ((1 << log_size) - 1), log_size, !dif, fast_tw, rev_type);
     }
     out_vec[tid] = *(scalar_vec + ((scalar_id * step) % n_scalars)) * in_vec[tid];
   }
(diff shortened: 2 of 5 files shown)


DmytroTym:
Your change fixes correctness issues for `fast_twiddles=false`, but they remain for fast twiddle version on sizes 13, 14, 16 (exactly the sizes on which stages are different for fast twiddles). I tried passing `fast_tw` flag into `batch_elementwise_mul_with_reorder_kernel` (currently it always uses false) but that didn't help either.

yshekel:
> Your change fixes correctness issues for `fast_twiddles=false`, but they remain for fast twiddle version on sizes 13, 14, 16, 19 (exactly the sizes on which stages are different for fast twiddles). I tried passing `fast_tw` flag into `batch_elementwise_mul_with_reorder_kernel` (currently it always uses false) but that didn't help either.

Yes you are right. I also was thinking about it some more and got confsued. I think that I pretty much undesrtand the issue but something is still not making total sense to me. Will keep working on that.

DmytroTym:
> About the fast twiddle, I think it's because they cannot be used in this function that multiplies since they are organized differently (in memory), I believe it should work if you passed the ordered ones to this function (in addition to the fast twiddle that are used by the NTT I mean)

Does it really matter how fast twiddles are organised in memory if the issue is probably in reordering, and not in ntt itself (which is where we actually use twiddles)? I'm guessing it's reordering because orderings other than `NM` and `MN` work fine.

yshekel:
> > About the fast twiddle, I think it's because they cannot be used in this function that multiplies since they are organized differently (in memory), I believe it should work if you passed the ordered ones to this function (in addition to the fast twiddle that are used by the NTT I mean)
> 
> Does it really matter how fast twiddles are organised in memory if the issue is probably in reordering, and not in ntt itself (which is where we actually use twiddles)? I'm guessing it's reordering because orderings other than `NM` and `MN` work fine.

I think the bug has to do with the logic that multiplies the coeffs with the twiddles for when we evaluate a coset or INTT from a coset. I mean the logic that is mixing the twiddles and multiplying by the coeffs.

Anyway, I am trying to convince myself then I will post here an update.

yshekel:
Ok I think I got it now. @DmytroTym @HadarIngonyama please tell if this makes sense to you :)

The root cause of the problem is that doing a digit-reverse for DIT is not equivalent to a DIF digi-reverse. The reason is quite simple. Consider a 3b number 'abc' constructed by 2 digits (one is 2b and the other 1b). If you reverse where first digit is 2b you get 'bca' but if first digit is 1b you get 'cab'.

What is important to understand is that DIF is mixing the inputs in a way that is the inverse of what DIT is doing, thus DIF followed by DIT is NN.

Consider the case NN NTT (one way): we can either do it via DIF compute (N->M) followed by DIF reverse (M->N) or we can do a DIT reverse (N-M) followed by a DIT (M->N) compute.

You can see that a DIT reverse is mixing the same as a DIF compute is mixing and similar for DIT compute and DIF reverse. 

Now when we evaluate a coset we first need to multiply the coeffs by some modulation (the powers of the coset gen). For INTT it's the opposite and done after the computation. I think the easiest way to see why is in the matrix form. consider x the vector, F the NTT matrix and D the diagonal modulation then you compute y=FDx in the forward way so backward is Dinv Finv y.

Anyway, the bug was that when you do an NM INTT (which is done via DIF) need to mix by this modulation (twiddles for subgroup case)  the same way the DIF NTT did which is a DIT reversal, not DIF reversal! Then the following MN NTT needs to also use the DIT reverse to match how the inputs where previously mixed. This was fine since it is a DIT compute.

Note that starting from M for some data would be wrong but anyway the definition of M is the mixing that a DIF is doing and it depends on alg (mixed vs radix2), the size, and fast-tw mode so should never assume that you have generated M form but only get it from an NM (I)NTT which is a DIF.

**Last thing that I would be happy if you can confirm: I believe that both DIT and DIF would make the test pass when mixing the twiddles for the coset since they are both conssistent and so the MN NTT is doing the opposite of the NM INTT in both cases. BUT the INTT result would be wrong in that case. So please confirm that DIT is then right reorder of the twiddles based on your understanding. If wrong then the computation is wrong but maybe it would be unnoticed**.

BTW do you know why do an NM on coset? the only case I think about it dividing by the vanishing poly which is zero if not using a coset so can do it via NM NTT followed by element-wise division followed by MN INTT which will not have the bug that was fixed here.

DmytroTym (approved):

PR #506, opened 2024-05-08
fix: bug regarding polynomial evaluations view in CUDA backend

yshekel:
fixing:
(1) not building polynomials and tests for grumpkin curve (no NTT)
(2) polynomial API C++ example compilation and (when compilation is fixed) memory corruption
(3) bug fix in poly CUDA backend regarding transformation to evaluations in some cases

### DIFF
--- a/examples/c++/polynomial-api/example.cu
+++ b/examples/c++/polynomial-api/example.cu
@@ -254,7 +254,8 @@ void example_clone(const int log0)
   std::cout << "h(x) = " << h(x) << " expected: " << g(x) << std::endl;
 }
 
-void example_EvenOdd() {
+void example_EvenOdd()
+{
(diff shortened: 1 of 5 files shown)


examples/c++/polynomial-api/example.cu
```diff
@@ -254,7 +254,8 @@ void example_clone(const int log0)
   std::cout << "h(x) = " << h(x) << " expected: " << g(x) << std::endl;
 }
 
-void example_EvenOdd() {
+void example_EvenOdd()
```
ChickenLover, line 257: what's up with the casing here?

ChickenLover (approved):

examples/c++/polynomial-api/example.cu
yshekel: fixed for the entire example

LeonHibnik (approved):

PR #507, opened 2024-05-08
Feat/roman/poseidon2

ChickenLover:
This PR adds support for poseidon2 permutation function as described in https://eprint.iacr.org/2023/323.pdf

Reference implementations used (and compared against):
https://github.com/HorizenLabs/poseidon2/tree/main
https://github.com/Plonky3/Plonky3/tree/main

Tasks:

- [x] Remove commented code and prints
- [ ] Add doc-comments to functions and structs
- [x] Fix possible issue with Plonky3 imports
- [x] Update NTT/Plonky3 test
- [x] Add Plonky3-bn254 test (impossible)


wrappers/rust/icicle-core/src/field.rs
```diff
@@ -82,6 +83,12 @@ impl<const NUM_LIMBS: usize, F: FieldConfig> FieldImpl for Field<NUM_LIMBS, F> {
         Self::from(limbs)
     }
 
+    fn from_hex(s: &str) -> Self {
+        let mut bytes = Vec::from_hex(&s[2..]).expect("Invalid hex string");
```
jeremyfelder, line 87: Should we add a check that the what we are chopping is indeed a `0x`?

ChickenLover: good catch, will add. I just copy-pasted this function from HorizonLabs for test purposes

icicle/src/poseidon2/extern.cu
```diff
@@ -0,0 +1,63 @@
...
+      P2_HASH_T(4)
+      P2_HASH_T(8)
+      P2_HASH_T(12)
+      P2_HASH_T(16)
+      P2_HASH_T(20)
+      P2_HASH_T(24)
+    default:
+      THROW_ICICLE_ERR(
+        IcicleError_t::InvalidArgument, "PoseidonHash: #arity must be one of [2, 3, 4, 8, 12, 16, 20, 24]");
+    }
```
jeremyfelder, line 60: Is the only reason for using this flow to check that the arity is of a supported arity or am I missing something else?

ChickenLover: the reason is to invoke different functions. width is a template parameter

ChickenLover: I could've made it like so
```
case 2:                                                                                                          
    return poseidon2_hash<scalar_t, 2>(input, output, number_of_states, *constants, *config);
case 3:                                                                                                          
    return poseidon2_hash<scalar_t, 3>(input, output, number_of_states, *constants, *config);
 ...
```

But I think it's too much boilerplate, so instead I used a macro here

wrappers/rust/icicle-curves/icicle-bn254/src/poseidon2/mod.rs
```diff
@@ -0,0 +1,35 @@
+use crate::curve::{ScalarCfg, ScalarField};
+
+use icicle_core::error::IcicleResult;
+use icicle_core::impl_poseidon2;
+use icicle_core::poseidon2::{DiffusionStrategy, MdsType, Poseidon2, Poseidon2Config, Poseidon2Constants};
+use icicle_core::traits::IcicleResultWrap;
+use icicle_cuda_runtime::device_context::DeviceContext;
+use icicle_cuda_runtime::error::CudaError;
+use icicle_cuda_runtime::memory::HostOrDeviceSlice;
```
jeremyfelder, line 9: Can we move these to the poseidon2 mod inside the `impl_poseidon2` macro?

wrappers/rust/icicle-fields/icicle-babybear/Cargo.toml
```diff
@@ -8,7 +8,7 @@ homepage.workspace = true
 repository.workspace = true
 
 [dependencies]
-icicle-core = { workspace = true }
+icicle-core = { workspace = true, features = [] }
```
jeremyfelder, line 11: Why is this necessary if there are no default feature on `core`?

wrappers/rust/icicle-curves/icicle-bn254/src/poseidon2/mod.rs
ChickenLover: we don't do that for any other primitive. So I guess not

wrappers/rust/icicle-fields/icicle-babybear/Cargo.toml
ChickenLover: This is just a left over from when I did testing against plonky3. It actually compiled ark-ec even without the ark feature. Probably will remove that after the CI pass

ChickenLover: And ark-ec doesn't compile with plonky3

wrappers/rust/icicle-curves/icicle-bn254/src/poseidon2/mod.rs
jeremyfelder: Ok, i guess its out of scope for this PR but its just more confusing boilerplate in each field/curve crate that can be moved to the core crate

ChickenLover: Yes, I totally agree with you

LeonHibnik (approved):

icicle/include/poseidon2/poseidon2.cuh
```diff
@@ -0,0 +1,130 @@
...
+    DiffusionStrategy diffusion;
+  };
+
+  /**
+   * @struct Poseidon2Config
+   * Struct that encodes various Poseidon2 parameters.
+   */
+  struct Poseidon2Config {
+    device_context::DeviceContext ctx; /**< Details related to the device such as its id and stream id. */
+    bool are_states_on_device;  /**< True if inputs are on device and false if they're on host. Default value: false. */
```
yshekel, line 60: minor but the comment says 'inputs' where the field is called 'states'

icicle/src/poseidon2/constants.cu
```diff
@@ -0,0 +1,120 @@
...
+#include "poseidon2/constants/grumpkin_poseidon2.h"
+using namespace poseidon2_constants_grumpkin;
+#elif FIELD_ID == BABY_BEAR
+#include "poseidon2/constants/babybear_poseidon2.h"
+using namespace poseidon2_constants_babybear;
+#endif
+
+namespace poseidon2 {
+  template <typename S>
+  cudaError_t create_optimized_poseidon2_constants(
```
yshekel, line 27: is there a way to release the constants that you allocate here? what is the intention here in terms of memory management?

icicle/src/poseidon2/kernels.cu
```diff
@@ -0,0 +1,233 @@
...
+    case 3:
+      return result2 * element;
+    case 5:
+      return S::sqr(result2) * element;
+    case 7:
+      return S::sqr(result2) * result2 * element;
+    case 11:
+      S result8 = S::sqr(S::sqr(result2));
+      return result8 * result2 * element;
+    }
```
yshekel, line 19: maybe add assert(false) in default case so the GPU at least throws some error instead of silently do nothing. I assume you check it earlier but still it wouldn't hurt to assert if here too.

icicle/src/poseidon2/poseidon.cu
```diff
@@ -0,0 +1,76 @@
...
+    poseidon2_permutation_kernel<S, T>
+      <<<poseidon_number_of_blocks<S, T>(number_of_states), poseidon_block_size, 0, stream>>>(
+        states, states_out, number_of_states, constants);
+    CHK_IF_RETURN(cudaPeekAtLastError());
+    return CHK_LAST();
+  }
+
+  template <typename S, int T>
+  cudaError_t poseidon2_hash(
+    S* states,
```
yshekel, line 27: better be const S* if states is not modified

yshekel (commented):
Looking good, wrote a few comments

icicle/include/poseidon2/poseidon2.cuh
ChickenLover: yes, comments are yet to be added, I want to do that tomorrow, for now it's fine I think

icicle/src/poseidon2/poseidon.cu
ChickenLover: states can be modified

icicle/src/poseidon2/constants.cu
ChickenLover: no, currently there is no way to release the constants. The intention was to store all the constants on-device like we do with twiddles

LeonHibnik: @yshekel good point here, we need to let users to control the free memory on the device, to be solved in next update

jeremyfelder (approved):

PR #513, opened 2024-05-12
Feat/vlad/poseidon go binding

vladfdp:
## Describe the changes

This PR adds a golang binding for poseidon.
Also it adds a test to verify that it runs for every curve and, for bls12-381 specifically, it checks correctness against a precomputed value.
The reason I didn't test every curve (or the different arities/number of states) is I couldn't find a library that supports everything we do to compare against.
I'll add tests for bn254 and bls12-371 later.

## Linked Issues

Resolves #

### DIFF
--- a/icicle/src/poseidon/Makefile
+++ b/icicle/src/poseidon/Makefile
@@ -1,3 +1,2 @@
-test_poseidon: test.cu poseidon.cu kernels.cu constants.cu
-	nvcc -o test_poseidon -I../../include -DFIELD_ID=2 -DCURVE_ID=2 test.cu
-	./test_poseidon
+test_poseidon : test.cu poseidon.cu kernels.cu constants.cu nvcc - o test_poseidon - I../../ include - DFIELD_ID =
+  2 - DCURVE_ID = 2 test.cu./ test_poseidon
--- a/wrappers/golang/core/poseidon.go
+++ b/wrappers/golang/core/poseidon.go
@@ -63,9 +63,14 @@ func PoseidonCheck[T any](input, output HostOrDeviceSlice, cfg *PoseidonConfig, 
 		panic(errorString)
 	}
 
-	if inputLen != arity*numberOfStates {
+	expectedInputLen := arity * numberOfStates
+	if cfg.InputIsAState {
+		expectedInputLen += numberOfStates
+	}
+
+	if inputLen != expectedInputLen {
 		errorString := fmt.Sprintf(
-			"input is not the right length for the given parameters:%d, should be: %d",
+			"input is not the right length for the given parameters: %d, should be: %d",
 			inputLen,
 			arity*numberOfStates,
 		)
@@ -74,7 +79,7 @@ func PoseidonCheck[T any](input, output HostOrDeviceSlice, cfg *PoseidonConfig, 
 
 	if outputLen != numberOfStates {
 		errorString := fmt.Sprintf(
-			"output is not the right length for the given parameters:%d, should be: %d",
+			"output is not the right length for the given parameters: %d, should be: %d",
 			outputLen,
 			numberOfStates,
 		)
--- a/wrappers/golang/core/poseidon.go
+++ b/wrappers/golang/core/poseidon.go
@@ -1,0 +1,77 @@
+package core
+
+import (
+	"fmt"
+	"unsafe"
+
+	cr "github.com/ingonyama-zk/icicle/v2/wrappers/golang/cuda_runtime"
+)
+
+type PoseidonConfig struct {
+	/// Details related to the device such as its id and stream id. See [DeviceContext](@ref device_context::DeviceContext).
+	Ctx                cr.DeviceContext
+	areInputsOnDevice  bool
+	areOutputsOnDevice bool
+	///If true, input is considered to be a states vector, holding the preimages in aligned or not aligned format.
+	///Memory under the input pointer will be used for states. If false, fresh states memory will be allocated and input will be copied into it */
+	InputIsAState bool
+	/// If true - input should be already aligned for poseidon permutation.
+	///* Aligned format: [0, A, B, 0, C, D, ...] (as you might get by using loop_state)
+	///* not aligned format: [A, B, 0, C, D, 0, ...] (as you might get from cudaMemcpy2D) */
+	Aligned bool
+	///If true, hash results will also be copied in the input pointer in aligned format
+	LoopState bool
+	///Whether to run the Poseidon asynchronously. If set to `true`, the poseidon_hash function will be
+	///non-blocking and you'd need to synchronize it explicitly by running `cudaStreamSynchronize` or `cudaDeviceSynchronize`.
+	///If set to false, the poseidon_hash function will block the current CPU thread. */
+	IsAsync bool
+}
+
+type PoseidonConstants[T any] struct {
+	Arity           int32
+	PartialRounds   int32
+	FullRoundsHalf  int32
+	RoundConstants  unsafe.Pointer
+	MdsMatrix       unsafe.Pointer
+	NonSparseMatrix unsafe.Pointer
+	SparseMatrices  unsafe.Pointer
+	DomainTag       T
+}
+
+func GetDefaultPoseidonConfig() PoseidonConfig {
+	ctx, _ := cr.GetDefaultDeviceContext()
+	return PoseidonConfig{
+		ctx,   // Ctx
+		false, // areInputsOnDevice
+		false, // areOutputsOnDevice
+		false, // inputIsAState
+		false, // aligned
+		false, // loopState
+		false, // IsAsync
+	}
+}
+
+func PoseidonCheck(input, output HostOrDeviceSlice, cfg *PoseidonConfig) (unsafe.Pointer, unsafe.Pointer, unsafe.Pointer) {
+
+	cfg.areInputsOnDevice = input.IsOnDevice()
+	cfg.areOutputsOnDevice = output.IsOnDevice()
+
+	fmt.Println("TODO: PoseidonCheck")
+	fmt.Println("input is on device", input.IsOnDevice())
+	fmt.Println("output is on device", output.IsOnDevice())
+	fmt.Println("cfg says input:", cfg.areInputsOnDevice, " and output:", cfg.areOutputsOnDevice)
+
+	if input.IsOnDevice() {
+		input.(DeviceSlice).CheckDevice()
+
+	}
+
+	if output.IsOnDevice() {
+		output.(DeviceSlice).CheckDevice()
+
+	}
+
+	cfgPointer := unsafe.Pointer(cfg)
+
+	return input.AsUnsafePointer(), output.AsUnsafePointer(), cfgPointer
+}
--- a/wrappers/golang/core/poseidon.go
+++ b/wrappers/golang/core/poseidon.go
@@ -51,18 +51,9 @@ func GetDefaultPoseidonConfig() PoseidonConfig {
 	}
 }
 
-func PoseidonCheck[T any](input, output HostOrDeviceSlice, cfg *PoseidonConfig, constants *PoseidonConstants[T], arity, numberOfStates int) (unsafe.Pointer, unsafe.Pointer, unsafe.Pointer) {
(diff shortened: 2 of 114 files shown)


LeonHibnik (approved):

icicle/src/poseidon/Makefile
ChickenLover, line 1: What's up with this change? I think it wasn't intended

wrappers/golang/internal/generator/poseidon/templates/poseidon_test.go.tmpl
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"testing"
+	
+	core "github.com/ingonyama-zk/icicle/v2/wrappers/golang/core"
+	cr "github.com/ingonyama-zk/icicle/v2/wrappers/golang/cuda_runtime"
+	{{.Field}} "github.com/ingonyama-zk/icicle/v2/wrappers/golang/{{.BaseImportPath}}"
+	poseidon "github.com/ingonyama-zk/icicle/v2/wrappers/golang/{{.BaseImportPath}}/poseidon"
+
+	{{if eq .Field "bls12_381"}}
```
ChickenLover, line 11: I think you can remove this debug now?

wrappers/golang/fields/babybear/extension/extension_field.go
```diff
@@ -6,10 +6,9 @@ import "C"
 import (
 	"encoding/binary"
 	"fmt"
-	"unsafe"
-
 	"github.com/ingonyama-zk/icicle/v2/wrappers/golang/core"
 	cr "github.com/ingonyama-zk/icicle/v2/wrappers/golang/cuda_runtime"
+	"unsafe"
```
ChickenLover, line 11: Why was it changed?

wrappers/golang/fields/babybear/ntt/ntt.go
```diff
@@ -5,13 +5,15 @@ package ntt
...
 
 import (
-	"unsafe"
-
 	"github.com/ingonyama-zk/icicle/v2/wrappers/golang/core"
 	cr "github.com/ingonyama-zk/icicle/v2/wrappers/golang/cuda_runtime"
 	babybear "github.com/ingonyama-zk/icicle/v2/wrappers/golang/fields/babybear"
 )
 
+import (
```
ChickenLover, line 13: Same here

wrappers/golang/core/poseidon.go
```diff
@@ -0,0 +1,98 @@
...
+	if arity != int(constants.Arity) {
+		errorString := fmt.Sprintf(
+			"arity: %d does not match Poseidon Constants: %d",
+			arity,
+			outputLen,
+		)
+		panic(errorString)
+	}
+
+	if inputLen != arity*numberOfStates {
```
jeremyfelder, line 66: Does the expected input length depend on whether `cfg.InputIsAState` is true? See https://github.com/ingonyama-zk/icicle/blob/main/wrappers/rust/icicle-core/src/poseidon/mod.rs#L160-L166
cc @ChickenLover

wrappers/golang/core/poseidon.go
```diff
@@ -0,0 +1,98 @@
...
+		errorString := fmt.Sprintf(
+			"input is not the right length for the given parameters:%d, should be: %d",
```
jeremyfelder, line 68: nit 
```suggestion
			"input is not the right length for the given parameters: %d, should be: %d",
```

wrappers/golang/core/poseidon.go
jeremyfelder: Should we allow an input length that is larger than `arity*numStates`

wrappers/golang/core/poseidon.go
```diff
@@ -0,0 +1,98 @@
...
+			inputLen,
+			arity*numberOfStates,
+		)
+		panic(errorString)
+	}
+
+	if outputLen != numberOfStates {
+		errorString := fmt.Sprintf(
+			"output is not the right length for the given parameters:%d, should be: %d",
```
jeremyfelder, line 77: nit
```suggestion
			"output is not the right length for the given parameters: %d, should be: %d",
```

wrappers/golang/curves/bls12377/tests/poseidon_test.go
```diff
@@ -0,0 +1,42 @@
...
+	core "github.com/ingonyama-zk/icicle/v2/wrappers/golang/core"
+	cr "github.com/ingonyama-zk/icicle/v2/wrappers/golang/cuda_runtime"
+	bls12_377 "github.com/ingonyama-zk/icicle/v2/wrappers/golang/curves/bls12377"
+	poseidon "github.com/ingonyama-zk/icicle/v2/wrappers/golang/curves/bls12377/poseidon"
+)
+
+func TestPoseidon(t *testing.T) {
+
+	arity := 2
+	numberOfStates := 1
```
jeremyfelder, line 15: Is this sufficient to show that the test is testing what we want?

wrappers/golang/curves/bls12377/tests/poseidon_test.go
```diff
@@ -0,0 +1,42 @@
...
+	scalarsCopy := core.HostSliceFromElements(scalars[:numberOfStates*arity])
+
+	stream, _ := cr.CreateStream()
+
+	var deviceInput core.DeviceSlice
+	scalarsCopy.CopyToDeviceAsync(&deviceInput, stream, true)
+	var deviceOutput core.DeviceSlice
+	deviceOutput.MallocAsync(numberOfStates*scalarsCopy.SizeOfElement(), scalarsCopy.SizeOfElement(), stream)
+
+	poseidon.PoseidonHash(deviceInput, deviceOutput, numberOfStates, arity, &cfg, &constants) //run Hash function
```
jeremyfelder, line 37: Should this be running async on the same stream as deviceInput and deviceOutput?

wrappers/golang/curves/bls12377/poseidon/poseidon.go
```diff
@@ -0,0 +1,57 @@
...
+	return core.GetDefaultPoseidonConfig()
+}
+
+func PoseidonHash[T any](scalars, results core.HostOrDeviceSlice, numberOfStates, arity int, cfg *core.PoseidonConfig, constants *core.PoseidonConstants[T]) core.IcicleError {
+	scalarsPointer, resultsPointer, cfgPointer := core.PoseidonCheck(scalars, results, cfg, constants, arity, numberOfStates)
+
+	cScalars := (*C.scalar_t)(scalarsPointer)
+	cResults := (*C.scalar_t)(resultsPointer)
+	cNumberOfStates := (C.int)(numberOfStates)
+	cArity := (C.int)(arity)
```
jeremyfelder, line 24: Why don't we use constants.Arity?

wrappers/golang/curves/bls12377/poseidon/poseidon.go
```diff
@@ -0,0 +1,57 @@
...
+	cFullRoundsHalfs := (C.int)(fullRoundsHalfs)
+	cPartialRounds := (C.int)(partialRounds)
+	cConstants := (*C.scalar_t)(constants.AsUnsafePointer())
+	cCtx := (*C.DeviceContext)(unsafe.Pointer(&ctx))
+	cPoseidonConstants := (*C.PoseidonConstants)(unsafe.Pointer(poseidonConstants))
+
+	__ret := C.bls12_377_create_optimized_poseidon_constants_cuda(cArity, cFullRoundsHalfs, cPartialRounds, cConstants, cCtx, cPoseidonConstants)
+	err := (cr.CudaError)(__ret)
+	return core.FromCudaError(err)
+}
```
jeremyfelder, line 46: @ChickenLover Whats the use case for calling this instead of `InitOptimizedPoseidonConstantsCuda`?

wrappers/golang/internal/generator/poseidon/generate.go
```diff
@@ -0,0 +1,34 @@
...
+		FieldPrefix    string
+		BaseImportPath string
+	}{
+		"poseidon",
+		field,
+		fieldPrefix,
+		baseDir,
+	}
+
+	testPath := poseidonTemplates["test"]
```
jeremyfelder, line 29: Not necessary to create a new var here

wrappers/golang/internal/generator/poseidon/templates/poseidon_test.go.tmpl
jeremyfelder: Its being used to create the result string from limbs to test against.

wrappers/golang/core/poseidon.go
vladfdp: > Does the expected input length depend on whether `cfg.InputIsAState` is true? See https://github.com/ingonyama-zk/icicle/blob/main/wrappers/rust/icicle-core/src/poseidon/mod.rs#L160-L166 cc @ChickenLover

I see, I'll fix it

wrappers/golang/curves/bls12377/tests/poseidon_test.go
vladfdp: well I would have liked to test for different length and arities but I couldn't find an implementation to check against that does everything. Especially for numberOfStates it's hard to find a library where it's not set to 1.
I'll add more tests later on, just need to find a library (in another language most likely) that support several curves and parameters

wrappers/golang/curves/bls12377/poseidon/poseidon.go
vladfdp: In poseidonCheck , called just before, we check that arity = constants.Arity

wrappers/golang/internal/generator/poseidon/generate.go
vladfdp: yeah not sure why it's there. Most likely copy and pasted without paying too much attention, will fix.

wrappers/golang/curves/bls12377/poseidon/poseidon.go
vladfdp: made a binding because there was an extern link. figured there's a use case for it

wrappers/golang/fields/babybear/ntt/ntt.go
jeremyfelder: Its a difference in gofmt versions. Even though it is in our template in two separate imports, the latest gofmt should move them to a single one. Using go version 1.20.X causes these changes and Italy is using 1.20.4

wrappers/golang/curves/bls12377/tests/poseidon_test.go
vladfdp: Do you mean that this not safe because the Hash function might be run on a slice that is not yet allocated ?

jeremyfelder: I guess that as well but also CopyToDeviceAsync might not have finished copying to the device pointer.
Its also odd to have some portion be async and some not in this case

wrappers/golang/core/poseidon.go
vladfdp: > Should we allow an input length that is larger than `arity*numStates`

well the end of the slice won't be taken into account right ? If users don't use it well you might end up with a hash where you can easily get a collision, checking that it is equal seems safer even if not strictly necessary

wrappers/golang/curves/bls12377/tests/poseidon_test.go
ChickenLover: I think using async everywhere is fine. Just make sure it all runs in a single stream

wrappers/golang/curves/bls12377/poseidon/poseidon.go
ChickenLover: I think Jeremy is right here. Less arguments for user to fill in == more error-prone

wrappers/golang/curves/bls12377/poseidon/poseidon.go
ChickenLover: @jeremyfelder `create` allows user to pass his own constants. While `Init` just loads pre-defined ones

ChickenLover (approved):

PR #514, opened 2024-05-12
Yshekel/poly view fix

yshekel:
This PR : 

- removes the poly API to access view of evaluations. This is a problematic API since it cannot handle small domains and for large domains requires the polynomial to use more memory than need to.
- adds evaluate_on_rou_domain() API instead that supports any domain size (powers of two size).
- the new API can compute to HOST or DEVICE memory
- Rust wrapper for evaluate_on_rou_domain()
- updated documentation: overview and Rust wrappers
- faster division by vanishing poly for common case where numerator is 2N and vanishing poly is of degree N.
- allow division a/b where deg(a)<deg(b) instead of throwing an error.

### DIFF
--- a/icicle/include/fields/field.cuh
+++ b/icicle/include/fields/field.cuh
@@ -1001,6 +1001,17 @@ public:
     }
     return (u == one) ? b : c;
   }
+
+  static constexpr HOST_DEVICE_INLINE Field pow(Field base, int exp)
+  {
+    Field res = one();
+    while (exp > 0) {
+      if (exp & 1) res = res * base;
+      base = base * base;
+      exp >>= 1;
+    }
+    return res;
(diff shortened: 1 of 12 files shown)


icicle/include/fields/field.cuh
```diff
@@ -1001,6 +1001,17 @@ public:
     }
     return (u == one) ? b : c;
   }
+
+  static constexpr HOST_DEVICE_INLINE Field pow(Field base, int exp)
+  {
+    Field res = one();
```
ChickenLover, line 1007: Maybe worth using Wide here? To save on reductions

ChickenLover (approved):

LeonHibnik (approved):

PR #518, opened 2024-05-16
Allow overriding compiler's chosen GPU arch via cmake

### LINKED ISSUE #167: [FEAT]: Allow overriding compiler's chosen GPU arch via cmake
## Description

There should be a flexible way to provide/override the compiler's chosen arch. Maybe the following would be best for that:
```
if(DEFINED CUDA_ARCH) // user defined arch takes priority
    set(CMAKE_CUDA_ARCHITECTURES ${CUDA_ARCH})
elseif(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.24.0") // otherwise, use native to detect GPU arch
    set(CMAKE_CUDA_ARCHITECTURES native)
endif() // otherwise fallback to compiler's chosen arch
```

See #159 and #163 for additional details

## Motivation

We should have a flexible way to provide/override the compiler's chosen arch.

liuhao230:
## Describe the changes

This PR modifies icicle/cmake/Common.cmake to set CMAKE_CUDA_ARCHITECTURES to ${CUDA_ARCH} if the user defines the arch, to set CMAKE_CUDA_ARCHITECTURES to native if the cmake version is greater than or equal to 3.24.0. This change has been successfully tested with cmake 3.22.0 and 3.25.2.

## Linked Issues

Resolves #167.

### DIFF
--- a/icicle/cmake/Common.cmake
+++ b/icicle/cmake/Common.cmake
@@ -14,12 +14,12 @@ endfunction()
 function(set_gpu_env)
     # add the target cuda architectures
     # each additional architecture increases the compilation time and output file size
-	if(DEFINED CUDA_ARCH) # user defined arch takes priority
-	    set(CMAKE_CUDA_ARCHITECTURES ${CUDA_ARCH})
-	elseif(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.24.0") # otherwise, use native to detect GPU arch
-	    set(CMAKE_CUDA_ARCHITECTURES native)
+    if(DEFINED CUDA_ARCH) # user defined arch takes priority
+        set(CMAKE_CUDA_ARCHITECTURES ${CUDA_ARCH} PARENT_SCOPE)
+    elseif(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.24.0") # otherwise, use native to detect GPU arch
+        set(CMAKE_CUDA_ARCHITECTURES native PARENT_SCOPE)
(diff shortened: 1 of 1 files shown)


ChickenLover (approved):
Thanks for your contribution @sss1h !
lgtm, but we'll wait for another approve

jeremyfelder (approved):

icicle/cmake/Common.cmake
```diff
@@ -14,10 +14,12 @@ endfunction()
 function(set_gpu_env)
     # add the target cuda architectures
     # each additional architecture increases the compilation time and output file size
-    if(${CMAKE_VERSION} VERSION_LESS "3.24.0")
-    set(CMAKE_CUDA_ARCHITECTURES ${CUDA_ARCH} PARENT_SCOPE)
+	if(DEFINED CUDA_ARCH) # user defined arch takes priority
+	    set(CMAKE_CUDA_ARCHITECTURES ${CUDA_ARCH})
```
yshekel, line 18: did it work for you? why did you remove the PARENT_SCOPE? is it not needed outside this function?
When I print it in the main cmake I get the wrong version:
message("CMAKE_CUDA_ARCHITECTURES=${CMAKE_CUDA_ARCHITECTURES}")

Are you sure it compiled the correct version?

sss1h: The PARENT_SCOPE indeed should be added. Thanks for your review.

PR #519, opened 2024-05-16
[FEAT]: Golang Bindings for pinned host memory

### LINKED ISSUE #43: Pinned host memory
Currently using pageable memory using cudaMalloc. Switching to pinned host memory using cudaMallocHost will perform better with higher bandwidth.

jeremyfelder:
## Describe the changes

This PR adds the capability to pin host memory in golang bindings allowing data transfers to be quicker. Memory can be pinned once for multiple devices by passing the flag `cuda_runtime.CudaHostRegisterPortable` or `cuda_runtime.CudaHostAllocPortable` depending on how pinned memory is called

TODO
- [ ] Rust bindings

## Linked Issues

Resolves #43

### DIFF
--- a/wrappers/rust/icicle-core/src/msm/tests.rs
+++ b/wrappers/rust/icicle-core/src/msm/tests.rs
@@ -2,7 +2,7 @@ use crate::curve::{Affine, Curve, Projective};
 use crate::msm::{msm, precompute_points, MSMConfig, MSM};
 use crate::traits::{FieldImpl, GenerateRandom};
 use icicle_cuda_runtime::device::{get_device_count, set_device, warmup};
-use icicle_cuda_runtime::memory::{DeviceVec, HostSlice};
+use icicle_cuda_runtime::memory::{CudaHostRegisterFlags, DeviceVec, HostSlice};
 use icicle_cuda_runtime::stream::CudaStream;
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
 
--- a/wrappers/rust/icicle-core/src/msm/tests.rs
+++ b/wrappers/rust/icicle-core/src/msm/tests.rs
@@ -133,7 +133,7 @@ where
     }
 
     let scalars = <<C as Curve>::ScalarField as FieldImpl>::Config::generate_random(largest_size);
-    
+
     let mut scalars_d = DeviceVec::<<C as Curve>::ScalarField>::cuda_malloc(test_size).unwrap();
     let stream = CudaStream::create().unwrap();
     scalars_d
@@ -144,13 +144,7 @@ where
     cfg.ctx
         .stream = &stream;
     cfg.is_async = true;
-    msm(
-        &scalars_d[..],
-        points,
-        &cfg,
-        &mut msm_results[..],
-    )
-    .unwrap();
+    msm(&scalars_d[..], points, &cfg, &mut msm_results[..]).unwrap();
 
     let mut msm_host_result = vec![Projective::<C>::zero(); 1];
     msm_results
@@ -171,13 +165,7 @@ where
             allocated_pinned_points[i] = points[i].clone()
         }
 
-        msm(
-            &scalars_d[..],
-            allocated_pinned_points,
-            &cfg,
-            &mut msm_results[..],
-        )
-        .unwrap();
+        msm(&scalars_d[..], allocated_pinned_points, &cfg, &mut msm_results[..]).unwrap();
 
         let mut msm_host_result = vec![Projective::<C>::zero(); 1];
         msm_results
--- a/wrappers/rust/icicle-core/src/msm/tests.rs
+++ b/wrappers/rust/icicle-core/src/msm/tests.rs
@@ -160,12 +160,13 @@ where
         .into();
     assert!(msm_res_affine.is_on_curve());
 
-    let allocated_pinned_points = HostSlice::allocate_pinned(points.len(), CudaHostAllocFlags::DEFAULT).unwrap();
-    allocated_pinned_points
-        .as_mut_slice()
-        .clone_from_slice(points.as_slice());
+    points.allocate_pinned(points.len(), CudaHostAllocFlags::DEFAULT).unwrap();
+    // let allocated_pinned_points = HostSlice::allocate_pinned(points.len(), CudaHostAllocFlags::DEFAULT).unwrap();
+    // allocated_pinned_points
+    //     .as_mut_slice()
+    //     .clone_from_slice(points.as_slice());
 
-    msm(&scalars_d[..], allocated_pinned_points, &cfg, &mut msm_results[..]).unwrap();
+    msm(&scalars_d[..], points, &cfg, &mut msm_results[..]).unwrap();
 
     let mut msm_host_result = vec![Projective::<C>::zero(); 1];
     msm_results
@@ -179,7 +180,7 @@ where
         .to_ark()
         .into();
     assert!(msm_res_affine.is_on_curve());
-    allocated_pinned_points.free_pinned();
+    points.free_pinned();
 
     stream
         .destroy()
--- a/wrappers/rust/icicle-core/src/msm/tests.rs
+++ b/wrappers/rust/icicle-core/src/msm/tests.rs
@@ -2,7 +2,7 @@ use crate::curve::{Affine, Curve, Projective};
 use crate::msm::{msm, precompute_points, MSMConfig, MSM};
 use crate::traits::{FieldImpl, GenerateRandom};
 use icicle_cuda_runtime::device::{get_device_count, set_device, warmup};
-use icicle_cuda_runtime::memory::{CudaHostRegisterFlags, DeviceVec, HostSlice};
+use icicle_cuda_runtime::memory::{CudaHostAllocFlags, CudaHostRegisterFlags, DeviceVec, HostOrDeviceSlice, HostSlice};
 use icicle_cuda_runtime::stream::CudaStream;
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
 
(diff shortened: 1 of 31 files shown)


wrappers/rust/icicle-cuda-runtime/src/memory.rs
```diff
@@ -117,6 +134,66 @@ impl<T> HostSlice<T> {
         self.0
             .iter_mut()
     }
+
+    pub fn is_pinnable(&self) -> bool {
+        let pinnable = get_device_attribute(cudaDeviceAttr::cudaDevAttrHostRegisterSupported, 0).unwrap();
```
yshekel, line 139: why hardcoded device 0?

wrappers/rust/icicle-cuda-runtime/src/memory.rs
```diff
@@ -117,6 +134,66 @@ impl<T> HostSlice<T> {
...
+    pub fn unpin(&self) -> CudaResult<()> {
+        unsafe {
+            let mut flags = 0;
+            let ptr = self.as_ptr() as *mut c_void;
+            cudaHostGetFlags(&mut flags, ptr).wrap()?;
+            cudaHostUnregister(ptr).wrap()
+        }
+    }
+
+    pub fn allocate_pinned(count: usize, flags: CudaHostAllocFlags) -> CudaResult<&'static mut Self> {
```
yshekel, line 163: why is the returned HostSlice with static lifetime?

wrappers/rust/icicle-cuda-runtime/src/memory.rs
```diff
@@ -117,6 +134,66 @@ impl<T> HostSlice<T> {
...
+            let mut flags: u32 = 0;
+            let ptr = self.as_ptr() as *mut c_void;
+            cudaHostGetFlags(&mut flags, ptr).wrap()?;
+            cudaFreeHost(ptr).wrap()
+        }
+    }
+
+    pub fn get_memory_flags(&self) -> CudaResult<u32> {
+        unsafe {
+            let mut flags: u32 = 1234;
```
yshekel, line 191: 1234 is intentional?

wrappers/rust/icicle-cuda-runtime/src/memory.rs
```diff
@@ -117,6 +134,66 @@ impl<T> HostSlice<T> {
...
+
+    pub fn is_pinnable(&self) -> bool {
+        let pinnable = get_device_attribute(cudaDeviceAttr::cudaDevAttrHostRegisterSupported, 0).unwrap();
+        let lockable =
+            get_device_attribute(cudaDeviceAttr::cudaDevAttrPageableMemoryAccessUsesHostPageTables, 0).unwrap();
+
+        pinnable == 1 && lockable == 0
+    }
+
+    pub fn pin(&self, flags: CudaHostRegisterFlags) -> CudaResult<()> {
```
yshekel, line 146: this is basically pinning the memory that the HostSlice refers to right? which is allocated by Rust somehow

wrappers/rust/icicle-core/src/msm/tests.rs
```diff
@@ -105,6 +105,87 @@ where
...
+    stream
+        .synchronize()
+        .unwrap();
+
+    let msm_res_affine: ark_ec::short_weierstrass::Affine<<C as Curve>::ArkSWConfig> = msm_host_result[0]
+        .to_ark()
+        .into();
+    assert!(msm_res_affine.is_on_curve());
+
+    if pin {
```
yshekel, line 162: if pinning not supported the test would still pass? When is it not supported?

wrappers/rust/icicle-core/src/msm/tests.rs
```diff
@@ -105,6 +105,87 @@ where
...
+        let allocated_pinned_points = HostSlice::allocate_pinned(points.len(), CudaHostAllocFlags::DEFAULT).unwrap();
+        for i in 0..points.len() {
+            allocated_pinned_points[i] = points[i].clone()
```
yshekel, line 165: just wondering, isn't a memcopy() equivalent in rust?

wrappers/rust/icicle-core/src/msm/tests.rs
```diff
@@ -105,6 +105,87 @@ where
...
+            .unwrap();
+        stream
+            .synchronize()
+            .unwrap();
+
+        let msm_res_affine: ark_ec::short_weierstrass::Affine<<C as Curve>::ArkSWConfig> = msm_host_result[0]
+            .to_ark()
+            .into();
+        assert!(msm_res_affine.is_on_curve());
+    }
```
yshekel, line 182: no need to call free_pinned() here? what leaks if not? I noticed you did it in the memory.rs test

yshekel (changes requested):
I reviewed the Rust part only

wrappers/rust/icicle-cuda-runtime/src/memory.rs
jeremyfelder: Since the HostSlice that is returned is created in the function and we are returning a reference to it, it needs a static lifetime.

wrappers/rust/icicle-cuda-runtime/src/memory.rs
jeremyfelder: Technically it can be any value above 16 since the bitflags are 0,1,2,4,8 and it needs to be a uint. I'm fine changing it to another number that won't be misinterpreted as a set flag

wrappers/rust/icicle-cuda-runtime/src/memory.rs
```diff
@@ -117,6 +134,66 @@ impl<T> HostSlice<T> {
...
+        let mut pinned_host_ptr = MaybeUninit::<*mut c_void>::uninit();
+
+        unsafe {
+            cudaHostAlloc(pinned_host_ptr.as_mut_ptr(), size, flags.bits).wrap()?;
+            let pinned_host_slice = from_raw_parts_mut(pinned_host_ptr.assume_init() as *mut T, count);
+            Ok(Self::from_mut_slice(pinned_host_slice))
+        }
+    }
+
+    pub fn free_pinned(&self) -> CudaResult<()> {
```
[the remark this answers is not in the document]
jeremyfelder: Yes. Since its not possible (to my knowledge) to know whether the pinned memory was through `pin` or `allocate_pinned` and each one has their own specific free function that needs to be called

wrappers/rust/icicle-cuda-runtime/src/memory.rs
jeremyfelder: Yes exactly. I should add dev comments

wrappers/rust/icicle-core/src/msm/tests.rs
```diff
@@ -105,6 +105,87 @@ where
...
+    let largest_size = 1 << 16;
+    // let test_sizes = [1 << 10, largest_size];
+    let test_size = largest_size;
+    let mut msm_results = DeviceVec::<Projective<C>>::cuda_malloc(1).unwrap();
+    let random_points = generate_random_affine_points_with_zeroes(largest_size, 2);
+    let points: &HostSlice<Affine<C>> = HostSlice::from_slice(&random_points);
+
+    let pin = points.is_pinnable();
+    if pin {
+        // points.pin(CudaHostRegisterFlags::DEFAULT).unwrap();
```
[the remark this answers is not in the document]
jeremyfelder: I will remove this and the `pin`/`unpin` capability for now since it isn't working correctly when the pointer is passed to a primitive

wrappers/rust/icicle-core/src/msm/tests.rs
jeremyfelder: Good catch, `allocate_pinned` should be supported always but `pin` (which calls `cudaHostRegister` under the hood) is not always supported. I will update the test

wrappers/rust/icicle-cuda-runtime/src/memory.rs
jeremyfelder: Removed for now as there were issues with using cudaHostRegister. But in general, you're correct, theres no reason to specifically check device 0 and this should be more robust to ensure some device supports cudaHostRegister

wrappers/rust/icicle-cuda-runtime/src/memory.rs
```diff
@@ -117,6 +134,66 @@ impl<T> HostSlice<T> {
...
+            get_device_attribute(cudaDeviceAttr::cudaDevAttrPageableMemoryAccessUsesHostPageTables, 0).unwrap();
+
+        pinnable == 1 && lockable == 0
+    }
+
+    pub fn pin(&self, flags: CudaHostRegisterFlags) -> CudaResult<()> {
+        unsafe {
+            let ptr = self.as_ptr() as *mut c_void;
+            let flags_to_set = flags.bits();
+            cudaHostRegister(ptr, self.len(), flags_to_set as c_uint).wrap()
```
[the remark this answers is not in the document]
jeremyfelder: Removed for now. cudaHostRegister would fail if the following was false (0). 
```rust
pinnable = get_device_attribute(cudaDeviceAttr::cudaDevAttrHostRegisterSupported, 0).unwrap();
```

wrappers/rust/icicle-cuda-runtime/src/memory.rs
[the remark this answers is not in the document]
yshekel: can you store it as a field in the HostSlice and based on that call free_pinned() when dropped?

LeonHibnik (approved):

wrappers/rust/icicle-cuda-runtime/src/memory.rs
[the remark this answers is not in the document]
yshekel: actually you need to unpin in both cases right?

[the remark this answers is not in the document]
jeremyfelder: > can you store it as a field in the HostSlice and based on that call free_pinned() when dropped?

This was my first thought but it is a breaking change which I was trying to avoid.

>Or is it not dropped since it's a view to memory that is owned by someone else?

HostSlice is a wrapper struct around already allocated host memory so yea owned by someone else. HostSlice can still be dropped but it doesn't drop the underlying slice.
 
>in that case the allocated_pinned() may be problematic

Maybe I missed something here, can you explain? Do you mean it will leak the static underlying slice?

yshekel (approved):

PR #520, opened 2024-05-16
bug fix regarding cmake set_gpu_env() for the case where using nvidia…

yshekel:


LeonHibnik (approved):

.github/changed-files.yml
```diff
@@ -17,6 +17,10 @@ cpp:
   - icicle/**/*.h
   - icicle/CMakeLists.txt
   - .github/workflows/cpp_cuda.yml
+  - icicle/CMakeLists.txt
```
jeremyfelder, line 20: Duplicate line of L#18

icicle/cmake/Common.cmake
```diff
@@ -16,51 +16,40 @@ function(set_gpu_env)
     # each additional architecture increases the compilation time and output file size
     if(DEFINED CUDA_ARCH) # user defined arch takes priority
         set(CMAKE_CUDA_ARCHITECTURES ${CUDA_ARCH} PARENT_SCOPE)
-    elseif(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.24.0") # otherwise, use native to detect GPU arch
+    elseif(UNIX AND ${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.24.0") # otherwise, use native to detect GPU arch
```
jeremyfelder, line 19: If we are supporting windows in general I think this is not the best place to fix the CI windows machine issue. Better to pass a "CUDA_ARCH" as a variable from within the CI job

yshekel: I will fix

.github/changed-files.yml
yshekel: fixed

icicle/cmake/Common.cmake
yshekel: removed

jeremyfelder (approved):

PR #521, opened 2024-05-19
[HOTFIX] rust msm benchmarks

LeonHibnik:
## Describe the changes

removes unused host to device copy, adds minimum limit to run MSM benchmarks


vhnatyk (approved):

jeremyfelder (approved):

yshekel (commented):
Unrelated question but why both lines 415,416 do the same thing and 416 shadows 415? I believe just a bug.

PR #523, opened 2024-05-19
fix: ntt mixed-radix bug regarding large ntts (and/or batch)

yshekel:
in some cases 32b values would wrap around and cause invalid accesses to wrong elements and memory addresses

### DIFF
--- a/icicle/src/ntt/kernel_ntt.cu
+++ b/icicle/src/ntt/kernel_ntt.cu
@@ -927,9 +927,10 @@ namespace mxntt {
   {
     CHK_INIT_IF_RETURN();
 
+    const uint64_t total_nof_elements = uint64_t(ntt_size) * batch_size;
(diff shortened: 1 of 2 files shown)


HadarIngonyama (commented):
looks fine, do these changes affect performance in any way?

yshekel:
> looks fine, do these changes affect performance in any way?

Yes, for some cases (size+batch) it is a little slower. Sometimes slightly faster (which is odd but).
It's probably since now the kernels use u64 which are more expensive instructions or more instructions or something.

I tried to minimize it. For example I initially made the meta.block_id u64 (Which make all ops with it u64) but then reverted and casted only where I need it (so only some ops are u64 and the struct is smaller etc.).

Do you have an idea how to support >4G elements but avoid it? I don't really think it's possible.

ChickenLover (approved):

HadarIngonyama (approved):

PR #525, opened 2024-05-23
Feat/roman/tree builder

ChickenLover:
# Updates:

## Hashing

 - Added SpongeHasher class
 - Can be used to accept any hash function as an argument
 - Absorb and squeeze are now separated
 - Memory management is now mostly done by SpongeHasher class, each hash function only describes permutation kernels

## Tree builder

 - Tree builder is now hash-agnostic. 
 - Tree builder now supports 2D input (matrices)
 - Tree builder can now use two different hash functions for layer 0 and compression layers

## Poseidon1

 - Interface changed to classes
 - Now allows for any alpha
 - Now allows passing constants not in a single vector
 - Now allows for any domain tag
 - Constants are now released upon going out of scope
 - Rust wrappers changed to Poseidon struct
 
 ## Poseidon2
 
 - Interface changed to classes
 - Constants are now released upon going out of scope
 - Rust wrappers changed to Poseidon2 struct
 
## Keccak

 - Added Keccak class which inherits SpongeHasher
 - Now doesn't use gpu registers for storing states
 
 To do:
- [x] Update poseidon1 golang bindings
- [x] Update poseidon1 examples
- [x] Fix poseidon2 cuda test
- [x] Fix poseidon2 merkle tree builder test
- [x] Update keccak class with new design
- [x] Update keccak test
- [x] Check keccak correctness
- [x] Update tree builder rust wrappers
- [x] Leave doc comments

Future work:  
- [ ] Add keccak merkle tree builder externs
- [ ] Add keccak rust tree builder wrappers
- [ ] Write docs
- [ ] Add example
- [ ] Fix device output for tree builder


icicle/include/hash/hash.cuh
```diff
@@ -0,0 +1,257 @@
...
+  __global__ void generic_squeeze_states_kernel(
+    const S* states, unsigned int number_of_states, unsigned int width, unsigned int rate, unsigned int offset, S* out)
+  {
+    int idx = (blockIdx.x * blockDim.x) + threadIdx.x;
+    if (idx >= number_of_states) { return; }
+
+    for (int i = 0; i < rate; i++) {
+      if (states == out) {
+        S element = states[idx * width + offset];
+        __syncthreads();
```
yshekel, line 32: note that this sync is per block

ChickenLover: yeah, I thought it would be fine when I wrote this, but now I see that this indeed will result in race condition

icicle/include/hash/hash.cuh
```diff
@@ -0,0 +1,257 @@
...
+        out[idx * rate + i] = states[idx * width + offset];
+      }
+    }
+  }
+
+  template <typename Image>
+  class Hash
+  {
+  public:
+    unsigned int width;
```
yshekel, line 44: width of what? output? 
Consider having those const members

icicle/include/hash/hash.cuh
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    virtual cudaError_t pad_many(
+      Image* states,
+      unsigned int number_of_states,
+      unsigned int input_block_len,
+      unsigned int rate,
+      const device_context::DeviceContext& ctx) const
+    {
+      return cudaError_t::cudaSuccess;
```
yshekel, line 59: (1) derived class doesn't have to implement it?
(2) where is it padding to? states? if yes, do you assume it is large enough?

icicle/include/hash/hash.cuh
```diff
@@ -0,0 +1,257 @@
...
+    virtual cudaError_t squeeze_states(
+      const Image* states,
+      unsigned int number_of_states,
+      unsigned int rate,
+      unsigned int offset,
+      bool align,
+      Image* output,
+      const device_context::DeviceContext& ctx) const = 0;
+
+    virtual cudaError_t run_permutation_kernel(
```
yshekel, line 71: (1) can it be inplace? do you assume that derived class supports both in-place and out-of-place? also for squeeze_states.
(2) what is aligned? to what?

icicle/include/hash/hash.cuh
```diff
@@ -0,0 +1,257 @@
...
+    };
+    return config;
+  }
+
+  template <typename H, typename PreImage, typename Image>
+  class SpongeHasher
+  {
+  public:
+    unsigned int get_width() const {
+      return static_cast<const H*>(this)->width;
```
yshekel, line 114: implies that H must derive from SpongeHasher?

icicle/include/hash/hash.cuh
ChickenLover: width of hash state, yeah, it needs a lot of comments explaining this stuff:)

icicle/include/hash/hash.cuh
ChickenLover: 1) yes
2) yes and yes

icicle/include/hash/hash.cuh
```diff
@@ -0,0 +1,257 @@
...
+
+    cudaError_t absorb_many(
+      const PreImage* inputs,
+      Image* states,
+      unsigned int number_of_states,
+      unsigned int input_block_len,
+      const SpongeConfig& cfg) const
+    {
+      CHK_INIT_IF_RETURN();
+      unsigned int width = get_width();
```
yshekel, line 141: not related but generally I suggest adding const wherever you don't expect variables to change so that the intention is clear and also if you modify by mistake, it doesn't compile. Like in rust.

icicle/include/hash/hash.cuh
```diff
@@ -0,0 +1,257 @@
...
+      unsigned int preimage_max_length = get_preimage_max_length();
+
+      if (cfg.input_rate * sizeof(PreImage) > preimage_max_length * sizeof(Image))
```
yshekel, line 144: (1) max_length is mix #elements? assuming posseidon.
(2) for keccak what is it? and what the Image type would be?

icicle/include/hash/hash.cuh
ChickenLover: 1) yes it can. All of our hash functions support in-place permutation and I don't know of a hash that won't support it. For squeeze_states - no, it can't be inplace as you pointed in the kernel comment. Already changed it locally
2) Align refers to domain separation. In poseidon1 for example you need to put a domain tag as the first element of your permutation state. Align boolean is used in a poseidon1 kernel and is ignored for all the other hashes

icicle/include/hash/hash.cuh
```diff
@@ -0,0 +1,257 @@
...
+      CHK_IF_RETURN(padding_error);
+
+      cudaError_t permutation_error =
+        static_cast<const H*>(this)->run_permutation_kernel(states, states, number_of_states, cfg.aligned, cfg.ctx);
+      CHK_IF_RETURN(permutation_error);
+
+      if (!cfg.is_async) CHK_IF_RETURN(cudaStreamSynchronize(cfg.ctx.stream));
+
+      return CHK_LAST();
+    }
```
yshekel, line 168: it is one absorb stage right?

icicle/include/hash/hash.cuh
```diff
@@ -0,0 +1,257 @@
...
+      unsigned int number_of_states,
+      unsigned int rate,
+      unsigned int offset,
+      bool align,
+      Image* output,
+      const device_context::DeviceContext& ctx) const = 0;
+
+    virtual cudaError_t run_permutation_kernel(
+      const Image* states, Image* output, unsigned int number_of_states, bool aligned, const device_context::DeviceContext& ctx) const = 0;
+  };
```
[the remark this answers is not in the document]
ChickenLover: what about it?

icicle/include/hash/hash.cuh
```diff
@@ -0,0 +1,257 @@
...
+
+  struct SpongeConfig {
```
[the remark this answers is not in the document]
ChickenLover: sure

icicle/include/hash/hash.cuh
ChickenLover: yes. This is the worst place from design standpoint:) I tried to replicate trait requirement from Rust. Initially I've done it in pur c++ style but it was worse

icicle/include/hash/hash.cuh
ChickenLover: 1) max length refers to domain separation. For poseidon2 and keccak it is set to be equal to width
2) Image type is u64

icicle/include/hash/hash.cuh
ChickenLover: yes, you can run it multiple times to absorb more inputs

icicle/include/merkle-tree/merkle.cuh
```diff
@@ -0,0 +1,94 @@
...
+
+#include "gpu-utils/device_context.cuh"
+#include "gpu-utils/error_handler.cuh"
+#include "utils/utils.h"
+#include "hash/hash.cuh"
+
+#include <iostream>
+#include <math.h>
+
+using namespace hash;
```
yshekel, line 13: I would avoid it since it makes the code harder to read and also for every file that includes it. Up to you. I generally prefer avoiding 'using namespace' especially in headers.

icicle/include/merkle-tree/merkle.cuh
```diff
@@ -0,0 +1,94 @@
...
+ */
+namespace merkle_tree {
+  static constexpr size_t GIGA = 1024 * 1024 * 1024;
+
+  /// Bytes per stream
+  static constexpr size_t STREAM_CHUNK_SIZE = 1024 * 1024 * 1024;
+
+  /// Flattens the tree digests and sum them up to get
+  /// the memory needed to contain all the digests
+  static size_t get_digests_len(uint32_t height, uint32_t arity)
```
yshekel, line 28: can add constexpr here and have it computed at compile time if inputs are known at compile time. If not then it is computed at runtime.

icicle/include/merkle-tree/merkle.cuh
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  /**
+   * @struct TreeBuilderConfig
+   * Struct that encodes various Tree builder parameters.
+   */
+  struct TreeBuilderConfig {
+    device_context::DeviceContext ctx; /**< Details related to the device such as its id and stream id. */
+    int arity;
+    int keep_rows; /**< How many rows of the Merkle tree rows should be written to output. '0' means all of them */
```
yshekel, line 47: by rows you mean root is a row, its children are next row and so on?

icicle/include/merkle-tree/merkle.cuh
```diff
@@ -0,0 +1,94 @@
...
+    };
+    return config;
+  }
+
+  /**
+   * Builds the Merkle tree
+   *
+   * @param leaves a pointer to the leaves layer. May be allocated on device or on host, regulated by the config
+   * Expected to have arity ^ (height - 1) elements
+   * @param digests a pointer to the digests storage. May only be allocated on the host
```
yshekel, line 73: why only allocate on host?

icicle/include/hash/hash.cuh
```diff
@@ -0,0 +1,257 @@
...
+      0,     // offset
+      false, // recursive_squeeze
+      false, // recursive_squeeze
+      false, // is_async
+    };
+    return config;
+  }
+
+  template <typename H, typename PreImage, typename Image>
+  class SpongeHasher
```
yshekel, line 110: please explain why Hash and SpongeHasher are different classes? Seems like hash is not suitable to non sponge hashes? or is it?

icicle/include/merkle-tree/merkle.cuh
ChickenLover: yes

icicle/src/poseidon/extern.cu
```diff
@@ -2,58 +2,91 @@
...
-   * @return `cudaSuccess` if the execution was successful and an error code otherwise.
-   */
-  extern "C" cudaError_t CONCAT_EXPAND(FIELD, poseidon_hash_cuda)(
-    scalar_t* input,
-    scalar_t* output,
-    int number_of_states,
-    int arity,
-    const PoseidonConstants<scalar_t>& constants,
-    PoseidonConfig& config)
+  template class Poseidon<scalar_t>;
```
yshekel, line 12: basically you expose all of those and then in FFI you reconstruct the hash structure and call those functions?

icicle/include/merkle-tree/merkle.cuh
ChickenLover: just legacy. Filecoin trees were tens of GBs. Will update

icicle/include/hash/hash.cuh
ChickenLover: yes, I will merge them into a single class

icicle/src/poseidon/extern.cu
ChickenLover: yes, ffi just wraps the methods of the class

yshekel (commented):
Great work!

I wrote some comments/questions and have a few more:
(1) I did not see the merkle trees for poseidon/poseidon2 in rust. Did I just miss it?
(2) I am not sure about where the tree builders are in terms of library structre. I mean is poseidon/poseidon2 in the field crate and keccak in hashes? 
(3) (I may have asked this already but) Please clarify the FFI design and specifically how do you see it for V3? I think that basically the ffi functions (under api) are what is exposed to ffi and also what we expect backend to implement. On top of that we will wrap and expose nicer api in all langauges. Does that make sense to you and this design?

ChickenLover:
@yshekel 
1) There are no rust bindings for now
2) Tree builder is just a template, so instantiations will be both in hash and in field libraries, depending on which hash it is
3) Well, if we go with the current FFI design, then we will need to reimplement the tree builder for every backend. For hash functions - yes, that will do fine, but I am not sure what to do with the tree builder in that case

yshekel:
> @yshekel
> 
> 1. There are no rust bindings for now
> 2. Tree builder is just a template, so instantiations will be both in hash and in field libraries, depending on which hash it is
> 3. Well, if we go with the current FFI design, then we will need to reimplement the tree builder for every backend. For hash functions - yes, that will do fine, but I am not sure what to do with the tree builder in that case

If we can implement a generic tree builder (that is backend agnostic) then we can have the backend APIs implement only the hashes and then wrap it with the generic tree builder that calls the hashing APIs and builds the tree (and expose that to ffi in addition to hashes). I am not sure it's possible but maybe it is.
Otherwise we would have to define the tree builder a backend API and have the backend implement it, per hash. Internally it can reuse templates etc. to avoid fully reimplementing like you did basically.

icicle/include/api/babybear.h
```diff
@@ -32,47 +32,49 @@ extern "C" cudaError_t babybear_create_poseidon2_cuda(
...
 );
 
-extern "C" cudaError_t babybear_poseidon2_permute_many_cuda(
+extern "C" cudaError_t babybear_poseidon2_absorb_many_cuda(
+  const poseidon2::Poseidon2<babybear::scalar_t>* poseidon,
+  const babybear::scalar_t* inputs,
+  babybear::scalar_t* states,
+  unsigned int number_of_states,
+  unsigned int input_block_len,
+  hash::SpongeConfig& cfg);
```
yshekel, line 49: probably better to make it const reference assuming you are not going to modify it inside

icicle/include/api/babybear.h
```diff
@@ -32,47 +32,49 @@ extern "C" cudaError_t babybear_create_poseidon2_cuda(
...
   poseidon2::MdsType mds_type,
   poseidon2::DiffusionStrategy diffusion,
   device_context::DeviceContext& ctx
 );
 
-extern "C" cudaError_t babybear_poseidon2_permute_many_cuda(
+extern "C" cudaError_t babybear_poseidon2_absorb_many_cuda(
+  const poseidon2::Poseidon2<babybear::scalar_t>* poseidon,
+  const babybear::scalar_t* inputs,
+  babybear::scalar_t* states,
```
yshekel, line 46: what if states was internal to poseidon? I mean the first param. Or do you prefer to decouple them? why?

icicle/include/api/babybear.h
```diff
@@ -32,47 +32,49 @@ extern "C" cudaError_t babybear_create_poseidon2_cuda(
...
-  device_context::DeviceContext& ctx);
+extern "C" cudaError_t
+  babybear_poseidon2_delete_cuda(poseidon2::Poseidon2<babybear::scalar_t>* poseidon, device_context::DeviceContext& ctx);
 
-extern "C" cudaError_t babybear_build_poseidon_merkle_tree(
+extern "C" cudaError_t babybear_build_poseidon2_merkle_tree(
   const babybear::scalar_t* leaves,
   babybear::scalar_t* digests,
   unsigned int height,
-  unsigned int arity,
```
yshekel: why did you remove arity? is it in the config now?

icicle/include/hash/keccak/keccak.cuh
```diff
@@ -33,24 +37,39 @@ namespace keccak {
...
+      const device_context::DeviceContext& ctx) const override;
+
+    cudaError_t run_permutation_kernel(
+      const uint64_t* states,
+      uint64_t* output,
+      unsigned int number_of_states,
+      bool aligned,
+      const device_context::DeviceContext& ctx) const override;
+
+    Keccak(unsigned int rate) : SpongeHasher<uint8_t, uint64_t>(25, 25, rate, 0){};
```
yshekel, line 70: 25 is always true for keccak? maybe add comment or make a constant with clear name

icicle/include/api/babybear.h
ChickenLover: the idea here is to give a flexible API for user to do any kind of memory operations. If you look into our tree builder for example, it features a lot of memory handling tricks, like using just 2 vectors to build an entire tree. Also, I don't really like adding states inside the class as it will require managing the release of that memory through ffi

icicle/include/api/babybear.h
ChickenLover: exactly. Actually, both variants are fine I guess, but given the fact that most of the trees are binary - adding it to config seems better

icicle/include/hash/hash.cuh
```diff
@@ -0,0 +1,209 @@
...
+   */
+  template <typename S>
+  __global__ void generic_squeeze_states_kernel(
+    const S* states, unsigned int number_of_states, unsigned int width, unsigned int rate, unsigned int offset, S* out)
+  {
+    int idx = (blockIdx.x * blockDim.x) + threadIdx.x;
+    if (idx >= number_of_states) { return; }
+
+    for (int i = 0; i < rate; i++) {
+      out[idx * rate + i] = states[idx * width + offset];
```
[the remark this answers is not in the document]
ChickenLover: nope, I removed that

icicle/include/hash/keccak/keccak.cuh
ChickenLover: It's doesn't have to be 25, but everyone only use 25. Yeah, adding a constant is nice here. If at any point someone will want to use non-25 variant - we can just inherit this class and override the constructor

jeremyfelder (commented):
Very few small comments

icicle/include/hash/hash.cuh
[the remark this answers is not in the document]
jeremyfelder: Lets remove the comment in the params comments that says we support in place

icicle/src/hash/keccak/test_keccak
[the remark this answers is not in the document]
ChickenLover: yes. good catch

icicle/src/fields/CMakeLists.txt
```diff
@@ -25,8 +25,7 @@ set(POLYNOMIAL_SOURCE_FILES
 # TODO: impl poseidon for small fields. note that it needs to be defined over the extension field!
 if (DEFINED CURVE)
   list(APPEND FIELD_SOURCE ${SRC}/poseidon/extern.cu)
-  list(APPEND FIELD_SOURCE ${SRC}/poseidon/poseidon.cu)
-  list(APPEND FIELD_SOURCE ${SRC}/poseidon/tree/merkle.cu)
+  # list(APPEND FIELD_SOURCE ${SRC}/poseidon/tree/merkle.cu)
```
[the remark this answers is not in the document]
ChickenLover: no, this will be removed

icicle/include/hash/keccak/keccak.cuh
ChickenLover: @jeremyfelder yes. we would, but I doubt we will need it

icicle/include/api/babybear.h
ChickenLover: I'll leave it like that for now, because in all of our primitives we use non-const configs. We can change them all at once in the future

wrappers/rust/icicle-core/src/poseidon/tests.rs
```diff
@@ -1,105 +1,48 @@
...
 where
-    <F as FieldImpl>::Config: Poseidon<F>,
+    <F as FieldImpl>::Config: PoseidonImpl<F>,
 {
-    for arity in [2, 4] {
-        let constants = init_poseidon::<F>(arity as u32);
+    for arity in [2, 4, 8, 11] {
+        let constants = init_poseidon::<F>(arity);
 
         _check_poseidon_hash_many(constants);
```
jeremyfelder, line 46: Nit: can we change the var name to poseidon
```suggestion
        let poseidon = init_poseidon::<F>(arity);

        _check_poseidon_hash_many(poseidon);
```

wrappers/rust/icicle-curves/icicle-bn254/src/lib.rs
```diff
@@ -6,7 +6,7 @@ pub mod ntt;
 pub mod polynomials;
 pub mod poseidon;
 pub mod poseidon2;
-pub mod tree;
+// pub mod tree;
```
jeremyfelder, line 9: Remove?

jeremyfelder (approved):

LeonHibnik (approved):

PR #528, opened 2024-05-27
bit reverse

nonam3e:
This PR adds bit reverse operation support to icicle


icicle/src/vec_ops/vec_ops.cu
```diff
@@ -164,4 +189,66 @@ namespace vec_ops {
 
     return CHK_LAST();
   }
+
+  template <typename E>
```
vhnatyk, line 193: existing template for vec_ops cannot be reused here?

nonam3e: I deleted the inplace template, it's merged with `bit_reverse`

icicle/src/vec_ops/vec_ops.cu
```diff
@@ -54,6 +54,31 @@ namespace vec_ops {
...
+    {
+      int tid = blockIdx.x * blockDim.x + threadIdx.x;
+      // Handling arbitrary vector size
+      if (tid < n) {
+        int reversed_index = __brev(tid) >> shift;
+        output[reversed_index] = input[tid];
+      }
+    }
+    template <typename E>
+    __global__ void bit_reverse_inplace_kernel(E* input, unsigned n, unsigned shift)
```
yshekel, line 69: you could merge those two kernels like the implementation in ntt.cu is doing and simplify the bit_reverse() since it would not have to choose which one to call

icicle/src/vec_ops/vec_ops.cu
```diff
@@ -164,4 +189,43 @@ namespace vec_ops {
...
+    E* d_output;
+    if (cfg.is_output_on_device) {
+      d_output = output;
+    } else {
+      // allocate output on gpu
+      CHK_IF_RETURN(cudaMallocAsync(&d_output, sizeof(E) * size, cfg.ctx.stream));
+    }
+    unsigned shift = __builtin_clz(size) + 1;
+    unsigned num_blocks = (size + MAX_THREADS_PER_BLOCK - 1) / MAX_THREADS_PER_BLOCK;
+    if (input == output) {
```
yshekel, line 206: I think this is interesting only if the input/output are on device.
if your input/output is not on device then anyway you can copy only the input, reverse in place and copy back to the host output. If input and output both on host it doesn't really matter if they are the same or not.

icicle/include/vec_ops/vec_ops.cuh
```diff
@@ -113,6 +113,27 @@ namespace vec_ops {
     device_context::DeviceContext& ctx,
     bool on_device,
     bool is_async);
+
+  struct BitReverseConfig {
```
yshekel, line 117: (1) looks like the vecOpsConfig so consider reusing it instead of adding a new struct.
(2) for inplace, input==output but the config may tell you that one is on device but the other is not. How do handle that?

nonam3e: (2) I use only `is_output_on_device` for inplace because `input` isn't used in this scenario.

icicle/src/vec_ops/vec_ops.cu
```diff
@@ -164,4 +189,39 @@ namespace vec_ops {
 
     return CHK_LAST();
   }
-} // namespace vec_ops
\ No newline at end of file
+
+  template <typename E>
+  cudaError_t bit_reverse(const E* input, unsigned size, BitReverseConfig& cfg, E* output)
```
yshekel, line 194: Note that you have a limit of 2^32-1 elements. Maybe use u64

yshekel (approved):

ChickenLover (approved):

PR #529, opened 2024-05-28
precompute bug fix

HadarIngonyama:
This PR fixes 2 things:

1. Removes the assertion regarding the precompute factor needing to be a power of 2. There is no such requirement and it works just fine for other values too.
2. Fixes the average bucket size for the large buckets threshold - it depends on the precompute factor.


ChickenLover (approved):

yshekel (approved):

nonam3e (approved):

jeremyfelder:
Merging with the failed spell check since it will be fixed in https://github.com/ingonyama-zk/icicle/pull/528

PR #531, opened 2024-06-04
Golang/poseidon wrappers

nonam3e:
This PR updates poseidon golang bindings


wrappers/golang/internal/generator/poseidon/templates/poseidon_test.go.tmpl
```diff
@@ -1,59 +1 @@
 package tests
```
jeremyfelder, line 1: We can remove this file if there are no tests in it. Also will need to remove the generation of it from the poseidon/generate.go

LeonHibnik (approved):

wrappers/golang/curves/bls12377/poseidon/poseidon.go
```diff
@@ -3,55 +3,110 @@ package poseidon
...
+	cSparseMatrices := (*C.scalar_t)(sparseMatrices.AsUnsafePointer())
+	cDomainTag := (*C.scalar_t)(unsafe.Pointer(&domainTag))
+	cCtx := (*C.DeviceContext)(unsafe.Pointer(ctx))
+	__ret := C.bls12_377_poseidon_create_cuda(&poseidon, cArity, cAlpha, cFullRoundsHalf, cPartialRounds, cScalars, cMdsMatrix, cNonSparseMatrix, cSparseMatrices, cDomainTag, cCtx)
+	err := core.FromCudaError((cr.CudaError)(__ret))
+	if err.IcicleErrorCode != core.IcicleSuccess {
+		return nil, err
+	}
+	p := Poseidon{handle: poseidon, width: arity + 1}
+	runtime.SetFinalizer(ctx, p.Delete)
```
jeremyfelder, line 39: Why is the Finalizer being set on the `ctx`? Its possible that the context may outlive the need of Poseidon and be used with other primitives making preventing Poseidon from being deleted

ChickenLover: I guess a simple solution here would be to just store device id inside Poseidon struct

icicle/include/api/babybear.h
```diff
@@ -94,12 +94,6 @@ extern "C" cudaError_t babybear_transpose_matrix_cuda(
   bool on_device,
   bool is_async);
 
-extern "C" cudaError_t babybear_bit_reverse_cuda(
```
ChickenLover, line 97: This was not intended:)
Please add bit_reverse into templates

wrappers/golang/curves/bls12377/poseidon/poseidon.go
jeremyfelder: Yes, this is what I discussed with @nonam3e:

"Maybe we need to add a DeviceId to the Poseidon struct in Go so that we can call SetFinalizer(&p, deletePoseidon) where deletePoseidon is a wrapper function around p.Delete that creates a new ctx object and assigns the correct DeviceId based on p.DeviceId"

specifically the wrapper function is required because the second argument to `SetFinalizer` must be a function that takes a single parameter of the type of the object in the first argument to `SetFinalizer`

wrappers/golang/core/sponge.go
```diff
@@ -0,0 +1,105 @@
...
+	/// Aligned format: [0, A, B, 0, C, D, ...] (as you might get by using loop_state)
+	/// not aligned format: [A, B, 0, C, D, 0, ...] (as you might get from cudaMemcpy2D)
+	RecursiveSqueeze bool
+
+	/// If true, hash results will also be copied in the input pointer in aligned format
+	Aligned bool
+
+	/// Whether to run the MSM asynchronously. If set to `true`, the MSM function will be non-blocking
+	/// and you'd need to synchronize it explicitly by running `cudaStreamSynchronize` or `cudaDeviceSynchronize`.
+	/// If set to `false`, the MSM function will block the current CPU thread.
```
jeremyfelder, line 30: Change mentions of MSM to SpongeHash operations or something more appropriate

jeremyfelder (approved):

PR #533, opened 2024-06-05
ntt cpu

ShanieWinitz:


icicle_v3/tests/test_field_api.cpp
```diff
@@ -204,13 +206,47 @@ TYPED_TEST(FieldApiTest, montgomeryConversion)
...
+  srand(time(0));
+  const int logn = rand() % 13 + 3;
   const int N = 1 << logn;
-  auto scalars = std::make_unique<TypeParam[]>(N);
-  FieldApiTest<TypeParam>::random_samples(scalars.get(), N);
 
-  auto out_main = std::make_unique<TypeParam[]>(N);
-  auto out_ref = std::make_unique<TypeParam[]>(N);
+  // // Randomize config
+  // TODO - iterate over different configs
```
yshekel, line 214: when are you going to do it? I think it's better than random since we know what we check. Otherwise CI can pass sometimes and fail in others, based on the randomized case.

icicle_v3/tests/test_field_api.cpp
```diff
@@ -204,13 +206,47 @@ TYPED_TEST(FieldApiTest, montgomeryConversion)
...
+  // const NTTDir dir = static_cast<NTTDir>(rand() % 2); // 0: forward, 1: inverse
+
+  // Set fixed config values for debugging
+  const int batch_size = 2;
+  const bool columns_batch = false;
+  const Ordering ordering = Ordering::kNM;
+  const NTTDir dir = NTTDir::kForward;
+
+  // Print config
+  ICICLE_LOG_DEBUG << "logn: " << logn;
```
yshekel, line 232: when done, please remove those logs.

icicle_v3/tests/test_field_api.cpp
```diff
@@ -224,31 +260,34 @@ TYPED_TEST(FieldApiTest, ntt)
...
 
-    ICICLE_CHECK(ntt_init_domain(scalar_t::omega(logn), init_domain_config));
-
     auto config = default_ntt_config<scalar_t>();
+    // config.stream = stream;
+    // config.coset_gen = coset_gen; // TODO - Implement. default: S::one()
+    config.batch_size = batch_size;       // default: 1
+    config.columns_batch = columns_batch; // default: false
+    config.ordering = ordering;           // default: kNN
+    config.are_inputs_on_device = true;   // TODO, ask yuval why set to true?
```
yshekel, line 269: for CPU it doesn't matter but for devices with own memory, like GPU/ZPU/FPGA, we need to know where the data resides.

icicle_v3/backend/cpu/src/field/cpu_ntt.cpp
```diff
@@ -1,33 +1,44 @@
...
+REGISTER_NTT_INIT_DOMAIN_BACKEND("CPU_REF", (cpu_ntt_init_domain));
+
+REGISTER_NTT_RELEASE_DOMAIN_BACKEND("CPU", cpu_ntt_release_domain);
+REGISTER_NTT_RELEASE_DOMAIN_BACKEND("CPU_REF", cpu_ntt_release_domain);
+
+REGISTER_NTT_BACKEND("CPU", (cpu_ntt<scalar_t, scalar_t>));
+REGISTER_NTT_BACKEND("CPU_REF", (cpu_ntt_ref));
+
+#ifdef EXT_FIELD
+REGISTER_NTT_EXT_FIELD_BACKEND("CPU", (cpu_ntt<scalar_t, extension_t>));
```
yshekel, line 43: probably this case also needs to be registered for CPU_REF.

icicle_v3/backend/cpu/src/cpu_device_api.cpp
```diff
@@ -92,3 +92,9 @@ class CpuDeviceAPI : public DeviceAPI
 };
 
 REGISTER_DEVICE_API("CPU", CpuDeviceAPI);
+
+class CpuRefDevice : public CpuDeviceAPI
```
yshekel, line 96: I think that if you rebase you can simply do 
REGISTER_DEVICE_API("CPU_REF", CpuDeviceAPI);
without having the class again. I remember that this was solved at some point. Not sure about that but if possible remove CpuRefDevice and use CpuDeviceAPI instead.

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
+#include "icicle/ntt.h"
```
yshekel, line 1: I know that I named the file ntt.template but maybe cpu_ntt.h is better to be more consistent with the code base which is full of templates but not named with 'template' in the file name.

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+#include <cmath>
+#include <cstdint>
+
+using namespace field_config;
+using namespace icicle;
+namespace ntt_template {
+  template <typename S>
+  class CpuNttDomain
+  {
+    // TODO - coset, mixed-radix NTT
```
yshekel, line 21: please put your name on TODOs (e.g. TODO Shanie) so that it is clear who is going to do it.

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+
+    template <typename U, typename E>
+    friend eIcicleError
+    cpu_ntt_ref(const Device& device, const E* input, int size, NTTDir dir, NTTConfig<S>& config, E* output);
+
+    template <typename U, typename E>
+    friend eIcicleError
+    cpu_ntt(const Device& device, const E* input, int size, NTTDir dir, NTTConfig<S>& config, E* output);
+
+    S* get_twiddles() { return twiddles; }
```
yshekel, line 45: make it 'const S*' since it should be read only right?

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+
+  template <typename S>
+  static CpuNttDomain<S>* s_ntt_domain = nullptr;
+
+  template <typename S = scalar_t>
+  eIcicleError generate_twiddles(const S& primitive_root, S* twiddles, int n)
+  {
+    S omega = primitive_root;
+    twiddles[0] = S::one();
+    for (int i = 1; i <= n; i++) {
```
yshekel, line 56: note that you compute n+1 values, meaning first and last values are one.
This is worth a comment.

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+    S* get_twiddles() { return twiddles; }
+  };
+
+  template <typename S>
+  static CpuNttDomain<S>* s_ntt_domain = nullptr;
+
+  template <typename S = scalar_t>
+  eIcicleError generate_twiddles(const S& primitive_root, S* twiddles, int n)
+  {
+    S omega = primitive_root;
```
yshekel, line 54: (1) check that twiddles is not null and maybe that n is a power of two?
(2) in the CUDA implementation I think that they check more than that to make sure this really is a primitive root. I think it would be better to have it here too.

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+    return eIcicleError::SUCCESS;
+  }
+
+  template <typename S = scalar_t>
+  eIcicleError cpu_ntt_init_domain(const Device& device, const S& primitive_root, const NTTInitDomainConfig& config)
+  {
+    // (1) check if need to refresh domain. TODO - reusing domain for different ntt sizes if possible
+    if (s_ntt_domain<S> != nullptr) { return eIcicleError::SUCCESS; }
+    // (2) build the domain
+    delete s_ntt_domain<S>;
```
yshekel, line 68: redundant delete. If ntt_domain is nullptr which is necessary for this code to execute

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+
+    S* get_twiddles() { return twiddles; }
+  };
+
+  template <typename S>
+  static CpuNttDomain<S>* s_ntt_domain = nullptr;
+
+  template <typename S = scalar_t>
+  eIcicleError generate_twiddles(const S& primitive_root, S* twiddles, int n)
+  {
```
yshekel, line 53: I would inline this code cpu_ntt_init_domain() since it's is implicitly assumed that the arguments are correct (e.g. n is power of two, the root is indeed this root etc) and only used by cpu_ntt_init_domain

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+  template <typename S = scalar_t>
+  eIcicleError cpu_ntt_init_domain(const Device& device, const S& primitive_root, const NTTInitDomainConfig& config)
+  {
+    // (1) check if need to refresh domain. TODO - reusing domain for different ntt sizes if possible
+    if (s_ntt_domain<S> != nullptr) { return eIcicleError::SUCCESS; }
+    // (2) build the domain
+    delete s_ntt_domain<S>;
+    s_ntt_domain<S> = new CpuNttDomain<S>();
+
+    bool found_logn = false;
```
yshekel, line 71: please add comments to explain what is being done here

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+    if (omega != S::one()) {
+      ICICLE_LOG_ERROR << "Primitive root provided to the InitDomain function is not in the subgroup";
+      return eIcicleError::INVALID_ARGUMENT;
+    }
+
+    // calculate twiddles
+    // Note: radix-2 INTT needs ONE in last element (in addition to first element), therefore have n+1 elements
+
+    s_ntt_domain<S>->twiddles = (S*)malloc((s_ntt_domain<S>->max_size + 1) * sizeof(S));
+    generate_twiddles(primitive_root, s_ntt_domain<S>->twiddles, s_ntt_domain<S>->max_size);
```
yshekel, line 93: return the value returned by generate_twiddles() since it is returning an eIcicleError so it could silently fail otherwise.

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+    free(s_ntt_domain<S>->twiddles);
+    delete s_ntt_domain<S>;
+    s_ntt_domain<S> = nullptr;
+    return eIcicleError::SUCCESS;
+  }
+
+  template <typename E = scalar_t>
+  eIcicleError bit_reverse(int size, int logn, E* output, int batch_size)
+  {
+    ICICLE_LOG_DEBUG << "BIT REVERSE";
```
yshekel, line 110: remove prints when merging

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+    int total_size = size * batch_size;
```
yshekel, line 111: note that total_size can be larger than 32b so you should revisit where you use int and consider u64.

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+  {
+    // release the memory
+    free(s_ntt_domain<S>->twiddles);
+    delete s_ntt_domain<S>;
+    s_ntt_domain<S> = nullptr;
+    return eIcicleError::SUCCESS;
+  }
+
+  template <typename E = scalar_t>
+  eIcicleError bit_reverse(int size, int logn, E* output, int batch_size)
```
yshekel, line 108: I would only pass logn and compute n=1<<logn to avoid a mismatch.

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+    s_ntt_domain<S>->max_size = (int)pow(2, s_ntt_domain<S>->max_log_size);
+    if (omega != S::one()) {
+      ICICLE_LOG_ERROR << "Primitive root provided to the InitDomain function is not in the subgroup";
+      return eIcicleError::INVALID_ARGUMENT;
+    }
+
+    // calculate twiddles
+    // Note: radix-2 INTT needs ONE in last element (in addition to first element), therefore have n+1 elements
+
+    s_ntt_domain<S>->twiddles = (S*)malloc((s_ntt_domain<S>->max_size + 1) * sizeof(S));
```
yshekel, line 92: I would avoid malloc/free and also new/delete and use smart pointers instead. In that case probably twiddles should be a unique_ptr.
Generally better avoid those since it avoids all the double frees and but of this kind and modern c++ has better solutions.

Even the Domain instance itself could be a value and not a pointer so that you don't need to call new and delete on it, if you want.

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+
+    template <typename U, typename E>
+    friend eIcicleError
+    cpu_ntt(const Device& device, const E* input, int size, NTTDir dir, NTTConfig<S>& config, E* output);
+
+    S* get_twiddles() { return twiddles; }
+  };
+
+  template <typename S>
+  static CpuNttDomain<S>* s_ntt_domain = nullptr;
```
yshekel, line 49: note that every cpp file that includes this header will have static s_ntt_domain of its own.
Maybe you need to use the extern keyword here and instantiate in in cpu_ntt.cpp. I am not sure this is the best way but as it is now, it can be a problem, where you allocate a domain but then maybe ecntt doesn't see it.

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+    int total_size = size * config.batch_size;
+    E* temp_elements = new E[total_size];
+
+    // ICICLE_LOG_DEBUG << "INITIAL INPUT";
+    // for (int i = 0; i < total_size; ++i) {
+    //     ICICLE_LOG_DEBUG << "input[" << i << "]: " << input[i];
+    // }
+
+    if (config.columns_batch) {
+      // Distribute input into columns if columns_batch is set
```
yshekel, line 139: transpose?

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+
+  template <typename S = scalar_t, typename E = scalar_t>
+  eIcicleError cpu_ntt_ref(const Device& device, const E* input, int size, NTTDir dir, NTTConfig<S>& config, E* output)
+  {
+    if (size & (size - 1)) { return eIcicleError::INVALID_ARGUMENT; }
+
+    int total_size = size * config.batch_size;
+    E* temp_elements = new E[total_size];
+
+    // ICICLE_LOG_DEBUG << "INITIAL INPUT";
```
yshekel, line 133: remove commented out code

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+    // ICICLE_LOG_DEBUG << "AFTER BIT REVERSE";
+    // for (int i = 0; i < total_size; ++i) {
+    //     ICICLE_LOG_DEBUG << "temp_elements[" << i << "]: " << temp_elements[i];
+    // }
+
+    S* twiddles = s_ntt_domain<S>->get_twiddles();
+
+    // NTT/INTT
+    int ntt_step = 0;
+    if (dit) {
```
yshekel, line 183: I would move dit and dif into functions to make the code easier to read

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+    if (dir == NTTDir::kInverse) {
+      // Normalize results
+      S inv_size = S::inv_log_size(logn);
+      for (int i = 0; i < total_size; ++i) {
+        temp_elements[i] = temp_elements[i] * inv_size;
+      }
+    }
+
+    if (config.columns_batch) {
+      // Distribute output into columns if columns_batch is set
```
yshekel, line 241: since you have it twice I would make it a function named transpose, assuming that's what it is doing

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+
+    if (config.columns_batch) {
+      // Distribute input into columns if columns_batch is set
+      for (int col = 0; col < config.batch_size; ++col) {
+        for (int row = 0; row < size; ++row) {
+          temp_elements[col * size + row] = input[row * config.batch_size + col];
+        }
+      }
+    } else {
+      std::copy(input, input + total_size, temp_elements);
```
yshekel, line 146: you can make temp_elements point to input rather than copy which is expensive

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+    return eIcicleError::SUCCESS;
+  }
+
+  template <typename S = scalar_t, typename E = scalar_t>
+  eIcicleError cpu_ntt_ref(const Device& device, const E* input, int size, NTTDir dir, NTTConfig<S>& config, E* output)
+  {
+    if (size & (size - 1)) { return eIcicleError::INVALID_ARGUMENT; }
+
+    int total_size = size * config.batch_size;
+    E* temp_elements = new E[total_size];
```
yshekel, line 131: prefer make_unique<S[]](size) over new.
It is automatically released so you don't leak even if you hit an exception or error that makes you return early.

icicle_v3/backend/cpu/include/ntt_template.h
ShanieWinitz: From now on I'll do that. Anyway, I'm removing this line, coset implementetion will be added in the next commit, and mixed radix is not for now (right?)

icicle_v3/backend/cpu/include/ntt_template.h
```diff
@@ -0,0 +1,264 @@
...
+
+#include <vector>
+#include <algorithm>
+#include <iostream>
+#include <cmath>
+#include <cstdint>
+
+using namespace field_config;
+using namespace icicle;
+namespace ntt_template {
```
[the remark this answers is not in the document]
ShanieWinitz: Maybe "cpu_ntt_h" or something else? "cpu_ntt" creates errors in cpu_ntt.cpp

icicle_v3/backend/cpu/include/ntt_template.h
ShanieWinitz: Right, fixed

icicle_v3/backend/cpu/include/ntt_template.h
ShanieWinitz: Fixed

icicle_v3/backend/cpu/include/ntt_template.h
ShanieWinitz: Fixed

icicle_v3/tests/test_field_api.cpp
yshekel: update: since rust tests are covering many cases, this test can doesn't have to as well.

icicle_v3/backend/cpu/include/ntt_template.h
ShanieWinitz: fixed

icicle_v3/backend/cpu/include/ntt_template.h
ShanieWinitz: removed

icicle_v3/backend/cpu/include/ntt_template.h
ShanieWinitz: fixed

icicle_v3/backend/cpu/include/ntt_template.h
ShanieWinitz: fixed

icicle_v3/backend/cpu/include/ntt_template.h
ShanieWinitz: fixed

icicle_v3/backend/cpu/include/ntt_template.h
ShanieWinitz: yes, fixed

icicle_v3/backend/cpu/include/ntt_template.h
ShanieWinitz: fixed

icicle_v3/backend/cpu/include/ntt_template.h
ShanieWinitz: fixed

icicle_v3/backend/cpu/include/ntt_template.h
ShanieWinitz: fixed

icicle_v3/tests/test_field_api.cpp
ShanieWinitz: ok

icicle_v3/tests/test_field_api.cpp
ShanieWinitz: ok

icicle_v3/backend/cpu/include/ntt_template.h
ShanieWinitz: Changed back to copy to ensure that freeing temp_elements does not cause the input to be freed, preventing a potential double-free error. 
added TODO: Later, remove temp_elements and perform all calculations in-place (implement NTT for the case where columns_batch=true, in-place).

icicle_v3/backend/cpu/src/cpu_device_api.cpp
ShanieWinitz: It doesn't work, so I left it as it was. Do you want to look at it together?

icicle_v3/backend/cpu/src/field/cpu_ntt.cpp
```diff
@@ -1,33 +1,44 @@
...
-  ICICLE_LOG_ERROR << "cpu_ntt_release_domain() not implemented";
-  return eIcicleError::API_NOT_IMPLEMENTED;
+  auto err = ntt_template::cpu_ntt_ref<scalar_t>(device, input, size, dir, config, output);
+  return err;
 }
 
-REGISTER_NTT_RELEASE_DOMAIN_BACKEND("CPU", cpu_ntt_release_domain);
\ No newline at end of file
+REGISTER_NTT_INIT_DOMAIN_BACKEND("CPU", (cpu_ntt_init_domain));
+REGISTER_NTT_INIT_DOMAIN_BACKEND("CPU_REF", (cpu_ntt_init_domain));
```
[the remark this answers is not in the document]
ShanieWinitz: As we talked, I left it like this.

icicle_v3/backend/cpu/src/field/cpu_ntt.cpp
ShanieWinitz: Fixed

wrappers/rust_v3/icicle-core/src/ntt/tests.rs
```diff
@@ -167,6 +167,7 @@ where
 }
 
 // This test is interpolating a coset, given evaluations on rou, and compares main to ref device.
+//TODO SHANIE - mixed radix not implemented in icicle-cpu-ntt
```
yshekel, line 170: make sure to delete that

icicle_v3/tests/test_field_api.cpp
```diff
@@ -247,13 +249,35 @@ TYPED_TEST(FieldApiTest, bitReverse)
 #ifdef NTT_ENABLED
 TYPED_TEST(FieldApiTest, ntt)
 {
-  const int logn = 20;
+  Log::set_min_log_level(Log::Debug);
```
yshekel, line 252: please remove

icicle_v3/tests/test_field_api.cpp
```diff
@@ -247,13 +249,35 @@ TYPED_TEST(FieldApiTest, bitReverse)
...
+  const int logn = rand() % 10 + 3;
   const int N = 1 << logn;
-  auto scalars = std::make_unique<TypeParam[]>(N);
-  FieldApiTest<TypeParam>::random_samples(scalars.get(), N);
+  const int log_batch_size = rand() % 10 + 3;
+  const int batch_size = 1 << log_batch_size;
+  const Ordering ordering = static_cast<Ordering>(rand() % 4); // TODO SHANIE - mixed radix (KNM, KMN) not supported.
+                                                               // After implementation, update this line to % 6
+  bool columns_batch;
+  if (ordering == Ordering::kMN || ordering == Ordering::kNM || logn == 7 || logn < 4) {
```
yshekel, line 262: Let's have logn>7 always and remove this line.

icicle_v3/tests/test_field_api.cpp
```diff
@@ -247,13 +249,35 @@ TYPED_TEST(FieldApiTest, bitReverse)
...
+  Log::set_min_log_level(Log::Debug);
+  srand(time(0));
+  // Randomize config
+  const int logn = rand() % 10 + 3;
   const int N = 1 << logn;
-  auto scalars = std::make_unique<TypeParam[]>(N);
-  FieldApiTest<TypeParam>::random_samples(scalars.get(), N);
+  const int log_batch_size = rand() % 10 + 3;
+  const int batch_size = 1 << log_batch_size;
+  const Ordering ordering = static_cast<Ordering>(rand() % 4); // TODO SHANIE - mixed radix (KNM, KMN) not supported.
```
yshekel, line 259: let's keep it like that. You don't need to implemented mixed-radix alg

icicle_v3/tests/test_field_api.cpp
```diff
@@ -269,30 +293,32 @@ TYPED_TEST(FieldApiTest, ntt)
...
-    config.ordering = Ordering::kNN;
     config.stream = stream;
+    config.coset_gen = coset_gen;
+    config.batch_size = batch_size;       // default: 1
+    config.columns_batch = columns_batch; // default: false
+    config.ordering = ordering;           // default: kNN
     config.are_inputs_on_device = true;
     config.are_outputs_on_device = true;
+    config.is_async = true; // default: false
+    ICICLE_CHECK(ntt_init_domain(scalar_t::omega(17), init_domain_config));
```
yshekel, line 305: define the log_ntt_domain_size as a variable (like ITERS on top) and use that instead of hardcoded 17.

icicle_v3/backend/cpu/src/curve/cpu_ecntt.cpp
```diff
@@ -2,11 +2,17 @@
...
-#include "ntt.template"
+#include "cpu_ntt.h"
 
 #include "icicle/curves/curve_config.h"
 
-using namespace field_config;
+using namespace curve_config;
 using namespace icicle;
 
+template <typename S, typename E>
```
yshekel, line 12: you should also register ecntt. Otherwise some tests will fail in rust.

icicle_v3/backend/cpu/src/cpu_device_api.cpp
yshekel: not that important. Ok leave it like that then

icicle_v3/backend/cpu/include/cpu_ntt.h
```diff
@@ -0,0 +1,331 @@
...
+#include <iostream>
+#include <cmath>
+#include <cstdint>
+#include <memory>
+#include <mutex>
+
+using namespace field_config;
+using namespace icicle;
+
+namespace cpu_ntt_h {
```
yshekel, line 18: I personally would remove the _h since it's a little weird but you can leave it like that if you like it

icicle_v3/backend/cpu/include/cpu_ntt.h
```diff
@@ -0,0 +1,331 @@
...
+  eIcicleError
+  CpuNttDomain<S>::cpu_ntt_init_domain(const Device& device, const S& primitive_root, const NTTInitDomainConfig& config)
+  {
+    // (1) check if need to refresh domain. TODO SHANIE - reusing domain for different ntt sizes if possible
+    if (s_ntt_domain.twiddles != nullptr) { return eIcicleError::SUCCESS; }
+
+    // Lock the mutex to ensure thread safety during initialization
+    std::lock_guard<std::mutex> lock(s_ntt_domain.domain_mutex); // Added for thread safety
+
+    // Double-checked locking
```
yshekel, line 57: comment not clear

icicle_v3/backend/cpu/include/cpu_ntt.h
```diff
@@ -0,0 +1,331 @@
...
+        if (!found_logn) {
+          ++s_ntt_domain.max_log_size;
+          found_logn = omega == S::one();
+          if (found_logn) break;
+        }
+      }
+
+      s_ntt_domain.max_size = (int)pow(2, s_ntt_domain.max_log_size);
+      if (omega != S::one()) {
+        ICICLE_LOG_ERROR << "Primitive root provided to the InitDomain function is not in the subgroup";
```
yshekel, line 75: What do you mean it's not in the subgroup?
I would rephrase the error to "not a root-of-unity" or something like that.

icicle_v3/backend/cpu/include/cpu_ntt.h
```diff
@@ -0,0 +1,331 @@
...
+        return eIcicleError::INVALID_ARGUMENT;
+      }
+
+      // calculate twiddles
+      // Note: radix-2 INTT needs ONE in last element (in addition to first element), therefore have n+1 elements
+
+      std::unique_ptr<S[]> temp_twiddles(new S[s_ntt_domain.max_size + 1]);
```
yshekel, line 82: please add comment why you need the temp twiddles. It's because otherwise twiddles are not nullptr but still not computed right?

BTW you'd better use std::make_unique<S[]>(max_size+1) instead of new.

icicle_v3/backend/cpu/include/cpu_ntt.h
```diff
@@ -0,0 +1,331 @@
...
+      s_ntt_domain.twiddles = std::move(temp_twiddles); // Assign twiddles using unique_ptr
+    }
+    return eIcicleError::SUCCESS;
+  }
+
+  template <typename S>
+  eIcicleError CpuNttDomain<S>::cpu_ntt_release_domain(const Device& device)
+  {
+    // No need to explicitly free memory since unique_ptr will handle it automatically
+    s_ntt_domain.twiddles.reset(); // Set twiddles to nullptr
```
yshekel, line 98: make sure to use the mutex here too. I think in CUDA it's missing but I think it is important

icicle_v3/backend/cpu/include/cpu_ntt.h
```diff
@@ -0,0 +1,331 @@
...
+            current_elements[i + j] = u + v;
+            current_elements[i + j + half_len] = (u - v) * twiddles[tw_idx];
+          }
+        }
+      }
+    }
+  }
+
+  template <typename E = scalar_t>
+  void transpose(const E* input, E* output, int rows, int cols)
```
yshekel, line 171: note that there is a transpose api in vec_ops.h you can use if you need.
Also it's true for bit-reverse BTW.
That way you don't need to implement it again.

icicle_v3/backend/cpu/include/cpu_ntt.h
```diff
@@ -0,0 +1,331 @@
...
+    E* elements,
+    int batch_size,
+    const S* twiddles = nullptr,
+    int stride = 0,
+    const S* arbitrary_coset = nullptr,
+    bool bit_rev = false,
+    NTTDir dir = NTTDir::kForward,
+    bool columns_batch = false)
+  {
+    int size = 1 << logn;
```
yshekel, line 193: note that this can potentially overflow

icicle_v3/backend/cpu/include/cpu_ntt.h
```diff
@@ -0,0 +1,331 @@
...
+        }
+      }
+    }
+    return eIcicleError::SUCCESS;
+  }
+
+  template <typename S = scalar_t, typename E = scalar_t>
+  eIcicleError cpu_ntt_ref(const Device& device, const E* input, int size, NTTDir dir, NTTConfig<S>& config, E* output)
+  {
+    if (size & (size - 1)) { return eIcicleError::INVALID_ARGUMENT; }
```
yshekel, line 217: use ICICLE_LOG_ERROR to log errors to user in case of invalid inputs

icicle_v3/backend/cpu/include/cpu_ntt.h
```diff
@@ -0,0 +1,331 @@
...
+      std::copy(input, input + total_size, temp_elements.get());
+    }
+    const int logn = int(log2(size));
+    const S* twiddles = CpuNttDomain<S>::s_ntt_domain.get_twiddles();
+    const int domain_max_size = CpuNttDomain<S>::s_ntt_domain.get_max_size();
+    S* arbitrary_coset = nullptr;
+    int coset_stride = 0;
+
+    if (config.coset_gen != S::one()) {
+      for (int i = 1; i <= domain_max_size; i++) {
```
yshekel, line 238: this loop is not very efficient. In CUDA we do it in O(1), here O(n).
You can leave it like that because it's the ref but in next PR please make sure to make it faster.

icicle_v3/backend/cpu/include/cpu_ntt.h
```diff
@@ -0,0 +1,331 @@
...
+    const uint64_t total_size = size * config.batch_size;
+    auto temp_elements = std::make_unique<E[]>(total_size);
+    if (config.columns_batch) {
+      transpose(input, temp_elements.get(), size, config.batch_size);
+    } else {
+      std::copy(input, input + total_size, temp_elements.get());
+    }
+    const int logn = int(log2(size));
+    const S* twiddles = CpuNttDomain<S>::s_ntt_domain.get_twiddles();
+    const int domain_max_size = CpuNttDomain<S>::s_ntt_domain.get_max_size();
```
yshekel, line 233: you should check the domain is large enough or throw an error otherwise

icicle_v3/backend/cpu/include/cpu_ntt.h
```diff
@@ -0,0 +1,331 @@
...
+        auto temp_cosets = std::make_unique<S[]>(domain_max_size + 1);
+        S coset = S::one();
+        temp_cosets[0] = coset;
+        S coset_gen = dir == NTTDir::kForward ? config.coset_gen : S::inverse(config.coset_gen);
+        for (int i = 1; i <= domain_max_size; i++) {
+          coset = coset * coset_gen;
+          temp_cosets[i] = coset;
+        }
+        arbitrary_coset = temp_cosets.get();
+        temp_cosets.release();
```
yshekel, line 254: release?
Why don't you allocate the arbitrary coset directly? as a unique_ptr I mean

icicle_v3/backend/cpu/include/cpu_ntt.h
```diff
@@ -0,0 +1,331 @@
...
+    int coset_stride = 0;
+
+    if (config.coset_gen != S::one()) {
+      for (int i = 1; i <= domain_max_size; i++) {
+        if (twiddles[i] == config.coset_gen) {
+          coset_stride = i;
+          break;
+        }
+      }
+      if (coset_stride == 0) {
```
yshekel, line 244: please add comments explaining what is being done in every step. It would make the code easier to understand

icicle_v3/backend/cpu/include/cpu_ntt.h
```diff
@@ -0,0 +1,331 @@
...
+      coset_mul(
+        logn, domain_max_size, temp_elements.get(), config.batch_size, twiddles, coset_stride, arbitrary_coset,
+        input_rev);
+    }
+
+    if (need_to_reorder) { reorder_by_bit_reverse(logn, temp_elements.get(), config.batch_size); }
+
+    // NTT/INTT
+    if (dit) {
+      ICICLE_LOG_DEBUG << "DIT";
```
yshekel, line 296: either print a clear message or remove

icicle_v3/backend/cpu/include/cpu_ntt.h
```diff
@@ -0,0 +1,331 @@
...
+      dit_ntt<S, E>(temp_elements.get(), size, config.batch_size, twiddles, dir, domain_max_size);
+    } else {
+      ICICLE_LOG_DEBUG << "DIF";
+      dif_ntt<S, E>(temp_elements.get(), size, config.batch_size, twiddles, dir, domain_max_size);
+    }
+
+    if (dir == NTTDir::kInverse) {
+      // Normalize results
+      S inv_size = S::inv_log_size(logn);
+      ICICLE_LOG_DEBUG << "inv_size: " << inv_size;
```
yshekel, line 306: here too

yshekel (commented):
overall looks good but I wrote a few comments

icicle_v3/backend/cpu/include/cpu_ntt.h
ShanieWinitz: changed to "ntt_cpu"

icicle_v3/backend/cpu/include/cpu_ntt.h
ShanieWinitz: ok

icicle_v3/backend/cpu/include/cpu_ntt.h
ShanieWinitz: fixed

icicle_v3/backend/cpu/include/cpu_ntt.h
ShanieWinitz: why? isn't logn maximum 30?

icicle_v3/backend/cpu/include/cpu_ntt.h
ShanieWinitz: fixed

icicle_v3/backend/cpu/include/cpu_ntt.h
ShanieWinitz: fixed

icicle_v3/backend/cpu/include/cpu_ntt.h
ShanieWinitz: added TODO comment

yshekel (approved):

PR #534, opened 2024-06-05
MSM - supporting all window sizes

HadarIngonyama:
This PR enables using MSM with any value of c.

Note: default c isn't necessarily optimal, the user is expected to choose c and the precomputation factor that give the best results for the relevant case.

### DIFF
--- a/icicle/src/msm/msm.cu
+++ b/icicle/src/msm/msm.cu
@@ -904,8 +904,6 @@ namespace msm {
       output_bases, bases, sizeof(A) * bases_size,
       are_bases_on_device ? cudaMemcpyDeviceToDevice : cudaMemcpyHostToDevice, stream));
 
-    // unsigned c = 16;
-    printf("precomp c = %d\n", c);
     unsigned total_nof_bms = (P::SCALAR_FF_NBITS - 1) / c + 1;
     unsigned shift = c * ((total_nof_bms - 1) / precompute_factor + 1);
 
--- a/icicle/src/msm/msm.cu
+++ b/icicle/src/msm/msm.cu
@@ -94,10 +94,11 @@ namespace msm {
       unsigned buckets_per_bm,
       unsigned write_phase,
       unsigned step,
-      unsigned num_of_threads)
+      unsigned num_of_threads,
+      unsigned nof_source_buckets)
     {
       const int tid = blockIdx.x * blockDim.x + threadIdx.x;
-      if (tid >= num_of_threads) { return; }
+      if (tid >= num_of_threads) return;
 
       // we need shifted tid because we don't want to be reducing into zero buckets, this allows to skip them.
       // for write_phase==1, the read pattern is different so we don't skip over anything.
@@ -113,6 +114,7 @@ namespace msm {
       const unsigned v_r_key =
         write_stride ? ((write_ind / buckets_per_bm) * 2 + write_phase) * write_stride + write_ind % buckets_per_bm
                      : write_ind;
+      if (read_ind + jump >= nof_source_buckets) printf("tid %d, read_ind %d, write_ind %d\n", tid, read_ind, v_r_key);
       v_r[v_r_key] = v[read_ind] + v[read_ind + jump];
     }
 
@@ -770,6 +772,8 @@ namespace msm {
           CHK_IF_RETURN(cudaMallocAsync(
             &temp_buckets2, sizeof(P) * source_buckets_count / 2 * batch_size, stream));  // for type2 reduction (serial, top window - odds)
           initialize_buckets_kernel<<<(target_buckets_count * batch_size + 255)/256,256>>>(target_buckets, target_buckets_count * batch_size);
+          cudaDeviceSynchronize();
+              printf("cuda err 0: %d\n", cudaGetLastError());
 
           if (source_bits_count > 0) {  //why do we need this condition?
             for (unsigned j = 0; j < target_bits_count; j++) {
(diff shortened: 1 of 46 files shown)


HadarIngonyama:
Fixed precomputation tests in rust and go. Precomputaion function interface has changed as a result.

wrappers/golang/core/msm.go
```diff
@@ -32,19 +32,19 @@ type MSMConfig struct {
...
 	AreScalarsMontgomeryForm bool
 
-	arePointsOnDevice bool
+	ArePointsOnDevice bool
 
 	/// True if coordinates of points are in Montgomery form and false otherwise. Default value: true.
 	ArePointsMontgomeryForm bool
 
-	areResultsOnDevice bool
+	AreResultsOnDevice bool
```
jeremyfelder, line 47: These should remain private as the `XXXCheck` functions should manipulate them internally based on the slice arguments. Its less error prone for the user

wrappers/golang/curves/bls12377/tests/msm_test.go
```diff
@@ -184,16 +184,18 @@ func TestPrecomputeBase(t *testing.T) {
 			_, e := precomputeOut.Malloc(points[0].Size()*points.Len()*int(precomputeFactor), points[0].Size())
 			assert.Equal(t, e, cr.CudaSuccess, "Allocating bytes on device for PrecomputeBases results failed")
 
-			e = msm.PrecomputeBases(points, precomputeFactor, 0, &cfg.Ctx, precomputeOut)
+			cfg.PrecomputeFactor = precomputeFactor
+			cfg.PointsSize = int32(points.Len())
+			cfg.ArePointsOnDevice = points.IsOnDevice()
```
jeremyfelder, line 189: Similar to the previous comment, this should be checked and updated in `core.PrecomputeBasesCheck` which is called in `msm.PrecomputeBases`

icicle/src/msm/msm.cu
```diff
@@ -74,10 +73,10 @@ namespace msm {
       unsigned* bucket_sizes_sum,
       unsigned* bucket_sizes,
       unsigned* large_bucket_thread_indices,
-      unsigned num_of_threads)
+      unsigned nof_threads)
     {
       const int tid = blockIdx.x * blockDim.x + threadIdx.x;
-      if (tid >= num_of_threads) { return; }
+      if (tid >= nof_threads) { return; }
```
jeremyfelder, line 79: We should be consistent on formatting; see line 99

```suggestion
      if (tid >= nof_threads) return;
```

icicle/src/msm/tests/msm_test.cu
```diff
@@ -170,65 +195,93 @@ int main()
...
-    false, // is_big_triangle
-    true,  // is_async
+    ctx,            // DeviceContext
+    N,              // points_size
+    precomp_factor, // precompute_factor
+    user_c,         // c
+    0,              // bitsize
+    10,             // large_bucket_factor
+    batch_size,     // batch_size
+    false,          // are_scalars_on_device
```
yshekel, line 205: why don't you use scalar_d if you already copied them to device? intentional? if yes, no need to allocate scalars_d and copy.

icicle/src/msm/msm.cu
```diff
@@ -726,62 +745,90 @@ namespace msm {
         NUM_BLOCKS = (nof_bms_in_batch + NUM_THREADS - 1) / NUM_THREADS;
         big_triangle_sum_kernel<<<NUM_BLOCKS, NUM_THREADS, 0, stream>>>(buckets, final_results, nof_bms_in_batch, c);
       } else {
+        // the recursive reduction algorithm works with 2 types of reduction that can run on parallel streams
```
yshekel, line 748: any chance you move all this reduction logic to a 'reduction_phase()' function? thanks

icicle/src/msm/tests/msm_test.cu
HadarIngonyama: It's just a test file, I copy to have all options available, I don't use scalar_d because in zprize you need to include scalar transfer time

icicle/src/msm/msm.cu
HadarIngonyama: in that case we should also have scalar_splitting_phase() sorting_phase() accumulation_phase() and so on, if we want this refactoring let's do it in a different PR

wrappers/golang/curves/bls12377/tests/msm_test.go
HadarIngonyama: ok, will fix, what about the rust test? is it important there too?

icicle/src/msm/msm.cu
```diff
@@ -846,32 +887,25 @@ namespace msm {
...
   template <typename A, typename P>
-  cudaError_t precompute_msm_bases(
-    A* bases,
-    int bases_size,
-    int precompute_factor,
-    int _c,
-    bool are_bases_on_device,
-    device_context::DeviceContext& ctx,
-    A* output_bases)
+  cudaError_t precompute_msm_bases(A* bases, int msm_size, MSMConfig& config, A* output_bases)
```
jeremyfelder, line 890: After discussing with the team, in order to prevent a breaking change right now, we think its best to:
1. create a second function that will call the old function after computing `c`
2. add a deprecation comment on the old function

wrappers/golang/curves/bls12377/tests/msm_test.go
jeremyfelder: Yes, there as well. `PrecomputeFactor` is fine to update manually as it is public but `points_size` is not public so it shouldn't be updated manually

icicle/src/msm/msm.cu
HadarIngonyama: actually, why do we need to call the old function? I will just give my new precompute function a different name and also keep the old one as deprecated.

yshekel: just so you don't implement twice.

jeremyfelder (approved):

LeonHibnik (approved):

PR #535, opened 2024-06-05
accumulate stwo

vhnatyk:
## Describe the changes

adds in-place vector addition and api as accumulate 

## Linked Issues

Resolves #


icicle/include/api/templates/fields/vec_ops.h
```diff
@@ -4,6 +4,9 @@ extern "C" cudaError_t ${FIELD}_mul_cuda(
 extern "C" cudaError_t ${FIELD}_add_cuda(
   ${FIELD}::scalar_t* vec_a, ${FIELD}::scalar_t* vec_b, int n, vec_ops::VecOpsConfig& config, ${FIELD}::scalar_t* result);
 
+extern "C" cudaError_t ${FIELD}_accumulate_cuda(
+  ${FIELD}::scalar_t* vec_a, ${FIELD}::scalar_t* vec_b, int n, vec_ops::VecOpsConfig& config);
+
```
jeremyfelder, line 9: Where is the implementation? I only see the header definitions

icicle/include/api/stark252.h
```diff
@@ -19,6 +19,9 @@ extern "C" cudaError_t stark252_mul_cuda(
 extern "C" cudaError_t stark252_add_cuda(
   stark252::scalar_t* vec_a, stark252::scalar_t* vec_b, int n, vec_ops::VecOpsConfig& config, stark252::scalar_t* result);
 
+extern "C" cudaError_t stark252_accumulate_cuda(
+  stark252::scalar_t* vec_a, stark252::scalar_t* vec_b, int n, vec_ops::VecOpsConfig& config, stark252::scalar_t* result);
```
LeonHibnik, line 23:
```suggestion
  stark252::scalar_t* vec_a, stark252::scalar_t* vec_b, int n, vec_ops::VecOpsConfig& config);
```

wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -207,6 +213,23 @@ where
...
+    cfg: &VecOpsConfig,
+) -> IcicleResult<()>
+where
+    F: FieldImpl,
+    <F as FieldImpl>::Config: VecOps<F>,
+{
+    assert_eq!(
+        cfg.is_a_on_device, cfg.is_result_on_device,
+        "input vector 'a' must have same 'on device' setting as result, since operation is performed in-place"
+    );
```
jeremyfelder, line 228: Is this needed? These values are not publicly accessible and are updated in the next line via `check_vec_ops_args` where `a` is passed as both `a` and `result` so it will always be true

icicle/src/vec_ops/vec_ops.cu
```diff
@@ -131,19 +132,19 @@ namespace vec_ops {
   }
 
   template <typename E>
-  cudaError_t mul(const E* vec_a, const E* vec_b, int n, VecOpsConfig& config, E* result)
+  cudaError_t mul(E* vec_a, const E* vec_b, int n, VecOpsConfig& config, E* result)
```
yshekel, line 135: why did you remove the consts? should this modify the inputs?

LeonHibnik: It should be in-place so yes

wrappers/rust/icicle-core/src/vec_ops/mod.rs
vhnatyk: removed

icicle/include/api/templates/fields/vec_ops.h
vhnatyk: added, was using add_inplace instead via add template

wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -83,6 +83,12 @@ pub trait VecOps<F> {
         cfg: &VecOpsConfig,
     ) -> IcicleResult<()>;
 
+    fn add_inplace(
```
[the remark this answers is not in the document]
vhnatyk: as @LeonHibnik pointed it's actually still allocating result due to missing logic - fixing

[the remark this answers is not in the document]
vhnatyk: fixed

icicle/include/api/babybear.h
```diff
@@ -56,6 +56,9 @@ extern "C" cudaError_t babybear_mul_cuda(
 extern "C" cudaError_t babybear_add_cuda(
   babybear::scalar_t* vec_a, babybear::scalar_t* vec_b, int n, vec_ops::VecOpsConfig& config, babybear::scalar_t* result);
 
+extern "C" cudaError_t babybear_accumulate_cuda(
+  babybear::scalar_t* vec_a, babybear::scalar_t* vec_b, int n, vec_ops::VecOpsConfig& config);
+
```
jeremyfelder, line 61: I don't see the extension definition for babybear and the vec_ops_ext header is not excluded in the generation script. Any idea why it wasn't generated?

icicle/src/vec_ops/vec_ops.cu
```diff
@@ -109,41 +112,49 @@ namespace vec_ops {
...
 
-    if (!config.is_a_on_device) { CHK_IF_RETURN(cudaFreeAsync(d_alloc_vec_a, config.ctx.stream)); }
-    if (!config.is_b_on_device) { CHK_IF_RETURN(cudaFreeAsync(d_alloc_vec_b, config.ctx.stream)); }
-
     if (!config.is_result_on_device) {
       CHK_IF_RETURN(cudaMemcpyAsync(result, d_result, n * sizeof(E), cudaMemcpyDeviceToHost, config.ctx.stream));
       CHK_IF_RETURN(cudaFreeAsync(d_result, config.ctx.stream));
     }
 
+    if (!config.is_a_on_device && !is_in_place) { CHK_IF_RETURN(cudaFreeAsync(d_alloc_vec_a, config.ctx.stream)); }
```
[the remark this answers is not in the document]
vhnatyk: if you mean d_result - it's freed at line 133, result should not be freed if not in place.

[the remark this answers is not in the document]
jeremyfelder: `d_result` should only be freed when `!config.is_result_on_device && !is_in_place` though, otherwise isn't it possible that d_result will point to `d_vec_a` and potentially free that?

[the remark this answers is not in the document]
jeremyfelder: Also, why does freeing `d_alloc_vec_a` depend on in place or not? It is only being allocated based on on_device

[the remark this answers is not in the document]
vhnatyk: because if in-place and on device `d_vec_a = d_alloc_vec_a = result = d_result`, thus we avoid freeing result if on device and double-free exception if not, we have tests covering all the cases, though not one test

icicle/include/api/babybear.h
vhnatyk: hi - as we discussed on daily few days ago and as @ChickenLover  suggested - it would be great to regenerate headers from script, however I tried and seems like it's worth separate pr as maybe won't be relevant to v3? and the templates differ significantly including formatting seems like quite few methods were added manually. I tried to fix them but it's a bit imho overkill for this pr and worth separate if will be actual, wdyt?

jeremyfelder (approved):

PR #537, opened 2024-06-06
Feat/vlad/signed msm

vladfdp:
## Describe the changes

This PR adds an option to run msm using signed digits.
Currently only works with big triangle
I set the default is_signed to false because big triangle is false by default.
I made the split scalars kernel into two different ones for readibility but I can make it a single one if necessary

## Linked Issues

Resolves #

### DIFF
--- a/icicle/include/msm/msm.cuh
+++ b/icicle/include/msm/msm.cuh
@@ -79,7 +79,7 @@ namespace msm {
                                  *   non-blocking and you'd need to synchronize it explicitly by running
                                  *   `cudaStreamSynchronize` or `cudaDeviceSynchronize`. If set to false, the MSM
                                  *   function will block the current CPU thread. */
-    bool is_signed; /**< Whether to run the MSM usign the signed digit trick. If set to true, the MSM function will use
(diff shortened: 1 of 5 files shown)


icicle/include/msm/msm.cuh
```diff
@@ -103,6 +106,7 @@ namespace msm {
       false, // are_results_on_device
       false, // is_big_triangle
       false, // is_async
+      false, // is_signed
```
jeremyfelder, line 109: If this allows for using less memory AND is faster, why don't we enable is as a default?

yshekel: he mentioned that it only works for with the big-triangle accumulation that is slower compare to the other accumulation algorithm.
@vladfdp can you say how does big-triangle + signed compared to the other accumulation without signed optimization?

wrappers/golang/core/msm.go
```diff
@@ -73,6 +77,7 @@ func GetDefaultMSMConfig() MSMConfig {
 		false, // areResultsOnDevice
 		false, // IsBigTriangle
 		false, // IsAsync
+		false, //IsSigned
```
yshekel, line 80: missing space

PR #544, opened 2024-06-18
Added v2 paper

Otsar-Raikou:


docs/docs/icicle/polynomials/overview.md
```diff
@@ -1,5 +1,8 @@
 # Polynomial API Overview
 
+> **Note:** Read our paper on the Polynomials API in ICICLE v2 by clicking [here](https://eprint.iacr.org/2024/973).
```
jeremyfelder, line 3: I'm not sure we want it to be an admonition but if we do we should use Docusaurus' correct syntax: https://docusaurus.io/docs/markdown-features/admonitions
```suggestion
:::note

Read our paper on the Polynomials API in ICICLE v2 by clicking [here](https://eprint.iacr.org/2024/973).

:::
```

Maybe we want an announcement bar instead - https://docusaurus.io/docs/2.x/api/themes/configuration#announcement-bar

jeremyfelder: image
The strip at the top of the page is an example of the Announcement Bar. Note that it is on every page in the docs...

LeonHibnik (approved):

PR #547, opened 2024-06-26
Feat/m31

nonam3e:
This PR adds support of the m31 Field


scripts/gen_c_api.py
```diff
@@ -64,6 +64,12 @@
         "field_ext.h",
         "vec_ops_ext.h",
         "ntt_ext.h",
+    },
+    "m31": {
+        "ntt_ext.h",
+        "ntt.h",
```
yshekel, line 70: should there be ntt for m31 or not? based on here yes but the m31.h and also rust is missing ntt. Why is that?

nonam3e: m31 doesn't support the current ntt algorithm, fft over the extension field is planned

icicle/include/fields/stark_fields/m31.cuh
```diff
@@ -0,0 +1,246 @@
...
+#include "fields/quartic_extension.cuh"
+
+namespace m31 {
+  template <class CONFIG>
+  class MersenneField : public Field<CONFIG>
+  {
+  public:
+    HOST_DEVICE_INLINE MersenneField(const MersenneField& other) : Field<CONFIG>(other) {}
+    HOST_DEVICE_INLINE MersenneField(uint32_t x = 0) : Field<CONFIG>({x}) {}
+    HOST_DEVICE_INLINE MersenneField(storage<CONFIG::limbs_count> x) : Field<CONFIG>{x} {}
```
yshekel, line 14: why x is passed by value? probably should be const ref

icicle/include/fields/stark_fields/m31.cuh
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    // Define assignment operator
+    HOST_DEVICE_INLINE MersenneField& operator=(const Field<CONFIG>& other)
+    {
+      if (this != &other) { Field<CONFIG>::operator=(other); }
+      return *this;
+    }
+
+    // HOST_DEVICE_INLINE MersenneField& operator=(const uint32_t& other) {
```
yshekel, line 37: remove commented code or uncomment if you need it

icicle/include/fields/stark_fields/m31.cuh
vhnatyk, line 1: why implementation like operators etc in .cuh file - seems other .cuh files are just definitions of consts etc

scripts/gen_c_api.py
vhnatyk: @nonam3e  - both of them radix2 and mixed-radix ntt?

> m31 doesn't support the current ntt algorithm, fft over the extension field is planned

nonam3e: @vhnatyk it's not possible to generate omega, that is why it's not possible to support both of them

icicle/include/fields/stark_fields/m31.cuh
nonam3e: all the current field functionality is implemented in .cuh files (field.cuh, host_math.cuh etc), so it's better to refactor it with the whole field structure if it's needed

jeremyfelder (commented):
Small comments

yuvalingo (approved):
Comments for future version:
Need to have all methods for both M31 and M31W
For example: M31 = add(M31, M31) & M31W = add(M31W, M31W).
M31 should be able to be casted to M31W.
The only function that mixes M31 and M31W should be the reduction function that takes M31W and returns M31.
I would write this as two classes with casting.
That way operator+ overloading will be implicitly determined based on arguements.
Finally, I think the names should be M31 and M31W rather than MersenneField and Wide.

vhnatyk (approved):
lgtm, though implementation in cuh needs refactoring

PR #548, opened 2024-06-26
RISC0 example using Polynomial API

svpolonsky:
## New Example

This new c++ example shows the basics of RISC0 protocol using our Polynomial API

### DIFF
--- a/examples/c++/risc0/README.md
+++ b/examples/c++/risc0/README.md
@@ -1,0 +1,44 @@
+# ICICLE example: RISC0's Fibonacci sequence proof using Polynomial API
+
+## Why RISC0?
+
+[RISC0 Protocol](https://www.risczero.com/) creates computational integrity proofs (a.k.a. Zero Knowledge Proofs) for programs executing on RISC-V architecture.
+The proofs are created for sequences of values in RISC-V registers, called execution traces.
+This approach is transparent to developers and enables the use of general purpose languages.
+
+## Best-Practices
+
+This example builds on [ICICLE Polynomial API](../polynomial-api/README.md) so we recommend to run it first.
+
+## Key-Takeaway
+
+RISC0 encodes execution traces into very large polynomials and commits them using Merkle trees.
+FRI speeds-up validation of such commitments by recursively generating smaller polynomials (and trees) from larger ones.
+The key enabler for *recursion* is the *redundancy* of polynomial commitments, hence the use of Reed-Solomon codes.
+
+## Running the example
+
+To run example, from project root directory:
+
+```sh
+cd examples/c++/risc0
+./compile.sh
+./run.sh
+```
+
+## What's in the example
+
+The example follows [STARK by Hand](https://dev.risczero.com/proof-system/stark-by-hand), structured in the following Lessons:
+
+1. The Execution Trace
+2. Rule checks to validate a computation
+3. Padding the Trace
+4. Constructing Trace Polynomials
+5. ZK Commitments of the Trace Data
+6. Constraint Polynomials
+7. Mixing Constraint Polynomials
+8. The Core of the RISC Zero STARK
+9. The DEEP Technique
+10. Mixing (Batching) for FRI
+11. FRI Protocol (Commit Phase)
+12. FRI Protocol (Query Phase)
--- a/examples/c++/risc0/compile.sh
+++ b/examples/c++/risc0/compile.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# Exit immediately on error
+set -e
+
+mkdir -p build/example
+mkdir -p build/icicle
+
+# Configure and build Icicle
+# cmake -S ../../../icicle/ -B build/icicle -DCMAKE_BUILD_TYPE=Release -DCURVE=bn254 -DG2=OFF
+cmake -S ../../../icicle/ -B build/icicle -DCMAKE_BUILD_TYPE=Release -DFIELD=babybear -DG2=OFF
+cmake --build build/icicle
+
+# Configure and build the example application
+cmake -S . -B build/example
+cmake --build build/example
--- a/examples/c++/risc0/compile.sh
+++ b/examples/c++/risc0/compile.sh
@@ -7,8 +7,7 @@ mkdir -p build/example
 mkdir -p build/icicle
 
 # Configure and build Icicle
-# cmake -S ../../../icicle/ -B build/icicle -DCMAKE_BUILD_TYPE=Release -DCURVE=bn254 -DG2=OFF
-cmake -S ../../../icicle/ -B build/icicle -DCMAKE_BUILD_TYPE=Release -DFIELD=babybear -DG2=OFF
+cmake -S ../../../icicle/ -B build/icicle -DCMAKE_BUILD_TYPE=Release -DFIELD=babybear
 cmake --build build/icicle
 
 # Configure and build the example application
--- a/examples/c++/risc0/example.cu
+++ b/examples/c++/risc0/example.cu
@@ -36,23 +36,15 @@ void p_print(Polynomial_t * p, int logn, scalar_t shift, std::string header = "P
 
 // value to polynomial
 Polynomial_t p_value(scalar_t value) {
-  auto coeff = std::make_unique<scalar_t[]>(1);
-  coeff[0] = value;
-  auto p_value = Polynomial_t::from_coefficients(coeff.get() , 1);
+  auto p_value = Polynomial_t::from_coefficients(&value , 1);
   return p_value;
 }
 
 Polynomial_t p_rotate(Polynomial_t* p, int logn) {
+  // rotate polynomial coefficients right by one position
   auto n = 1 << logn;
   auto evaluations_rou_domain = std::make_unique<scalar_t[]>(n);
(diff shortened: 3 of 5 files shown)


examples/c++/risc0/compile.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# Exit immediately on error
+set -e
+
+mkdir -p build/example
+mkdir -p build/icicle
+
+# Configure and build Icicle
+# cmake -S ../../../icicle/ -B build/icicle -DCMAKE_BUILD_TYPE=Release -DCURVE=bn254 -DG2=OFF
+cmake -S ../../../icicle/ -B build/icicle -DCMAKE_BUILD_TYPE=Release -DFIELD=babybear -DG2=OFF
```
yshekel, line 11: I would remove the G2 param which is only relevant to curves. Or replace with EXT_FIELD which could be on here.
Also please deelte the commented line

examples/c++/risc0/example.cu
```diff
@@ -0,0 +1,284 @@
...
+    std::cout << i << ": " << (*p)(x) << std::endl;
+    x = x*omega;
+  }
+}
+
+// value to polynomial
+Polynomial_t p_value(scalar_t value) {
+  auto coeff = std::make_unique<scalar_t[]>(1);
+  coeff[0] = value;
+  auto p_value = Polynomial_t::from_coefficients(coeff.get() , 1);
```
yshekel, line 41: You can use the value directly, no need to copy it to heap:

auto p_value = Polynomial_t::from_coefficients(&value , 1);

examples/c++/risc0/example.cu
```diff
@@ -0,0 +1,284 @@
...
+}
+
+Polynomial_t p_rotate(Polynomial_t* p, int logn) {
+  auto n = 1 << logn;
+  auto evaluations_rou_domain = std::make_unique<scalar_t[]>(n);
+  // TBD: rebase to use evaluate_on_rou_domain
+  // p->evaluate_on_rou_domain(logn, evaluations_rou_domain);
+  auto omega = scalar_t::omega(logn);
+  auto x = scalar_t::one();
+  for (int i = 0; i < n; ++i) {
```
yshekel, line 52: you can do this via poly api and have it compute way faster
https://github.com/ingonyama-zk/icicle/blob/f812f071fa11144dab08e9abbd2f2be872b861b2/icicle/include/polynomials/polynomials.h#L71

examples/c++/risc0/example.cu
```diff
@@ -0,0 +1,284 @@
...
+
+// value to polynomial
+Polynomial_t p_value(scalar_t value) {
+  auto coeff = std::make_unique<scalar_t[]>(1);
+  coeff[0] = value;
+  auto p_value = Polynomial_t::from_coefficients(coeff.get() , 1);
+  return p_value;
+}
+
+Polynomial_t p_rotate(Polynomial_t* p, int logn) {
```
yshekel, line 45: rotate right or left?

examples/c++/risc0/example.cu
yshekel: actually I see you have a TBD for that

examples/c++/risc0/example.cu
```diff
@@ -0,0 +1,284 @@
...
+  // p->evaluate_on_rou_domain(logn, evaluations_rou_domain);
+  auto omega = scalar_t::omega(logn);
+  auto x = scalar_t::one();
+  for (int i = 0; i < n; ++i) {
+    evaluations_rou_domain[i] = (*p)(x);
+    x = x*omega;
+  }
+  scalar_t tmp  = evaluations_rou_domain[n-1];
+  for (int i = n-1; i > 0; --i) {
+    evaluations_rou_domain[i] = evaluations_rou_domain[i-1];
```
yshekel, line 58: Maybe this can be computed via coset ntt by evaluate_on_rou_domain and avoiding this copy loop.

examples/c++/risc0/compile.sh
krakhit: I agree...please remove

examples/c++/risc0/example.cu
```diff
@@ -0,0 +1,284 @@
...
+}
+
+std::unique_ptr<scalar_t[]> InterpolateRootsOfUnity(Polynomial_t * p, int n, scalar_t shift = scalar_t::one()) {
+  const int deg = p->degree();
+  auto input = std::make_unique<scalar_t[]>(n);
+  // TBD: check if scalar_t constructor initializes to zero
+  for (int i = 0; i < n; ++i) {
+    input[i] = scalar_t::zero();
+  }
+  p->copy_coeffs(input.get(), 0/*start*/, deg);
```
yshekel, line 87: I think you should use a coeff view and compute the ntt directly from that 
https://github.com/ingonyama-zk/icicle/blob/f812f071fa11144dab08e9abbd2f2be872b861b2/icicle/include/polynomials/polynomials.h#L82

Just remember to set ntt_config.is_input_on_device=true (or some flag named like that)

examples/c++/risc0/example.cu
```diff
@@ -0,0 +1,284 @@
...
+  rv_t rv_c3_trace[] = {0, 0, 0, 1, 0, 0, 0, 0};
+  auto c1_trace = std::make_unique<scalar_t[]>(n);
+  auto c2_trace = std::make_unique<scalar_t[]>(n);
+  auto c3_trace = std::make_unique<scalar_t[]>(n);
+  to_ff(rv_c1_trace, c1_trace.get(), n);
+  to_ff(rv_c2_trace, c2_trace.get(), n);
+  to_ff(rv_c3_trace, c3_trace.get(), n);
+
+  std::cout << "Lesson 2: Rule checks to validate a computation" << std::endl;
+  std::cout << "We defined functions to create rule-checking polynomials. Their names start with p_rule_check." << std::endl;
```
yshekel, line 139: Is this part missing?

examples/c++/risc0/example.cu
yshekel: update: oh I see that you use it for the blowup part. Maybe the name of the function could be better but anyway in that case I see why you cannot use views. Maybe we need NTT to be able to compute from a short input and use zeros when accessing outside of its size. so we can compute 4N evals from N coeffs without any expensive memory padding.

examples/c++/risc0/compile.sh
svpolonsky: done

examples/c++/risc0/example.cu
svpolonsky: done

examples/c++/risc0/example.cu
svpolonsky: done

examples/c++/risc0/example.cu
svpolonsky: added comment: rotate polynomial coefficients right by one position

examples/c++/risc0/example.cu
svpolonsky: NTT would be n*logn, copy is n. I'd keep as is.

examples/c++/risc0/example.cu
svpolonsky: changed function name from InterpolateRootsOfUnity to InterpolateOnLargerDomain

examples/c++/risc0/example.cu
svpolonsky: Changed to "We use rule-checking polynomials."

examples/c++/risc0/example.cu
yshekel: You do the ntt anyway. I meant that you can avoid the O(n) copy and make it O(1), since you shift by one element

examples/c++/risc0/example.cu
yshekel: I would rename the method but ok

yshekel (approved):

examples/c++/risc0/example.cu
```diff
@@ -0,0 +1,275 @@
...
+  coeffs[0] = ya - coeffs[1] * xa;
+}
+
+std::unique_ptr<scalar_t[]> InterpolateOnLargerDomain(Polynomial_t * p, int n, scalar_t shift = scalar_t::one()) {
+  const int deg = p->degree();
+  auto input = std::make_unique<scalar_t[]>(n);
+  // TBD: check if scalar_t constructor initializes to zero
+  for (int i = 0; i < n; ++i) {
+    input[i] = scalar_t::zero();
+  }
```
jeremyfelder, line 78: I believe the first line here will create value-initalized scalars: https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique.

jeremyfelder (approved):

examples/c++/risc0/README.md
```diff
@@ -0,0 +1,44 @@
...
+```sh
+cd examples/c++/risc0
+./compile.sh
+./run.sh
+```
+
+## What's in the example
+
+The example follows [STARK by Hand](https://dev.risczero.com/proof-system/stark-by-hand), structured in the following Lessons:
+
```
krakhit, line 32: Could you just highlight here the parts that are actually implemented in the code. For example, commitments are not.

PR #550, opened 2024-06-28
implement mmcs

ChickenLover:


icicle/include/hash/hash.cuh
```diff
@@ -123,6 +126,17 @@ namespace hash {
       Image* output,
       const device_context::DeviceContext& ctx) const = 0;
 
+    virtual cudaError_t absorb_2d(
+      const Matrix<PreImage>* inputs,
+      Image* states,
+      unsigned int number_of_inputs,
+      uint64_t number_of_rows,
```
yshekel, line 133: what is number_of_rows? of what? Matrices have [width, height] fields so it's not clear to me what this refers to.

icicle/src/merkle-tree/mmcs.cu
```diff
@@ -0,0 +1,458 @@
...
+
+using matrix::Matrix;
+
+namespace merkle_tree {
+
+  template <typename L, typename D>
+  cudaError_t hash_leaves(
+    const Matrix<L>* leaves,
+    unsigned int number_of_inputs,
+    uint64_t number_of_rows,
```
yshekel, line 15: what is number_of_rows? of what? Matrices have [width, height] fields so it's not clear to me what this refers to.

icicle/src/merkle-tree/mmcs.cu
```diff
@@ -0,0 +1,458 @@
...
+    D* states,
+    D* digests,
+    unsigned int digest_elements,
```
yshekel, line 18: digest_elements is what? total size of this array?

icicle/src/merkle-tree/mmcs.cu
```diff
@@ -0,0 +1,458 @@
...
+    const SpongeHasher<L, D>& hasher,
+    const device_context::DeviceContext& ctx)
+  {
+    SpongeConfig sponge_config = default_sponge_config(ctx);
+    sponge_config.are_inputs_on_device = true;
```
yshekel, line 23: what is this assumption about being on device? why not pass the config from outside and avoid this assumption?

icicle/src/merkle-tree/mmcs.cu
```diff
@@ -0,0 +1,458 @@
...
+  struct SubtreeParams {
+    unsigned int number_of_inputs;
+    unsigned int arity;
+    unsigned int digest_elements;
+    size_t number_of_rows;
+    size_t number_of_rows_padded;
+    size_t subtree_idx;
+    size_t number_of_subtrees;
+    uint64_t subtree_height;
+    size_t segment_size;
```
yshekel, line 53: what is segment? maybe add a comment

icicle/src/merkle-tree/mmcs.cu
```diff
@@ -0,0 +1,458 @@
...
+    }
+
+    return CHK_LAST();
+  }
+
+  template <typename L, typename D>
+  struct SubtreeParams {
+    unsigned int number_of_inputs;
+    unsigned int arity;
+    unsigned int digest_elements;
```
yshekel, line 47: is it digest size?

icicle/src/merkle-tree/mmcs.cu
yshekel: I figured out (1) but still it's not that clear from looking at the API only

icicle/src/merkle-tree/mmcs.cu
```diff
@@ -0,0 +1,458 @@
...
+    cudaStream_t& stream = tree_config.ctx.stream;
+
+    if (number_of_inputs == 0) THROW_ICICLE_ERR(IcicleError_t::InvalidArgument, "No matrices provided");
+
+    if (compression.preimage_max_length < tree_config.arity * tree_config.digest_elements)
+      THROW_ICICLE_ERR(
+        IcicleError_t::InvalidArgument,
+        "Hash max preimage length does not match merkle tree arity multiplied by digest elements");
+
+    std::vector<Matrix<L>> sorted_inputs(number_of_inputs);
```
yshekel, line 214: might be faster to use a list rather than a vector for the sorting if you have a lot of inputs

icicle/src/merkle-tree/mmcs.cu
```diff
@@ -0,0 +1,458 @@
...
+    std::partial_sort_copy(
+      inputs, inputs + number_of_inputs, sorted_inputs.begin(), sorted_inputs.end(),
+      [](const Matrix<L>& left, const Matrix<L>& right) { return left.height > right.height; });
```
yshekel, line 217: when two matrices have the same height, you could have two valid orders. Can it be a problem if the order is not deterministic?

icicle/src/merkle-tree/mmcs.cu
```diff
@@ -0,0 +1,458 @@
...
+    //       std::cout << inputs[i].values[j * inputs[i].width + k] << " ";
+    //     }
+    //     std::cout << std::endl;
+    //   }
+    //   std::cout << "Matrix " << i << ": " << sorted_inputs[i].width << ", " << sorted_inputs[i].height << std::endl;
+    // }
+
+    uint64_t max_height = sorted_inputs[0].height;
+
+    // Calculate maximum additional memory needed for injected matrices
```
yshekel, line 243: please make this part a separate function

icicle/src/merkle-tree/mmcs.cu
```diff
@@ -0,0 +1,458 @@
...
+    }
+
+    std::cout << "Max aux total elements " << max_aux_total_elements << std::endl;
+
+    uint64_t number_of_bottom_layer_rows = next_pow_of_two(max_height);
+    size_t leaves_info_memory = number_of_inputs * sizeof(Matrix<L>);
+
+    unsigned int tree_height = get_height(number_of_bottom_layer_rows);
+
+    // This will determine how much splitting do we need to do
```
yshekel, line 270: here too if you can make it a separate function it would be easier to understand what's going on here

icicle/src/merkle-tree/mmcs.cu
```diff
@@ -0,0 +1,458 @@
...
+      subtree_leaves_memory = std::max(subtree_leaves_elements * sizeof(L), subtree_digests_size * sizeof(D));
+      subtree_memory_required =
+        sizeof(D) * subtree_states_size + subtree_leaves_memory + subtree_aux_elements * sizeof(L) + leaves_info_memory;
+    }
+    unsigned int cap_height = tree_height - subtree_height;
+    size_t caps_len = pow(tree_config.arity, cap_height) * tree_config.digest_elements;
+
+    size_t available_memory, _total_memory;
+    CHK_IF_RETURN(cudaMemGetInfo(&available_memory, &_total_memory));
+    available_memory -= GIGA / 8; // Leave 128 MB just in case
```
yshekel, line 304: note that maybe you don't even have 128MB free and you would wrap around here.

icicle/src/merkle-tree/mmcs.cu
```diff
@@ -0,0 +1,458 @@
...
+
+    // We can effectively parallelize memory copy with streams
+    // as long as they don't operate on more than `STREAM_CHUNK_SIZE` bytes
+    const size_t number_of_streams = std::min((uint64_t)(available_memory / STREAM_CHUNK_SIZE), number_of_subtrees);
+    cudaStream_t* streams = static_cast<cudaStream_t*>(malloc(sizeof(cudaStream_t) * number_of_streams));
```
yshekel, line 309: minor better use std::vector here or maybe

icicle/src/merkle-tree/mmcs.cu
```diff
@@ -0,0 +1,458 @@
...
+    for (size_t i = 0; i < number_of_streams; i++) {
+      CHK_IF_RETURN(cudaStreamCreate(&streams[i]));
+    }
+
+    bool caps_mode = tree_config.keep_rows && tree_config.keep_rows <= cap_height;
```
yshekel, line 314: please add comment explaining what caps are

icicle/src/merkle-tree/mmcs.cu
```diff
@@ -0,0 +1,458 @@
...
+    // #endif
+
+    // Allocate memory for the states, injected leaves and digests
+    // These are shared by streams in a pool
+    D* states_ptr;
+    L *aux_ptr, *leaves_ptr;
+    CHK_IF_RETURN(cudaMallocAsync(&states_ptr, subtree_states_size * number_of_streams * sizeof(D), stream));
+    CHK_IF_RETURN(cudaMemsetAsync(states_ptr, 0, subtree_states_size * number_of_streams * sizeof(D), stream));
+    CHK_IF_RETURN(cudaMallocAsync(&leaves_ptr, subtree_leaves_memory * number_of_streams, stream));
+    CHK_IF_RETURN(cudaMallocAsync(&aux_ptr, subtree_aux_elements * number_of_streams * sizeof(L), stream));
```
yshekel, line 343: what is aux? please add comment

icicle/src/merkle-tree/mmcs.cu
```diff
@@ -0,0 +1,458 @@
...
+    std::cout << "Cap height = " << cap_height << std::endl;
+    std::cout << "Enabling caps mode? " << caps_mode << std::endl;
+    std::cout << std::endl;
+    // #endif
+
+    // Allocate memory for the states, injected leaves and digests
+    // These are shared by streams in a pool
+    D* states_ptr;
+    L *aux_ptr, *leaves_ptr;
+    CHK_IF_RETURN(cudaMallocAsync(&states_ptr, subtree_states_size * number_of_streams * sizeof(D), stream));
```
yshekel, line 340: did you account for those allocations? I mean where you leave 128MB and assume the rest is going to be used.

icicle/src/merkle-tree/mmcs.cu
```diff
@@ -0,0 +1,458 @@
...
+    CHK_IF_RETURN(cudaMallocAsync(&aux_ptr, subtree_aux_elements * number_of_streams * sizeof(L), stream));
+    // Wait for these allocations to finish
+    CHK_IF_RETURN(cudaStreamSynchronize(stream));
+
+    std::cout << "Allocated " << subtree_states_size * number_of_streams << " elements for states" << std::endl;
+    std::cout << "Allocated " << subtree_leaves_memory * number_of_streams << " bytes for leaves" << std::endl;
+    std::cout << "Allocated " << subtree_aux_elements * number_of_streams << " elements for aux leaves" << std::endl;
+
+    // Build subtrees in parallel. This for loop invokes kernels that can run in a pool of size `number_of_streams`
+    for (size_t subtree_idx = 0; subtree_idx < number_of_subtrees; subtree_idx++) {
```
yshekel, line 352: are subtrees only the bottom part of the tree? or also upper levels? I don't really understand that from the code. I honestly don't really understand the exact mapping of the tree to subtrees. Maybe a comment could help here.

icicle/src/merkle-tree/mmcs.cu
```diff
@@ -0,0 +1,458 @@
...
+      );
+      CHK_IF_RETURN(subtree_result);
+    }
+
+    for (size_t i = 0; i < number_of_streams; i++) {
+      CHK_IF_RETURN(cudaStreamSynchronize(streams[i]));
+    }
+
+    // Finish the top-level tree if any
+    if (cap_height > 0) {
```
yshekel, line 403: seems that this part is the higher level right? is it always one subtree?

icicle/src/merkle-tree/mmcs.cu
```diff
@@ -0,0 +1,458 @@
...
+        fold_layers<L, D>(sorted_inputs, states_ptr, digests_ptr, aux_ptr, d_leaves_info, digests, top_params);
+      CHK_IF_RETURN(fold_error);
+    }
+
+    if (caps_mode) { free(caps); }
+    CHK_IF_RETURN(cudaFreeAsync(states_ptr, stream));
+    CHK_IF_RETURN(cudaFreeAsync(leaves_ptr, stream));
+    if (!tree_config.is_async) return CHK_STICKY(cudaStreamSynchronize(stream));
+    for (size_t i = 0; i < number_of_streams; i++) {
+      CHK_IF_RETURN(cudaStreamSynchronize(streams[i]));
```
yshekel, line 451: why do you need to sync again here?

icicle/src/merkle-tree/mmcs.cu
```diff
@@ -0,0 +1,458 @@
...
+        d_leaves_info, leaves_info, params.number_of_leaves_to_inject * sizeof(Matrix<L>), cudaMemcpyHostToDevice,
+        params.ctx->stream));
+      free(leaves_info);
+    }
+
+    return CHK_LAST();
+  }
+
+  template <typename L, typename D>
+  cudaError_t maybe_copy_digests(D* digests, L* big_tree_digests, SubtreeParams<L, D>& params)
```
yshekel, line 108: please add comment when is it used and why

yshekel (commented):
I reviewed the mmcs.cu part as we discussed.
Overall I think it's ok in terms of the logic but hard to read. So many variables and the main function is quite long. 
I understand that it's complex and there are different cases that you need to account for but still it can be easier to read if you have smaller functions with clear names and add comments that can help understand what is going on and why.

questions:
(1) Is there a way to have this logic work also for non-matrices? I mean to build large trees where the leaves are field elements or something else? Obviously we can have 1X1 matrix but maybe it won't be efficient. Also not covering other types that are not field elements.
(2) did you consider multi-gpu? Seems that you are almost there.
(3) would that work with hash such as keccak as first level?
(4) I think this feature should have a lot of tests since the code is complex code and there are quite a lot of cases.

icicle/src/merkle-tree/mmcs.cu
ChickenLover: Number of rows is the number of rows of the matrices. This is equal to their `height`

icicle/src/merkle-tree/mmcs.cu
ChickenLover: It's not an assumption, it is a requirement of the current implementation

icicle/src/merkle-tree/mmcs.cu
ChickenLover: segment is a memory segment that corresponds to a single layer of merkle tree digests

icicle/src/merkle-tree/mmcs.cu
ChickenLover: Yes, I also thought about this - I am not sure. In Plonky3 it is also not deterministic

icicle/src/merkle-tree/mmcs.cu
```diff
@@ -0,0 +1,458 @@
...
+    }
+    uint64_t subtree_aux_elements = max_aux_total_elements;
+
+    size_t subtree_leaves_memory = std::max(subtree_leaves_elements * sizeof(L), subtree_digests_size * sizeof(D));
+    size_t subtree_memory_required =
+      sizeof(D) * subtree_states_size + subtree_leaves_memory + subtree_aux_elements * sizeof(L) + leaves_info_memory;
+    while (subtree_memory_required > STREAM_CHUNK_SIZE) {
+      number_of_subtrees *= tree_config.arity;
+      subtree_height--;
+      subtree_bottom_layer_rows /= tree_config.arity;
```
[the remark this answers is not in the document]
ChickenLover: yes, all the variables are derived from `number_of_bottom_layer_rows` which is a next_pow_of_two, and current mmcs only works for powers of two

icicle/src/merkle-tree/mmcs.cu
ChickenLover: yes

icicle/src/merkle-tree/mmcs.cu
ChickenLover: only at the bottom rn. Top subtrees only make sense for trees of height greater than ~45, which I think is not the case for any existing protocol.

icicle/src/merkle-tree/mmcs.cu
ChickenLover: yes and yes

icicle/src/merkle-tree/mmcs.cu
ChickenLover: actually I don't, looks like I forgot to remove that after some debugging

ChickenLover:
@yshekel 
1. For that use cases we have a simple merkle tree in `merkle.cu`
2. No, not really. In theory, one can use multiple GPUs for different subtrees
3. Not rn. Keccak needs to implement `absorb_2d` and `compress_and_inject` kernels. Also, current mmcs only works with field elements and not bytes or words
4. Yes, it should, I will add them

icicle/include/poseidon2/kernels.cuh
```diff
@@ -2,10 +2,27 @@
...
+#include "matrix/matrix.cuh"
 #include "poseidon2/constants.cuh"
 #include "gpu-utils/modifiers.cuh"
 
+using matrix::Matrix;
+
 namespace poseidon2 {
+  static DEVICE_INLINE unsigned int d_next_pow_of_two(unsigned int v)
+  {
+    v--;
```
jeremyfelder, line 16: If `v` is a power of two, should this still return the next power of two or `v`?

icicle/include/poseidon2/kernels.cuh
```diff
@@ -217,12 +225,168 @@ namespace poseidon2 {
...
+      state[i] = input[idx * input_len + i];
+    }
+
+    permute_state<S, T>(state, constants);
+
+    UNROLL
+    for (int i = 0; i < output_len; i++) {
+      output[idx * output_len + i] = state[i];
+    }
+  }
```
jeremyfelder, line 300: Does it make sense to have an "internal" kernel for these kernels since they are all extremely similar except for the lengths of the forloops?

icicle/include/poseidon2/kernels.cuh
ChickenLover: If `v` is a power of two, this should return `v`

icicle/include/poseidon2/kernels.cuh
ChickenLover: good catch, I actually needed to remove compress_many_kernel

PR #551, opened 2024-07-01
Field creation automated through macros

DmytroTym:
## Describe the changes

Params files for fields now only require modulus specified by the user (also twiddle generator and/or non-residue in case either or both are needed). Everything else gets generated by a macro which seems to be much cleaner and easier than using python scripts / doing it by hand like it used to be before.

## Future work

I also tried to do some general cleanup and minor perf improvements of primitives. Spent quite a bit of time trying to create correction-minimising formulas (sort of what Niall does [here](https://youtu.be/KAWlySN7Hm8?si=94BQG-TO4vnD9FbE&t=2111) but for all fields and not just EC formula for a specific base field). This turned out to be much harder than expected and though I think I got it most of the way, the performance gains don't seem to justify the remaining effort considering significant increase in code complexity. If anyone is interested in this direction of work please let me know, I have an (almost) working prototype locally.

Most primitives related improvements I can think of are easier to do so if anything is particularly important to icicle right now, let me know. I can probably work on it in the background. Cheers.


icicle/include/fields/params_gen.cuh
```diff
@@ -0,0 +1,146 @@
+#pragma once
+#ifndef PARAMS_GEN_H
+#define PARAMS_GEN_H
+
+#include "storage.cuh"
+#include "host_math.cuh"
+
+using namespace host_math;
```
jeremyfelder, line 8: I think its better to not use this so the functions need the namespace specifier and its easier to follow where the functions are coming from.

It would also allow us to remove the `_host` suffix on the function names

icicle/include/fields/params_gen.cuh
```diff
@@ -0,0 +1,146 @@
...
+      } else {
+        rs = left_shift<NLIMBS, 1>(rs);
+        storage<NLIMBS> temp = {};
+        rs = add_sub_limbs_host<NLIMBS, true, true>(rs, modulus, temp) ? rs : temp;
+      }
+    }
+    return rs;
+  }
+
+  constexpr unsigned floorlog2(uint32_t x) { return x == 1 ? 0 : 1 + floorlog2(x >> 1); }
```
jeremyfelder, line 56: Is this faster than floor(std::log2(x))?

DmytroTym: I just need constexpr behaviour and `std::log2(x)` isn't constexpr. This is only run in compile time so the speed is not that important I guess

jeremyfelder (approved):

LeonHibnik (approved):

PR #552, opened 2024-07-03
Fix/vlad/affine to proj

vladfdp:
## Describe the changes

This PR fixes some of the affine -> projective functions by adding a condition
If the point in affine form is zero then return the projective zero

## Linked Issues

Resolves #

### DIFF
--- a/wrappers/golang/internal/generator/curves/templates/curve.go.tmpl
+++ b/wrappers/golang/internal/generator/curves/templates/curve.go.tmpl
@@ -114,11 +114,8 @@ func (a {{.CurvePrefix}}Affine) ToProjective() {{.CurvePrefix}}Projective {
 	var z {{.CurvePrefix}}BaseField
 
 	if (a.X == z.Zero()) && (a.Y == z.Zero()) {
-		return {{.CurvePrefix}}Projective{
-			X: z.Zero(),
-			Y: z.One(),
-			Z: z.Zero(),
-		}
+        var p {{.CurvePrefix}}Projective
+		return p.Zero()
 	}
 
 	return {{.CurvePrefix}}Projective{
--- a/wrappers/golang/internal/generator/curves/templates/curve.go.tmpl
+++ b/wrappers/golang/internal/generator/curves/templates/curve.go.tmpl
@@ -44,9 +44,7 @@ func (p *{{.CurvePrefix}}Projective) FromAffine(a {{.CurvePrefix}}Affine) {{.Cur
 	z.One()
 
 	if (a.X == z.Zero()) && (a.Y == z.Zero()) {
-		p.X = z.Zero()
-		p.Y = z.One()
-		p.Z = z.Zero()
+		p.Zero()
 	}else{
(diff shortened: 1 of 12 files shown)


wrappers/golang/internal/generator/curves/templates/curve.go.tmpl
```diff
@@ -43,9 +43,15 @@ func (p *{{.CurvePrefix}}Projective) FromAffine(a {{.CurvePrefix}}Affine) {{.Cur
 	z := {{.CurvePrefix}}BaseField{}
 	z.One()
 
+	if (a.X == z.Zero()) && (a.Y == z.Zero()) {
+		p.X = z.Zero()
+		p.Y = z.One()
+		p.Z = z.Zero()
```
jeremyfelder, line 49: Better to use the Zero function from the Projective point directly so we don't have multiple locations of its internal logic
```suggestion
		p.Zero()
```

wrappers/golang/internal/generator/curves/templates/curve.go.tmpl
```diff
@@ -109,6 +115,14 @@ func (a *{{.CurvePrefix}}Affine) FromLimbs(x, y []uint32) {{.CurvePrefix}}Affine
 func (a {{.CurvePrefix}}Affine) ToProjective() {{.CurvePrefix}}Projective {
 	var z {{.CurvePrefix}}BaseField
 
+	if (a.X == z.Zero()) && (a.Y == z.Zero()) {
+		return {{.CurvePrefix}}Projective{
+			X: z.Zero(),
+			Y: z.One(),
+			Z: z.Zero(),
+		}
```
jeremyfelder, line 123: Same here as above
```suggestion
        var p {{.CurvePrefix}}Projective
		return p.Zero()
```

jeremyfelder (commented):
Looks good for the fix.

I don't think it is too much work to extern the [from_affine](https://github.com/ingonyama-zk/icicle/blob/main/icicle/include/curves/projective.cuh#L30) function in cpp and use that directly in Rust and Go so we truly have a single implementation for this logic

LeonHibnik (approved):

jeremyfelder:
>I don't think it is too much work to extern the [from_affine](https://github.com/ingonyama-zk/icicle/blob/main/icicle/include/curves/projective.cuh#L30) function in cpp and use that directly in Rust and Go so we truly have a single implementation for this logic

If we choose not to do this for the hotfix, then lets add a test case in both Rust and Go:
- Rust - https://github.com/ingonyama-zk/icicle/blob/main/wrappers/rust/icicle-core/src/tests.rs#L24
- Go
  - https://github.com/ingonyama-zk/icicle/blob/main/wrappers/golang/internal/generator/curves/templates/curve_test.go.tmpl#L38
  - https://github.com/ingonyama-zk/icicle/blob/main/wrappers/golang/internal/generator/curves/templates/curve_test.go.tmpl#L88

jeremyfelder (approved):

PR #557, opened 2024-07-18
hotfix for slow twiddles on small domain sizes

vhnatyk:
## Describe the changes

This PR is a hotfix for ntt domains of log2 sizes < 4

## Linked Issues

Resolves #

### DIFF
--- a/icicle/src/ntt/kernel_ntt.cu
+++ b/icicle/src/ntt/kernel_ntt.cu
@@ -706,7 +706,7 @@ namespace mxntt {
       // TODO: should throw exception
       // TODO: fast twiddles should be used only with mixed radix ntt
       // THROW_ICICLE_ERR(IcicleError_t::InvalidArgument, "log_size must be >= 4");
-      printf("warning: log_size must be at least 4, got %d\n", log_size); // TODO: logging
+      printf("warning: when using fast twiddles, log_size must be at least 4, got %d\n", log_size); // TODO: logging
       return cudaSuccess;
     }
 
--- a/icicle/src/ntt/kernel_ntt.cu
+++ b/icicle/src/ntt/kernel_ntt.cu
@@ -703,11 +703,8 @@ namespace mxntt {
     CHK_INIT_IF_RETURN();
 
     if (log_size < 4) {
-      // TODO: should throw exception
       // TODO: fast twiddles should be used only with mixed radix ntt
-      // THROW_ICICLE_ERR(IcicleError_t::InvalidArgument, "log_size must be >= 4");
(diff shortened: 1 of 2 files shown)


vhnatyk:
"proper fix" - the fast twiddles must only be used with mixed radix ntt, so should be part of config and not decoupled or validated. But maybe it's late to refactor v2 for this

icicle/src/ntt/kernel_ntt.cu
```diff
@@ -702,6 +702,14 @@ namespace mxntt {
   {
     CHK_INIT_IF_RETURN();
 
+    if (log_size < 4) {
+      // TODO: should throw exception
+      // TODO: fast twiddles should be used only with mixed radix ntt
+      // THROW_ICICLE_ERR(IcicleError_t::InvalidArgument, "log_size must be >= 4");
+      printf("warning: log_size must be at least 4, got %d\n", log_size); // TODO: logging
```
HadarIngonyama, line 709: "warning: when using fast twiddles, log_size must be at least 4..

HadarIngonyama (commented):
why print and not throw an exception?

yshekel (commented):
I would add the following to init_domain() just before constructing the fast twiddles, to disable it from there rather than calling a function that seems to return success but doesn't do anything.

fast_twiddles_mode &= domain.max_log_size >= 4; // fast twiddles cannot be built for smaller domain

icicle/src/ntt/kernel_ntt.cu
vhnatyk: umm, throwing would be a breaking change

vhnatyk:
> I would add the following to init_domain() just before constructing the fast twiddles, to disable it from there rather than calling a function that seems to return success but doesn't do anything.
> 
> fast_twiddles_mode &= domain.max_log_size >= 4; // fast twiddles cannot be built for smaller domain

so this would solve one particular call  to a `generate_external_twiddles_fast_twiddles_mode` , not fix the function itself, but since it's called only once perhaps it's indeed ok

icicle/src/ntt/ntt.cu
```diff
@@ -469,6 +469,8 @@ namespace ntt {
       CHK_IF_RETURN(mxntt::generate_external_twiddles_generic(
         primitive_root, domain.twiddles, domain.internal_twiddles, domain.basic_twiddles, domain.max_log_size,
         ctx.stream));
+      
+      fast_twiddles_mode &= domain.max_log_size >= 4; // fast twiddles cannot be built for smaller domain
```
HadarIngonyama, line 473: so the user will think he is using fast twiddles but he isn't? why not just have the exception? or at least add a print in this case

PR #558, opened 2024-07-18
Fix typos

omahs:
Fix typos


jeremyfelder (approved):
@omahs Thanks for finding all of those typos! 🚀 We really appreciate it

LeonHibnik (approved):

PR #570, opened 2024-08-04
feat: add `num_bits()` function

Ethan-000:
## Describe the changes

adding a `num_bits()` function similar to https://github.com/arkworks-rs/algebra/blob/dcf73a5f9610ba9d16a3c8e0de0b3835e5e5d5e4/ff/src/biginteger/mod.rs#L482

this could be useful for small field optimizations

## Linked Issues

Resolves #

### DIFF
--- a/icicle/include/fields/field.cuh
+++ b/icicle/include/fields/field.cuh
@@ -124,14 +124,13 @@ public:
    */
   static constexpr HOST_DEVICE_INLINE unsigned num_of_reductions() { return CONFIG::num_of_reductions; }
 
-  static __host__ __device__ unsigned num_bits(const Field& x) {
+  static __host__ __device__ unsigned num_bits(const Field& x)
+  {
     unsigned ret = 32 * TLC;
     for (unsigned i = TLC; i-- > 0;) {
       int leading = __clz(x.limbs_storage.limbs[i]);
       ret -= leading;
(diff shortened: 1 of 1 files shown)


LeonHibnik:
Thank you for your contribution @Ethan-000!
Can you format your code according to https://github.com/ingonyama-zk/icicle?tab=readme-ov-file#development-contributions

LeonHibnik (approved):

icicle/include/fields/field.cuh
```diff
@@ -124,6 +124,17 @@ public:
    */
   static constexpr HOST_DEVICE_INLINE unsigned num_of_reductions() { return CONFIG::num_of_reductions; }
 
+  static __host__ __device__ unsigned num_bits(const Field& x)
+  {
+    unsigned ret = 32 * TLC;
```
yshekel, line 129: Please use sizeof(x.limbs[0]) rather than hardcoding 32, in case we modify the limb size in the future.
Also please add a short comment explaining what it does.

yshekel (changes requested):

icicle/include/fields/field.cuh
Ethan-000: done

yshekel (approved):

PR #572, opened 2024-08-05
Feat/roman/dcct

ChickenLover:
## Describe the changes

This PR...

## Linked Issues

Resolves #


icicle/include/fields/stark_fields/m31.cuh
```diff
@@ -177,10 +186,29 @@ namespace m31 {
...
+      if (logn > CONFIG::ext_omegas_count) THROW_ICICLE_ERR(IcicleError_t::InvalidArgument, "Field: Invalid inv index");
+#else
+      if (logn > CONFIG::ext_omegas_count) {
+        printf(
+          "CUDA ERROR: field.cuh: error on inv_log_size(logn): logn(=%u) > omegas_count (=%u)", logn,
+          CONFIG::ext_omegas_count);
+        assert(false);
+      }
+#endif // __CUDA_ARCH__
+      // storage_array<CONFIG::omegas_count, 1> const inv = CONFIG::inv;
```
yshekel, line 203: delete comment

icicle/include/fields/stark_fields/m31.cuh
```diff
@@ -220,5 +249,56 @@ namespace m31 {
   /**
    * Extension field of `scalar_t` enabled if `-DEXT_FIELD` env variable is.
    */
-  typedef ExtensionField<fp_config, scalar_t> extension_t;
+  typedef ComplexExtensionField<fp_config, scalar_t> c_extension_t;
+
+  const c_extension_t ROU = {{2}, {1268011823}};
+
+  // namespace quad_extension_config {
```
yshekel, line 256: remove

icicle/include/ntt/ntt_impl.cuh
```diff
@@ -6,7 +6,10 @@
 #include "ntt/ntt.cuh" // for enum Ordering
 
 namespace mxntt {
-
+#ifdef DCCT
+  template <typename S, typename R>
```
yshekel, line 10: is R an extension field? why R?

yshekel (commented):
Overall looks good to me but I have a few comments:
(1) I don't see that you define DCCT in cmake (and cargo) for m31 and build dcct
(2) Probably NTT tests will fail so we need some correctness test in CI.
(3) It would be better in terms of maintainability if you split dcct code from ntt sources. Assuming it doesn't required you to refactor too much.

icicle/include/fields/stark_fields/m31.cuh
ChickenLover: done and done

vhnatyk (approved):
what's in pr lgtm - except rust bindings and tests, but we can add these as well as other todos.  @yshekel request is 100% legit, and a bit raw feel overall - let's merge it as we agreed on a daily and I need it asap)

PR #573, opened 2024-08-06
Add tasks manager classes to ease multithreading for cpu

Koren-Brand:
Added tasks_manager.h - a small library to aid multithreaded CPU coding.

### DIFF
--- a/icicle_v3/backend/cpu/include/tasks_manager.h
+++ b/icicle_v3/backend/cpu/include/tasks_manager.h
@@ -1,0 +1,227 @@
+#pragma once 
+#include <atomic>
+#include <thread> // TODO check windows support
+#include <stdexcept>
+
+#define LOG_TASKS_PER_THREAD 2
+#define TASKS_PER_THREAD (1 << LOG_TASKS_PER_THREAD)
+#define TASK_IDX_MASK (TASKS_PER_THREAD - 1)
+#define MANAGER_SLEEP_USEC 10
+#define THREAD_SLEEP_USEC 1
+
+template<class Task>
+class TasksManager {
+public:
+  // constructor
+  TasksManager(const int nof_workers);
+
+  // Get a task that isn't occupied
+  // return value signals if there is a valid result to be handled before dispatching a new calculation
+  // Blocking functions
+  bool get_free_task(Task*& task_slot);
+  void get_completed_task(Task*& completed_task); // NOTE the user must handle this task's result before asking for new free tasks
+private:
+  class Worker {
+  public:
+    Worker();
+    ~Worker();
+    void run();
+    // Get a task that isn't occupied
+    // return value signals if there is a valid result to be handled before dispatching a new calculation
+    bool get_free_task(Task*& task_slot);
+    void get_completed_task(Task*& completed_task); // NOTE the user must handle this task's result before asking for new free tasks
+    bool are_all_idle();
+  private:
+    std::thread task_executor;
+    std::vector<Task> mTasksFifo;
+    int tail;
+    int head;
+    bool kill;
+  };
+
+  std::vector<Worker> workers;
+};
+
+class TaskBase {
+  public:
+    TaskBase() : status(idle), father_fifo_tail(nullptr) {}
+    virtual void execute() = 0; // COMMENT should it be private? still needs to be friend of worker
+    void dispatch(); // USER FUNC
+
+    inline bool is_ready_for_work(); // TODO friend functions of worker
+    inline bool can_push_task();
+    inline bool has_result();
+    inline bool is_idle();
+    inline void set_working();
+    inline void set_pending_result();
+    inline void set_handled_result();
+
+    inline void link_father_tail(int* tail_pointer) { father_fifo_tail = tail_pointer; } 
+
+    // Blocking function
+    void wait_done(); // NOTE the user must handle this task's result before asking for new free tasks  
+
+  protected:
+    enum TaskStatus {idle, set_task, dispatched, working, pending_result}; // TODO CAPS and eTaskStatus
+    std::atomic<TaskStatus> status;
+  private:
+    int* father_fifo_tail;
+};
+
+
+bool TaskBase::is_ready_for_work() {
+  return status.load(std::memory_order_acquire) == dispatched;
+}
+
+bool TaskBase::can_push_task() {
+  TaskStatus curr_status = status.load(std::memory_order_acquire);
+  return curr_status == pending_result || curr_status == idle;
+}
+
+bool TaskBase::has_result() {
+  return status.load(std::memory_order_acquire) == pending_result;
+}
+
+bool TaskBase::is_idle() {
+  return status.load(std::memory_order_acquire) == idle;
+}
+
+void TaskBase::dispatch() {
+  (*father_fifo_tail)++;
+  status.store(dispatched, std::memory_order_release);
+}
+
+void TaskBase::set_working() {
+  status.store(working, std::memory_order_release);
+}
+
+void TaskBase::set_pending_result() {
+  status.store(pending_result, std::memory_order_release);
+}
+
+void TaskBase::set_handled_result() {
+  status.store(idle, std::memory_order_release);
+}
+
+void TaskBase::wait_done() {
+  while (has_result()) {
+    std::this_thread::sleep_for(std::chrono::microseconds(MANAGER_SLEEP_USEC));
+  }
+}
+
+template<class Task>
+TasksManager<Task>::Worker::Worker() 
+: mTasksFifo(TASKS_PER_THREAD), 
+  tail(0),
+  head(0),
+  kill(false)
+  {
+    for (Task& task : mTasksFifo) task.link_father_tail(&tail);
+    task_executor = std::thread(&TasksManager<Task>::Worker::run, this);
+  }
+
+template<class Task>
+TasksManager<Task>::Worker::~Worker() {
+  kill = true;
+  task_executor.join();
+}
+
+template<class Task>
+void TasksManager<Task>::Worker::run() {
+  while (true) {
+    for (head = 0; head < mTasksFifo.size(); head++)
+    {      
+      Task* task = &mTasksFifo[head];
+      if (!task->is_ready_for_work()) {
+        std::this_thread::sleep_for(std::chrono::microseconds(THREAD_SLEEP_USEC));        
+        continue;
+      }
+      task->set_working();
+      task->execute();
+      task->set_pending_result();
+    }
+    if (kill) return;
+  }
+}
+
+template<class Task>
+bool TasksManager<Task>::Worker::get_free_task(Task*& task_slot) {
+  for (int i = 0; i < mTasksFifo.size(); i++)
+  {
+    int tail_adjusted_idx = (i + tail) & TASK_IDX_MASK; // TODO check % is optimized when base is power of 2
+
+    if (mTasksFifo[tail_adjusted_idx].can_push_task()) // Either idle or pending_result are valid
+    {
+      task_slot = &mTasksFifo[tail_adjusted_idx];
+      return task_slot->has_result();
+    }
+  }
+  task_slot = nullptr;
+  return false;
+}
+
+template<class Task>
+void TasksManager<Task>::Worker::get_completed_task(Task*& task_slot) {
+  for (int i = 0; i < mTasksFifo.size(); i++)
+  {
+    int tail_adjusted_idx = (i + tail) & TASK_IDX_MASK;
+
+    if (mTasksFifo[tail_adjusted_idx].has_result())
+    {
+      task_slot = &mTasksFifo[tail_adjusted_idx];
+      return;
+    }
+  }
+  task_slot = nullptr;
+}
+
+template<class Task>
+bool TasksManager<Task>::Worker::are_all_idle()
+{
+  bool all_tasks_idle = true;
+  for (Task& task : mTasksFifo) all_tasks_idle = all_tasks_idle && task.is_idle();
+  return all_tasks_idle;
+}
+
+template<class Task>
+TasksManager<Task>::TasksManager(const int nof_workers) 
+: workers(nof_workers) 
+{}
+
+template<class Task>
+bool TasksManager<Task>::get_free_task(Task*& task_slot) {
+  bool has_task = false;
+  do
+  {
+    for (Worker& worker : workers)
+    {
+      has_task = worker.get_free_task(task_slot);
+      if (task_slot != nullptr) break;
+    }
+  } while (task_slot == nullptr);
+  if (has_task) task_slot->set_handled_result();
+  return has_task;
+}
+
+template<class Task>
+void TasksManager<Task>::get_completed_task(Task*& completed_task) {
+  completed_task = nullptr;
+  bool all_tasks_idle = true;
+  do
+  {
+    all_tasks_idle = true;
+    for (Worker& worker : workers)
+    {
+      worker.get_completed_task(completed_task);
+      if (completed_task != nullptr)
+      {
+        completed_task->set_handled_result();
+        return;
+      }
+
+      all_tasks_idle = all_tasks_idle && worker.are_all_idle();
+    }
+  } while (!all_tasks_idle);
+  // No with_taskd tasks were found in the loop - no complete tasks left to be handled
+  completed_task = nullptr;
+}
--- a/icicle_v3/backend/cpu/include/tasks_manager.h
+++ b/icicle_v3/backend/cpu/include/tasks_manager.h
@@ -1,7 +1,8 @@
 #pragma once 
 #include <atomic>
-#include <thread> // TODO check windows support
+#include <thread>
 #include <stdexcept>
+#include <cassert>
 
 #define LOG_TASKS_PER_THREAD 2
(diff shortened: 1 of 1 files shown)


icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,227 @@
...
+#include <stdexcept>
+
+#define LOG_TASKS_PER_THREAD 2
+#define TASKS_PER_THREAD (1 << LOG_TASKS_PER_THREAD)
+#define TASK_IDX_MASK (TASKS_PER_THREAD - 1)
+#define MANAGER_SLEEP_USEC 10
+#define THREAD_SLEEP_USEC 1
+
+template<class Task>
+class TasksManager {
```
mickeyasa, line 13: add explanation about the purpose of this class

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,227 @@
...
+public:
+  // constructor
+  TasksManager(const int nof_workers);
+
+  // Get a task that isn't occupied
+  // return value signals if there is a valid result to be handled before dispatching a new calculation
+  // Blocking functions
+  bool get_free_task(Task*& task_slot);
+  void get_completed_task(Task*& completed_task); // NOTE the user must handle this task's result before asking for new free tasks
```
mickeyasa, line 22: get_completed_task: 
explain what is that function

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,227 @@
...
+
+  protected:
+    enum TaskStatus {idle, set_task, dispatched, working, pending_result}; // TODO CAPS and eTaskStatus
+    std::atomic<TaskStatus> status;
+  private:
+    int* father_fifo_tail;
+};
+
+
+bool TaskBase::is_ready_for_work() {
```
mickeyasa, line 72: you can implement those functions inside the class

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,227 @@
...
+  };
+
+  std::vector<Worker> workers;
+};
+
+class TaskBase {
+  public:
+    TaskBase() : status(idle), father_fifo_tail(nullptr) {}
+    virtual void execute() = 0; // COMMENT should it be private? still needs to be friend of worker
+    void dispatch(); // USER FUNC
```
mickeyasa, line 49: especially user functions need more documentation

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,227 @@
...
+    bool kill;
+  };
+
+  std::vector<Worker> workers;
+};
+
+class TaskBase {
+  public:
+    TaskBase() : status(idle), father_fifo_tail(nullptr) {}
+    virtual void execute() = 0; // COMMENT should it be private? still needs to be friend of worker
```
mickeyasa, line 48: keep it public, remove the comment and add description

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,227 @@
...
+    void dispatch(); // USER FUNC
+
+    inline bool is_ready_for_work(); // TODO friend functions of worker
```
mickeyasa, line 51: descriptions

mickeyasa (commented):
Koren, please add much more descriptions.
Especially for the functionality that the user is aware of

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,309 @@
...
+   */
+  TasksManager(const int nof_workers) : workers(nof_workers) {}
+
+  /**
+   * @brief Get free slot to insert new task to be executed. This is a blocking function - until a free task is found.
+   * @param task_slot - pointer to the internal task to allow the user to edit in the new task.
+   * @return boolean value indication of the current content of the task - true indicates that `task_slot` holds a 
+   * previous result that should be handled by the user before writing the new task.
+   */
+  bool get_free_task(Task*& task_slot);
```
yshekel, line 35: I don't understand. If it's a free task then why do I need to handle a completed task first? 
I would expect to it to fail and so I would have to call the get_completed_task() and handle it. Maybe there you can modify the completed one and redispatch it.

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,309 @@
...
+     * @brief Constructor of `Worker`.
+     * Inits default values for the class's members and launches the thread.
+     */
+    Worker();
+    /**
+     * @brief Destructor of `Worker`.
+     * Signals the thread to terminate and joins it with main. The destructor does not handle existing tasks' results 
+     * and assumes the user have already handled all the results via `TasksManager`'s api.
+     */
+    ~Worker();
```
yshekel, line 62: virtual destructor

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,309 @@
...
+    void run();
+
+    /**
+     * @brief Get free slot to insert new task to be executed. This isn't a blocking function - it checks all worker's 
+     * tasks and if no free one is found a nullptr is returned.
+     * @param task_slot - pointer to the internal task of the worker to allow the user to edit in the new task.
+     * @return boolean value indication of the current content of the task - true indicates that `task_slot` holds a 
+     * previous result that should be handled by the user before writing the new task.
+     */
+    bool get_free_task(Task*& task_slot);
```
yshekel, line 78: same comment that I wrote above here again

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,309 @@
...
+     * all worker's tasks and if no completed one is found a nullptr is returned.
+     * @param completed_task - pointer to the internal task holding a previous result. 
+     * Again if using completed_task to assign additional tasks, the existing result must be handled before hand.
+     */
+    void get_completed_task(Task*& completed_task);
+
+    /**
+     * @brief Check if all worker's tasks are idle (free with no result to be handled)
+     */
+    bool are_all_idle();
```
yshekel, line 91: is_idle() instead?

Koren-Brand: "is" referring to idle? because are_all refers to the tasks of the worker

icicle_v3/backend/cpu/include/tasks_manager.h
mickeyasa: the behavior of the function is correct. i,e, this is what needed.
the name of the function can change to be more clear e.g. get_free_or_completed_task

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,309 @@
...
+  inline bool has_result();
+  inline bool is_idle();
+  inline void set_working();
+  inline void set_pending_result();
+  inline void set_handled_result();
+
+  /**
+   * @brief link this task with the tail idx of the Worker it belongs to. Used only once after construction.
+   */
+  inline void link_father_tail(int* tail_pointer) { father_fifo_tail = tail_pointer; }
```
yshekel, line 138: this seems wrong to me. You should avoid this coupling with the fifo class because you couple the Task to a fifo implementation now but task doesn't need this constraint.

icicle_v3/backend/cpu/include/tasks_manager.h
Koren-Brand: the code marked the task returned as if its result has been handled for it to not give it to the user again in the case of 2 consecutive calls to get_free_task
The main issue is for completed task but free task is implemented similarly

I can add another state - HAS_RESULT_BUT_ALREADY_SENT_TO_USER but it is a bit cumbersome

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,309 @@
...
+
+  /**
+   * @brief Signal for the `Worker` owning the task that it is ready to be executed. 
+   * Also increase's the workers tail idx.
+   * User function.
+   */
+  void dispatch();
+
+  // Getters and setter for the various states of the tasks `status`.
+  inline bool is_ready_for_work(); // DISPATCHED
```
yshekel, line 127: why not is_dispatched() then? ready for work is unclear.

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,309 @@
...
+  inline bool can_push_task(); // IDLE or PENDING_RESULT
```
yshekel, line 128: consider removing it and only have is_idle() and has_result()

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,309 @@
...
+  };
+
+  std::vector<Worker> workers; // Vector of workers/thread to be ran simultaneously.
+};
+
+/**
+ * @class TaskBase
+ * @brief abstract base for the task supported by TasksManager above.
+ */
+class TaskBase {
```
yshekel, line 107: It's more kind of a TaskEntry really. No?

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,309 @@
...
+bool TaskBase::is_idle() {
+  return status.load(std::memory_order_acquire) == IDLE;
+}
+
+void TaskBase::dispatch() {
+  (*father_fifo_tail)++;
+  status.store(DISPATCHED, std::memory_order_release);
+}
+
+void TaskBase::set_working() {
```
yshekel, line 177: don't you check valid state transitions? Don't assume only happy path

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,309 @@
...
+
+template<class Task>
+TasksManager<Task>::Worker::Worker() 
+: mTasksFifo(TASKS_PER_THREAD), 
+  tail(0),
+  head(0),
+  kill(false)
+  {
+    // Tail idx linking can only be done after task and worker initialization
+    for (Task& task : mTasksFifo) task.link_father_tail(&tail);
```
yshekel, line 203: not good

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,309 @@
...
+    for (head = 0; head < mTasksFifo.size(); head++)
+    {      
+      Task* task = &mTasksFifo[head];
+      if (!task->is_ready_for_work()) {
+        // Sleep as the thread apparently isn't fully utilized currently
+        std::this_thread::sleep_for(std::chrono::microseconds(THREAD_SLEEP_USEC));       
+        continue;
+      }
+      task->set_working();
+      task->execute();
```
yshekel, line 226: have the execute call set_working(). It's too easy to mess it up like that.

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,309 @@
...
+
+template<class Task>
+void TasksManager<Task>::Worker::run() {
+  while (true) {
+    for (head = 0; head < mTasksFifo.size(); head++)
+    {      
+      Task* task = &mTasksFifo[head];
+      if (!task->is_ready_for_work()) {
+        // Sleep as the thread apparently isn't fully utilized currently
+        std::this_thread::sleep_for(std::chrono::microseconds(THREAD_SLEEP_USEC));
```
yshekel, line 222: why not sleep after you check all fifo rather than after every check?

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,309 @@
...
+  return all_tasks_idle;
+}
+
+template<class Task>
+bool TasksManager<Task>::get_free_task(Task*& task_slot) {
+  bool has_task = false;
+  do
+  {
+    for (Worker& worker : workers)
+    {
```
yshekel, line 279: I think you should have the sleep here and then quickly scan all workers instead of sleeping there

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,309 @@
...
+      has_task = worker.get_free_task(task_slot);
```
yshekel, line 280: are you sure everything is synced correctly? with the atomic variables I mean

icicle_v3/backend/cpu/include/tasks_manager.h
yshekel: referring to the worker. Just a rename

icicle_v3/backend/cpu/include/tasks_manager.h
Koren-Brand: will change

icicle_v3/backend/cpu/include/tasks_manager.h
Koren-Brand: I had asserts but removed them
Should I re-add them? Or throw a sort of Icicle error instead?

(all the status functions should only be accessed by the worker anyway - maybe changing to private + friend would be better)

icicle_v3/backend/cpu/include/tasks_manager.h
Koren-Brand: Why?

icicle_v3/backend/cpu/include/tasks_manager.h
Koren-Brand: the issue is that execute is a virtual function will be defined by the user
so it (and set_pending_result) is called in run to avoid cases where the user forgets to update the status and the code is stuck

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,299 @@
...
+  {
+    // Init thread only after finishing all other setup to avoid data races
+    task_executor = std::thread(&TasksManager<Task>::Worker::worker_loop, this);
+  }
+
+template<class Task>
+TasksManager<Task>::Worker::~Worker() {
+  kill = true;
+  task_executor.join();
+  task_executor.termin
```
LeonHibnik, line 184: typo here? shouldn't it be `task_executor.terminate();`?

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,299 @@
...
+}
+
+template<class Task>
+void TasksManager<Task>::Worker::worker_loop() {
+  while (true) {
```
LeonHibnik, line 189: Why not use `while (!kill) {` and remove the condition below?

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,299 @@
+#pragma once 
+#include <atomic>
+#include <thread>
+#include <stdexcept>
+#include <cassert>
+
+#define LOG_TASKS_PER_THREAD 2
```
LeonHibnik, line 7: Shouldn't we set that based on cpu arch? for future optimization

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,298 @@
...
+  while (true)
+  {
+    for (int i = 0; i < m_workers.size(); i++)
+    {
+      m_next_worker_idx = (m_next_worker_idx < m_workers.size() - 1)? m_next_worker_idx + 1 : 0;
+
+      task = m_workers[m_next_worker_idx].get_idle_or_completed_task();
+      if (task != nullptr) { return task; }
+    }
+    // std::this_thread::sleep_for(std::chrono::microseconds(MANAGER_SLEEP_USEC));
```
yshekel, line 267: remove comment if not required

icicle_v3/backend/cpu/include/tasks_manager.h
```diff
@@ -0,0 +1,298 @@
...
+    // Flag sent by reference to get_completed_task below - if a task that isn't idle is found the flag is set to false
+    all_idle = true; 
+    for (int i = 0; i < m_workers.size(); i++)
+    {
+      m_next_worker_idx = (m_next_worker_idx < m_workers.size() - 1)? m_next_worker_idx + 1 : 0;
+
+      completed_task = m_workers[m_next_worker_idx].get_completed_task(all_idle);
+      if (completed_task != nullptr) { return completed_task; }
+    }
+    // std::this_thread::sleep_for(std::chrono::microseconds(MANAGER_SLEEP_USEC));
```
yshekel, line 286: here too

yshekel (commented):
Looks better. Please address last comments and we can merge

icicle_v3/backend/cpu/include/tasks_manager.h
Koren-Brand: it's not the number of threads but the size of the threads' fifo. It's currently set heuristically to reduce the idle time of worker threads.

yshekel (approved):

PR #574, opened 2024-08-07
Cpu basic math optimization

HadarIngonyama:
This PR makes the CPU use 64-bit operations instead of 32-bit operations


icicle_v3/include/icicle/fields/host_math.h
```diff
@@ -112,18 +174,52 @@ namespace host_math {
...
+      (NLIMBS_A % 2 == 0 || NLIMBS_A == 1) && (NLIMBS_B % 2 == 0 || NLIMBS_B == 1),
+      "odd number of limbs is not supported\n");
+    if constexpr (USE_32) {
+      multiply_raw_32<NLIMBS_A, NLIMBS_B>(as, bs, rs);
+      return;
+    } else if constexpr ((NLIMBS_A == 1 && NLIMBS_B == 2) || (NLIMBS_A == 2 && NLIMBS_B == 1)) {
+      multiply_raw_32<NLIMBS_A, NLIMBS_B>(as, bs, rs);
+      return;
+    } else if constexpr (NLIMBS_A == 1 && NLIMBS_B == 1) {
+      rs.limbs[0] = host_math::madc_cc(as.limbs[0], bs.limbs[0], 0, rs.limbs[1]);
```
yshekel, line 212: (1) this madc_cc is doing a redundant +0. right?
(2) if rs.limbs[1] is not zero, it is contributing to the result as carry in. Is that a bug?

Same for madc_cc_64

HadarIngonyama: fixed

PR #580, opened 2024-08-18
Aviad/blake2s cpu

aviadingo:
## Describe the changes

1. Added template Hasher class.
2. Added blake2s cpu implementation which inherits from the general-purpose hasher and overrides the run_hash function.
3. Hash many is currently a dummy implementation runing the hashes in a loop
4. Included tests for the batched version and the single version using input vectors with variable length.

### DIFF
--- a/icicle/backend/cpu/include/hash/blake2/cpu_blake2s.h
+++ b/icicle/backend/cpu/include/hash/blake2/cpu_blake2s.h
@@ -9,10 +9,10 @@
 
 namespace icicle {
 
-  class Blake2s_cpu : public HashBackend
+  class Blake2sCpu : public HashBackend
   {
   public:
-    explicit Blake2s_cpu(int total_input_limbs) : HashBackend(total_input_limbs, BLAKE2S_OUTBYTES / sizeof(limb_t)) {}
+    explicit Blake2sCpu(int total_input_limbs) : HashBackend(total_input_limbs, BLAKE2S_OUTBYTES / sizeof(limb_t)) {}
 
     eIcicleError hash_single(
       const limb_t* input_limbs,
--- a/icicle/backend/cpu/include/hash/keccak/cpu_keccak.h
+++ b/icicle/backend/cpu/include/hash/keccak/cpu_keccak.h
@@ -13,7 +13,7 @@ namespace icicle {
 #define SHA3_TRACE_BUF(format, buf, l)
 
 /*
- * This flag is used to configure "pure" Keccak_cpu, as opposed to NIST SHA3.
+ * This flag is used to configure "pure" KeccakCpu, as opposed to NIST SHA3.
  */
 #define SHA3_USE_KECCAK_FLAG 0x80000000
 #define SHA3_CW(x)           ((x) & (~SHA3_USE_KECCAK_FLAG))
@@ -36,11 +36,11 @@ namespace icicle {
   enum SHA3_RETURN { SHA3_RETURN_OK = 0, SHA3_RETURN_BAD_PARAMS = 1 };
   typedef enum SHA3_RETURN sha3_return_t;
 
-  class Keccak_cpu : public HashBackend
+  class KeccakCpu : public HashBackend
   {
   public:
     // Constructor
-    explicit Keccak_cpu(int total_input_limbs, int total_output_limbs, enum SHA3_FLAGS flag)
+    explicit KeccakCpu(int total_input_limbs, int total_output_limbs, enum SHA3_FLAGS flag)
         : HashBackend(total_input_limbs, total_output_limbs), sha_flag(flag)
     {
     }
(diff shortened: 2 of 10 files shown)


icicle_v3/backend/cpu/include/hash/blake2/blake2s.h
```diff
@@ -0,0 +1,117 @@
+#ifndef BLAKE2_H
+#define BLAKE2_H
+
+typedef unsigned char BYTE;
```
ChickenLover, line 4: I think you should put these typedefs into a namespace or remove them, as this can clash with other names in a global namespace

icicle/include/icicle/hash.h
```diff
@@ -0,0 +1,70 @@
...
+     * @brief Pure virtual function to run a single hash.
+     * @param input_limbs Pointer to the input limbs.
+     * @param output_limbs Pointer to the output limbs.
+     * @return Error code of type eIcicleError.
+     */
+    virtual eIcicleError run_single_hash(
+      const limb_t* input_limbs,
+      limb_t* output_limbs,
+      const HashConfig& config,
+      const limb_t* secondery_input_limbs = nullptr) const = 0;
```
ChickenLover, line 53: secondery -> secondary

icicle/include/icicle/hash.h
```diff
@@ -0,0 +1,70 @@
...
+  };
+
+  /**
+   * @brief Abstract class representing a hash function. In order to support structures like mmcs, we support
+   * side-inputs to the hash. These are inputs that are inserted in an intermediate stage of the calculations.
+   */
+  class Hash
+  {
+  public:
+    const int m_total_input_limbs;           // Total number of regular input limbs
```
ChickenLover, line 32: These names are a little confusing. In hashes it is usually called `rate`. So maybe smth like `input_rate` and `output_rate` would be better? Also, what does `m` stand for here?

icicle/include/icicle/hash.h
ChickenLover, line 53: What does it mean to have secondary inputs in a single hash?

icicle/backend/cpu/include/hash/blake2/blake2s.h
```diff
@@ -0,0 +1,116 @@
+
+#pragma once
+#include <stddef.h>
+#include <stdint.h>
+#include <string.h>
+#include "icicle/hash.h"
+
+using namespace icicle;
+namespace blake2s_cpu {
```
ChickenLover, line 9: All of Icicle types should be defined inside `icicle` namespace. So you need to have smth like

```
namespace icicle {
  namespace blake2s_cpu {
     ...
  }
}
```

But even so, I gave it a second thought on these BYTE, WORD, LONG. It's not worth for the users to use blakes2s through `icicle::blake2s_cpu::Blake2s` just because of these 3 typedefs. Maybe it's worth to just remove them. Another solution would be to move these typedefs into some namespace in utils.h or smth and then import them to your code wherever needed

e. g.
```
// utils.h
namespace icicle {
  namespace asm_types {
    typedef unsigned char BYTE;
    typedef unsigned int WORD;
    typedef unsigned long long LONG;
  }
}

// blake2s.h
namespace icicle {
  using namespace asm_types;
}
```

icicle/backend/cpu/include/hash/blake2/blake2s.h
```diff
@@ -0,0 +1,116 @@
...
+#if defined(_MSC_VER)
+#define BLAKE2_INLINE __inline
+#elif defined(__GNUC__)
+#define BLAKE2_INLINE __inline__
+#else
+#define BLAKE2_INLINE
+#endif
+#else
+#define BLAKE2_INLINE inline
+#endif
```
ChickenLover, line 31: This looks very foolproof :) but all of our modifiers are defined inside utils/modifiers.h. So I think you should either remove it or move to modifiers.h, or use one of already defined ones

icicle/backend/cpu/include/hash/blake2/blake2s.h
```diff
@@ -0,0 +1,116 @@
...
+    run_single_hash(const limb_t* input_limbs, limb_t* output_limbs, const HashConfig& config, const limb_t* secondery_input_limbs = nullptr) const override;
+    eIcicleError run_multiple_hash(
+      const limb_t* input_limbs,
+      limb_t* output_limbs,
+      int nof_hashes,
+      const HashConfig& config,
+      const limb_t* secondery_input_limbs = nullptr) const override;
+
+  private:
+    enum blake2s_constant {
```
ChickenLover, line 48: Why is that an enum? You can just do

```
unsigned int BLAKE2S_BLOCKBYTES = 64;
...
```

icicle/backend/cpu/include/hash/keccak/keccak.h
```diff
@@ -0,0 +1,124 @@
...
+
+  class Keccak : public Hash
+  {
+  public:
+    // Constructor
+    explicit Keccak(int total_input_limbs, int total_output_limbs, enum SHA3_FLAGS flag)
+        : Hash(total_input_limbs, total_output_limbs), sha_flag(flag)
+    {
+    }
+    const size_t bitSize = this->m_total_output_limbs * sizeof(limb_t) * 8;
```
ChickenLover, line 47: we agreed to use snake case everywhere

ChickenLover:
@aviadingo there is no more `yshekel/V3` afaik. I guess you need to rebase to main

ChickenLover:
@aviadingo is this still relevant? can we close it?

PR #581, opened 2024-08-19
Koren/v3 msm

Koren-Brand:
## Describe the changes

MSM multithreaded implementation on CPU

## Linked Issues

Resolves #


icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,667 @@
...
+#include "tasks_manager.h"
+
+using namespace icicle;
+
+#ifdef DUMMY_TYPES // for testing
+using A = DummyP;
+using P = DummyP;
+using scalar_t = DummyScalar;
+#else
+using A = curve_config::affine_t;
```
yshekel, line 24: it now seems confusing that both those and the templates use A,P. Also, it doesn't support G2. 
Now that everything is template I think you can move this logic to the test. It would be better than here.

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,667 @@
...
+ */
+template <typename A, typename P>
+class EcAddTask : public TaskBase
+{
+public:
+  /**
+   * @brief constructor for the task ensuring Ps are zeros and other fields are init to invalid values (to avoid
+   * falsely handling a result at the start). The execution adds Ps m_p1,m_p2 and stores the result in m_p1.
+   */
+  EcAddTask() : TaskBase(), m_p1(P::zero()), m_p2(P::zero()), m_return_idx(-1) {}
```
yshekel, line 42: use clearer names than m_p1, m_p2

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,667 @@
...
+
+  /**
+   * @brief Function to be executed by the tasks manager. It can be configured according to the value odf
+   */
+  void execute()
+  {
+    switch (m_p2_config) {
```
yshekel, line 49: (1) use better names for the enum values. what is PER?
(2) ADD_P1_AND_AFFINE_P1 adds p1 and p2?
(3) I don't get the name  'm_p2_config'

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,667 @@
...
+
+  /**
+   * @brief Set up phase 1 addition between a new base and the target bucket.
+   * @param bucket - bucket to be added. It isn't passed as Per as the bucket's value can be changed during
+   * addition execution.
+   * @param base - affine base to be added. Passed as a Per as the input is constant and it saves copying.
+   * @param negate_affine - flag to indicate that the base needs to be subbed instead of added.
+   * @param is_montgomery - flag to indicate that the base is in Montgomery form and first needs to be converted.
+   */
+  void set_phase1_addition_with_affine(const P& bucket, const A base, int bucket_idx)
```
yshekel, line 70: why base passed by value?

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,667 @@
...
+   * @param bucket - the bucket value to be added to the existing result in p1.
+   */
+  void set_phase1_collision_task(P& bucket)
+  {
+    m_p2 = bucket;
+    m_p2_config = ADD_P1_P2_BY_VALUE;
+    dispatch();
+  }
+
+  P m_p1;                 // One of the addends, and holds the addition result afterwards
```
yshekel, line 139: addends? is that a word?

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,667 @@
...
+   */
+  Msm(const MSMConfig& config);
+
+  /**
+   * @brief Destructor for Msm class.
+   * Ensures phase 3 threads have finished and joins them (other deletion are implemented implicitly).
+   */
+  ~Msm()
+  {
+    for (std::thread& p3_thread : m_p3_threads) {
```
yshekel, line 176: if p is phase then use the word phase. It makes code hard to read when have to guess what letters stand for. 
Even if longer, it is easier and shorter to read.

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,667 @@
...
+  /**
+   * @brief Push addition task during phase 1.
+   * Th function also handles completed addition results while attempting to insert the new addition task (including
+   * potential collision that requires more urgent ec-addition of the result and the current stored value).
+   * @param task_bkt_idx - address in m_buckets in which to store the result in the future.
+   * @param bkt - the P from m_buckets.
+   * @param base - the P from the input bases
+   * @param negate_base - flag to signal the task to subtract base instead of adding it.
+   */
+  void phase1_push_addition(const unsigned int task_bkt_idx, const P bkt, const A base);
```
yshekel, line 251: just a note: you don't really need the const here. What does it give you in terms of api?

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,667 @@
...
+   * The function also handles the potential collision similarly to push_addition above.
+   */
+  void phase1_wait_for_completion();
+
+  /**
+   * @brief Phase 2 of MSM. Function handling the initial parallel part of summing the bucket modules. It splits the
+   * BMs into segments, summing each separately and passing the segments sum to be handled by the final accumulator.
+   * @return vector containing segments line and triangle sums for the final accumulator.
+   */
+  void bm_sum(std::shared_ptr<std::vector<BmSumSegment>>& segments_ptr);
```
yshekel, line 263: (1) why pass shared_ptr by reference?
(2) why do you even need a shared_ptr here? you could just have the vector and pass it by reference instead of passing the shared_ptr by reference.

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,667 @@
...
+  /**
+   * @brief Final accumulation required for MSM calculation. the function will either launch a thread to perform the
+   * calculation (`phase3_tread` function) or by the main thread, depending on the position in the batch (Last msm or
+   * not).
+   * @param segments_ptr - Per to vector containing the segment calculations of phase 2
+   * @param idx_in_batch - idx of the current MSM in the batch.
+   * @param result - output, Per to write the MSM result to. Memory for the Per has already been allocated by
+   * the user.
+   */
+  void final_accumulator(std::shared_ptr<std::vector<BmSumSegment>>& segments_ptr, int idx_in_batch, P* result);
```
yshekel, line 281: same comment here

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,667 @@
...
+
+  /**
+   * @brief Phase 3 function to be ran by a separate thread (or main in the end of the run) - i.e. without using the
+   * tasks manager. It performs the remaining serial addition in each BM and sums them to one final MSM result.
+   * @param result - Per to write the MSM result to. Memory for the Per has already been allocated by the user.
+   */
+  void phase3_thread(std::shared_ptr<std::vector<BmSumSegment>> segments_ptr, P* result);
```
yshekel, line 288: This makes more sense but do you really need the shared_ptr or can simply pass vector reference or move it.

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,667 @@
...
+  /**
+   * @brief Function for resetting class members between batch runs.
+   */
+  void batch_run_reset() { std::fill(m_bkts_occupancy.begin(), m_bkts_occupancy.end(), false); }
+};
+
+template <typename A, typename P>
+Msm<A, P>::Msm(const MSMConfig& config)
+    : manager(
+        config.ext->get<int>("n_threads") > 0 ? config.ext->get<int>("n_threads")
```
yshekel, line 299: (1) note that ext may be null. You must check it before call get.
(2) add the n_threads to https://github.com/ingonyama-zk/icicle/blob/yshekel/V3/icicle_v3/include/icicle/backend/msm_config.h 
(3) use the value from (2) rather than hard coded string.

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,667 @@
...
+      m_segment_size(std::max((int)(m_num_bkts >> m_log_num_segments), 1)),
+
+      m_p3_threads(m_batch_size - 1)
+{
+}
+
+template <typename A, typename P>
+void Msm<A, P>::run_msm(
+  const scalar_t* scalars, const A* bases, const unsigned int msm_size, const unsigned int batch_idx, P* results)
+{
```
yshekel, line 323: Those are the phases? please add comment here explaining the top level logic of the algorithm

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,667 @@
...
+
+// None class functions below:
+
+/**
+ * @brief Function to check the MSM config is valid for calculating in CPU.
+ * @return - status if the config is supported or not.
+ */
+eIcicleError not_supported(const MSMConfig& conf)
+{
+  if (conf.is_async) { return eIcicleError::INVALID_DEVICE; }
```
yshekel, line 593: why? just be sync. It's not really an error.

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,667 @@
...
+  // There is only host (CPU) therefore the following configs are not supported:
+  if (conf.are_scalars_on_device | conf.are_points_on_device | conf.are_results_on_device) {
```
yshekel, line 595: again, why do you care about that? for CPU, device and host memory is the same. So what?

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,667 @@
...
+{
+  Msm<A, P>* msm = new Msm<A, P>(config);
+
+  if (not_supported(config) != eIcicleError::SUCCESS) return not_supported(config);
+
+  const unsigned int c = config.c;
+  const unsigned int precompute_factor = config.precompute_factor;
+  const int num_bms = ((scalar_t::NBITS - 1) / (precompute_factor * c)) + 1;
+
+  if (config.ext->get<int>("n_threads") <= 0) { return eIcicleError::INVALID_ARGUMENT; }
```
yshekel, line 623: must check that ext is not null

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,667 @@
...
+  const MSMConfig& config,
+  A* output_bases) // Pre assigned?
+{
+  int precompute_factor = config.precompute_factor;
+  bool is_mont = config.are_points_montgomery_form;
+  // bool is_mont=false;
+  const unsigned int c = config.c;
+  const unsigned int num_bms_no_precomp = (scalar_t::NBITS - 1) / c + 1;
+  const unsigned int shift = c * ((num_bms_no_precomp - 1) / precompute_factor + 1);
+  for (int i = 0; i < nof_bases; i++) {
```
yshekel, line 656: is that not going to be parallel? Do you know approximately how long it takes for typical cases? It's done once but still if slow can be an issue.

icicle_v3/tests/test_curve_api.cpp
```diff
@@ -86,18 +91,34 @@ class CurveApiTest : public ::testing::Test
...
+      // Note: allocating the precompute_bases on device since CUDA backend assumes that.
+      // TODO: fix CUDA backend to support host memory too.
+      A* precomp_bases = nullptr;
+      ICICLE_CHECK(icicle_malloc((void**)&precomp_bases, N * precompute_factor * sizeof(A)));
+      ICICLE_CHECK(msm_precompute_bases(bases.get(), N, config, precomp_bases));
+
       START_TIMER(MSM_sync)
       for (int i = 0; i < iters; ++i) {
-        msm(scalars.get(), bases.get(), N, config, result);
+        ICICLE_CHECK(msm(scalars.get(), precomp_bases, N, config, result));
```
yshekel, line 111: need to set config.bases_on_device=true since icicle_malloc() allocates on the device.

wrappers/rust_v3/icicle-core/src/msm/tests.rs
```diff
@@ -182,14 +179,12 @@ pub fn check_msm_skewed_distributions<C: Curve + MSM<C>>()
...
-                                // let test_threshold = 1 << 11; // TODO - uncomment this line after implementing fast msm
-                                // let batch_sizes = [1, 3, 1 << 4]; // TODO - uncomment this line after implementing fast msm
-    let batch_sizes = [1, 3]; // TODO - remove this line after implementing fast msm
+    let test_sizes = [1 << 10, 10000];
+    let test_threshold = 1 << 11;
+    let batch_sizes = [1, 3, 1 << 4];
     let rng = &mut thread_rng();
     for test_size in test_sizes {
-        let test_threshold = test_size >> 2; // TODO - remove this line after implementing fast msm
+        let test_threshold = test_size >> 2;
```
yshekel, line 187: need to remove this line

yshekel (changes requested):
Overall looks good but I had some comments.
Please ask Miki/Hadar to review the algorithm part. I did not really review it.

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
Koren-Brand: Passed by value due to intermediate calculations done in bucket_accumulator which are stored in a local variable - meaning there is no valid address to be passed 
there are todos to make the intermediate calcuations as preprocessing which would also allow to pass the affine point by reference, but it will be added in a future update

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
Koren-Brand: 1. typo
2. adds p1 and affine p2
3. it's to config which p2 will be added

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
Koren-Brand: yes the input element in an addition are called addends

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
Koren-Brand: 1+2 handled

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
Koren-Brand: Using segments_ptr.get()?

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
Koren-Brand: A smart pointer is required to allow passing the vector from phase 2 functions on the main thread to the phase 3 function on a separate thread without the memory being deallocated (the vector is not a member of the class - it is defined for each phase 2 during MSM run)

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,667 @@
...
+      m_p3_threads(m_batch_size - 1)
+{
+}
+
+template <typename A, typename P>
+void Msm<A, P>::run_msm(
+  const scalar_t* scalars, const A* bases, const unsigned int msm_size, const unsigned int batch_idx, P* results)
+{
+  bucket_accumulator(scalars, bases, msm_size);
+  auto segments = std::make_shared<std::vector<BmSumSegment>>(m_num_bms * m_num_bm_segments);
```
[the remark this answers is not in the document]
Koren-Brand: Without a smart pointer, the pointer seems to have been freed at the end of the run_msm function which doesn't necessarily finish before the phase 3 function on a separate thread

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
Koren-Brand: removed from the function

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
Koren-Brand: removed from the function

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
Koren-Brand: handled

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
Koren-Brand: Currently not
It can be added in the future (Maybe with the preprocess for phase 1 mentioned above)

icicle_v3/tests/test_curve_api.cpp
Koren-Brand: Added the proper config

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
Koren-Brand: done

icicle_v3/backend/cpu/include/timer.cpp
```diff
@@ -0,0 +1,34 @@
+#include <iostream>
```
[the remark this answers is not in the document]
Koren-Brand: will do

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
yshekel: about (2), I meant the typo

icicle/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,663 @@
...
+      task->set_idle();
+    }
+    task = manager.get_completed_task();
+  }
+}
+
+template <typename A, typename P>
+void Msm<A, P>::phase2_bm_sum(std::vector<BmSumSegment>& segments)
+{
+  phase2_setup(segments);
```
HadarIngonyama, line 426: as we discussed this is probably extraneous

icicle/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,663 @@
...
+  if (batch_idx < m_batch_size - 1) { batch_run_reset(); }
+}
+
+template <typename A, typename P>
+void Msm<A, P>::phase1_bucket_accumulator(const scalar_t* scalars, const A* bases, const unsigned int msm_size)
+{
+  const int coeff_bit_mask_no_sign_bit = m_num_bkts - 1;
+  const int coeff_bit_mask_with_sign_bit = (1 << m_c) - 1;
+  // NUmber of windows / additions per scalar in case num_bms * precompute_factor exceed scalar width
+  const int num_additions_per_scalar = ((scalar_t::NBITS - 1) / m_c) + 1; // +1 for ceiling
```
HadarIngonyama, line 329: nof windows/nof original bms

icicle/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,663 @@
...
+        curr_segment.m_idx_in_segment--;
+
+        // Line sum (if not the last one in the segment)
+        if (curr_segment.m_idx_in_segment >= 0) {
+          int bkt_idx = curr_segment.m_segment_mem_start + curr_segment.m_idx_in_segment;
+          if (m_bkts_occupancy[bkt_idx]) {
+            int return_idx = task->m_return_idx;
+            // Due to the choice of num segments being less than half of total tasks there ought to be an idle task for
+            // the line sum
+            task = manager.get_idle_task();
```
HadarIngonyama, line 471: as we discussed, use get idle for both tasks or assign thread per segment in advance

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,657 @@
...
+ *                  results array.
+ */
+template <typename A, typename P>
+eIcicleError cpu_msm(
+  const Device& device, const scalar_t* scalars, const A* bases, int msm_size, const MSMConfig& config, P* results)
+{
+  int c = config.c;
+  if (c < 1) { c = std::max((int)std::log2(msm_size) - 1, 8); }
+  if (scalar_t::NBITS % c == 0) {
+    std::cerr << "Currerntly c (" << c << ") mustn't divide scalar width (" << scalar_t::NBITS
```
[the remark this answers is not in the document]
Koren-Brand: It does I currently have a bug  regarding this 
fixing it is on the todo list

icicle_v3/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -0,0 +1,657 @@
...
+      m_num_bm_segments(std::min((int)(1 << m_log_num_segments), (int)(m_num_bms * m_num_bkts))),
+      m_segment_size(std::max((int)(m_num_bkts >> m_log_num_segments), 1)),
+
+      m_phase3_threads(m_batch_size - 1)
+{
+}
+
+template <typename A, typename P>
+void Msm<A, P>::run_msm(
+  const scalar_t* scalars, const A* bases, const unsigned int msm_size, const unsigned int batch_idx, P* results)
```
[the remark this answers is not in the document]
Koren-Brand: @yshekel is it ok? (given the existing api)

icicle/backend/cpu/src/curve/cpu_msm.hpp
Koren-Brand: will be dealt within  a future version

yshekel (approved):

yshekel (commented):
note the rust msm is not right. You need to handle the TODOs in msm test in rust and remove the test in v3 dir since it was renamed

PR #608, opened 2024-09-04
Hashing and merkle tree API

yshekel:


icicle/include/icicle/hash/hash.h
```diff
@@ -0,0 +1,91 @@
...
+      const limb_t* secondary_input_limbs = nullptr) const
+    {
+      return m_backend->hash_many(input_limbs, output_limbs, nof_hashes, config, secondary_input_limbs);
+    }
+
+    /**
+     * @brief Get the total number of input limbs.
+     * @return The total number of input limbs as uint64_t.
+     */
+    uint64_t total_input_limbs() const { return m_backend->total_input_limbs(); }
```
mickeyasa, line 73: inline

icicle/include/icicle/hash/hash.h
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * @brief Get the total number of output limbs.
+     * @return The total number of output limbs as uint64_t.
+     */
+    uint64_t total_output_limbs() const { return m_backend->total_output_limbs(); }
```
mickeyasa, line 79: all functions should be inline

icicle/include/icicle/hash/hash_config.h
```diff
@@ -0,0 +1,36 @@
...
+  struct HashConfig {
+    icicleStreamHandle stream;  /**< Stream for asynchronous execution. */
+    bool are_inputs_on_device;  ///< True if inputs are on the device, false if on the host. Default is false.
+    bool are_outputs_on_device; ///< True if outputs are on the device, false if on the host. Default is false.
+    bool is_async; ///< True to run the hash asynchronously, false to run it synchronously. Default is false.
+    ConfigExtension* ext =
+      nullptr; ///< Backend-specific extensions. This allows customization for specific device backends.
+  };
+
+  static HashConfig default_hash_config()
```
mickeyasa, line 24: why do we need this functions?
Why not implement it at HashConfig Constructor?

icicle/include/icicle/hash/keccak.h
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * @brief Creates a Keccak-256 hash object.
+   *
+   * This function constructs a Hash object configured for Keccak-256, with the
+   * appropriate backend selected based on the current device.
+   *
+   * @param total_input_limbs Number of input limbs for the Keccak-256 hash.
+   * @return Hash object encapsulating the Keccak-256 backend.
+   */
+  Hash create_keccak_256_hash(uint64_t total_input_limbs);
```
mickeyasa, line 16: why functions and not constructors?
This commant apply to the whole file

icicle/include/icicle/hash/merkle_tree_config.h
```diff
@@ -0,0 +1,47 @@
...
+
+  /**
+   * @brief Generates a default configuration for the Merkle tree.
+   *
+   * This function provides a default configuration for Merkle tree operations where the inputs, results,
+   * and paths are assumed to be on the host (CPU), and the execution is synchronous by default.
+   *
+   * @return A default MerkleTreeConfig with synchronous execution and host-based data handling.
+   */
+  static MerkleTreeConfig default_merkle_tree_config()
```
mickeyasa, line 35: why do we need that functions?
Set them on the constructor as default

icicle/include/icicle/common.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include <cstdint>
+
+namespace icicle {
+
+  // Define limb type for hashing
+  using limb_t = uint32_t;
```
mickeyasa, line 8: this should be taken rom Field.

icicle/include/icicle/device.h
```diff
@@ -1,6 +1,8 @@
 #pragma once
 
 #include <cstring>
+#include <string>
```
mickeyasa, line 4: did you change that for Hash

icicle/include/icicle/hash/merkle_tree.h
```diff
@@ -0,0 +1,73 @@
...
+    {
+      return m_backend->build(leaves, config, secondary_leaves);
+    }
+
+    /**
+     * @brief Get the root of the Merkle tree.
+     * @param root Output parameter for the root of the tree.
+     * @return Error code of type eIcicleError.
+     */
+    eIcicleError get_root(const limb_t*& root) const { return m_backend->get_root(root); }
```
mickeyasa, line 36: all functions should be inline

icicle/include/icicle/hash/merkle_tree.h
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * @brief Get the Merkle path from a specified element index.
+     * @param leaves Pointer to the leaves of the tree.
+     * @param element_idx Index of the element for which the path is to be retrieved.
+     * @param path Output parameter for the Merkle path.
+     * @param config Configuration struct for the Merkle tree.
+     * @return Error code of type eIcicleError.
+     */
+    eIcicleError
+    get_path(const limb_t* leaves, uint64_t element_idx, limb_t* path, const MerkleTreeConfig& config) const
```
mickeyasa, line 47: please add input parameter 
const bool is_pruned

icicle/include/icicle/hash/merkle_tree.h
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * @brief Verify an element against a Merkle path.
+     * @param path Pointer to the Merkle path.
+     * @param element_idx Index of the element to verify.
+     * @param verification_valid Output parameter indicating if the verification is valid.
+     * @param config Configuration struct for the Merkle tree.
+     * @return Error code of type eIcicleError.
+     */
+    eIcicleError
+    verify(const limb_t* path, unsigned int element_idx, bool& verification_valid, const MerkleTreeConfig& config)
```
mickeyasa, line 61: also need an indication if the path is pruned 
If the path is a class, that data can be inside

icicle/include/icicle/hash/merkle_tree.h
```diff
@@ -0,0 +1,73 @@
...
+    {
+      return m_backend->verify(path, element_idx, verification_valid, config);
+    }
+
+  private:
+    std::shared_ptr<MerkleTreeBackend> m_backend; ///< Shared pointer to the Merkle tree backend.
+  };
+
+  MerkleTree create_merkle_tree(
```
mickeyasa, line 70: why not simple constructor?

icicle/tests/test_hash_api.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+TEST_F(HashApiTest, MerkleTree)
+{
+  // const uint64_t nof_leaves_limbs = 100; // Number of input limbs
+  // auto leaves = std::make_unique<uint32_t[]>(nof_leaves_limbs);
+  // // Randomize the input array
+  // randomize(leaves.get(), nof_leaves_limbs);
+
+  auto config = default_merkle_tree_config();
+  auto layer0_hash = create_hash_sum(5, 2); // input 5 limbs, output 2 limbs
```
mickeyasa, line 188: initializer list will work here?

icicle/include/icicle/backend/hash/merkle_tree_backend.h
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    virtual ~MerkleTreeBackend() = default;
+
+    // Pure virtual methods to be implemented by derived classes
+    virtual eIcicleError
+    build(const limb_t* leaves, const MerkleTreeConfig& config, const limb_t* secondary_leaves = nullptr) = 0;
+    virtual eIcicleError get_root(const limb_t*& root) const = 0;
+    virtual eIcicleError
+    get_path(const limb_t* leaves, uint64_t element_idx, limb_t* path, const MerkleTreeConfig& config) const = 0;
```
mickeyasa, line 41: add input
const bool is_pruned

icicle/include/icicle/hash/hash.h
yshekel: will inline everywhere yes

icicle/include/icicle/hash/hash_config.h
yshekel: I am not sure why but this is how the library is everywhere so we better be consistent across APIs.
We can have constructors instead, or in addition maybe so that we don't break exiting code.

icicle/include/icicle/hash/merkle_tree_config.h
yshekel: same answer as for hash

icicle/include/icicle/common.h
yshekel: it's not defined anywhere currently.
I could define it in field.h  instead but I prefer to avoid it since I don't want to include it everywhere if I don't need it.
I think that field.h should include this file since limb is not restricted to field elements but a more general concept.

icicle/include/icicle/device.h
yshekel: I didn't have the includes which is a mistake. It works if you include them before including device.h but it's still a mistake to rely on previous includes.

icicle/include/icicle/hash/merkle_tree.h
yshekel: I will refactor this path thing as we discussed. Also I think that the verification should be a method of the path object and not the tree since typically the verifier will call it and he doesn't have the tree but only the path and root.
Also I want to go over plonky3 and see their API to make sure this is what users need.

icicle/include/icicle/hash/merkle_tree.h
yshekel: yes will do that

icicle/include/icicle/hash/merkle_tree.h
yshekel: Here I can actually do it indeed. Do you prefer that?

icicle/tests/test_hash_api.cpp
yshekel: what do you mean here? allocating them directly into a vector like
std::vector<Hash> layer_hashes = {create_hash_sum(..), create_hash_sum(..),create_hash_sum(..),}
?
I think yes

icicle/include/icicle/hash/hash_config.h
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * @brief Configuration struct for the hash.
+   *
+   * This struct holds configuration options for hash operations. It allows the user
+   * to specify whether inputs and outputs reside on the device (e.g., GPU) or the host (e.g., CPU),
+   * whether the hash operation should run asynchronously, and any backend-specific extensions.
+   */
+  struct HashConfig {
+    icicleStreamHandle stream;  /**< Stream for asynchronous execution. */
```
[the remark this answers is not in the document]
yshekel: There is no DeviceContext in V3 but you get the device in every backend API so you have it.

icicle/tests/test_hash_api.cpp
```diff
@@ -0,0 +1,215 @@
...
+public:
+  HashSumBackend(uint64_t total_input_limbs, uint64_t total_output_limbs, uint64_t total_secondary_input_limbs = 0)
+      : HashBackend(total_input_limbs, total_output_limbs, total_secondary_input_limbs)
+  {
+  }
+
+  virtual eIcicleError hash_single(
+    const limb_t* input_limbs,
+    limb_t* output_limbs,
+    const HashConfig& config,
```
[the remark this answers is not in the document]
yshekel: If you take it in the constructor you couple your tree to hash only data on device or on host but why would you do that? I don't see an advantage.

icicle/src/hash/keccak.cpp
```diff
@@ -0,0 +1,72 @@
...
+    std::shared_ptr<HashBackend> backend;
+    ICICLE_CHECK(Keccak512Dispatcher::execute(total_input_limbs, backend));
+    Hash keccak{backend};
+    return keccak;
+  }
+
+  /*************************** C API ***************************/
+
+  extern "C" Hash* create_keccak_256_hash_c_api(uint64_t total_input_limbs)
+  {
```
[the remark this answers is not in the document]
yshekel: Yes that's correct. I will have to do it in a later PR when adding rust for the finalized C++ API.

icicle/tests/test_hash_api.cpp
```diff
@@ -0,0 +1,215 @@
...
+  // merkle_tree.get_root(root);
+  // std::cout << "Root = " << *root << std::endl;
+  // limb_t* path;
+  // uint path_size;
+  // uint64_t element_idx = 5;
+  // merkle_tree.allocate_path(path, path_size);
+  // merkle_tree.get_path(leaves, element_idx, path, config);
+  // std::cout << “path(“<< element_idx << “):” << std::endl;
+  // merkle_tree.print_path(path);
+  // std::cout << “verify:” << std::endl;
```
Koren-Brand, line 211: shouldn't these lines be uncommented? otherwise, the test only checks that the tree doesn't crash upon building it

icicle/tests/test_hash_api.cpp
```diff
@@ -0,0 +1,215 @@
...
+{
+  // const uint64_t nof_leaves_limbs = 100; // Number of input limbs
+  // auto leaves = std::make_unique<uint32_t[]>(nof_leaves_limbs);
+  // // Randomize the input array
+  // randomize(leaves.get(), nof_leaves_limbs);
+
+  auto config = default_merkle_tree_config();
+  auto layer0_hash = create_hash_sum(5, 2); // input 5 limbs, output 2 limbs
+  auto layer1_hash = create_hash_sum(6, 3); // input 6 limbs, output 3 limbs
+  auto layer2_hash = create_hash_sum(6, 2); // input 6 limbs, output 2 limbs
```
Koren-Brand, line 190: What about testing hashes with side inputs?

icicle/tests/test_hash_api.cpp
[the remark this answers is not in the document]
yshekel: That's for the user to decide but yes you are right as long as we can reuse the data on GPU. Otherwise it doesn't help much and also sometimes the data is too large for GPU memory.

icicle/tests/test_hash_api.cpp
yshekel: yes it's WIP

icicle/tests/test_hash_api.cpp
yshekel: Don't worry you will add all the tests. I am just doing the API

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
```diff
@@ -0,0 +1,57 @@
...
+    CPUMerkleTreeBackend(
+      const std::vector<Hash>& layer_hashes, uint64_t leaf_element_size_in_limbs, uint64_t output_store_min_layer = 0)
+        : MerkleTreeBackend(layer_hashes, leaf_element_size_in_limbs, output_store_min_layer)
+    {
+    }
+
+    ~CPUMerkleTreeBackend() = default;
+
+    eIcicleError
+    build(const limb_t* leaves, const MerkleTreeConfig& config, const limb_t* secondary_leaves = nullptr) override
```
ChickenLover, line 18: What are `secondary_leaves`?

icicle/tests/test_hash_api.cpp
```diff
@@ -0,0 +1,164 @@
...
+  // ret root and merkle-path to an element
+  uint64_t root = 0; // assuming output is 8B
+  uint64_t element_idx = 5;
+  const auto path_size = merkle_tree.calculate_merkle_path_size();
+  MerklePath merkle_path{path_size};
+  merkle_tree.get_merkle_root(&root);
+  merkle_tree.get_merkle_path(leaves, element_idx, config, merkle_path);
+
+  bool verification_valid = merkle_tree.verify(leaves + element_idx, element_idx, merkle_path, &root, config);
+  // ASSERT_EQ(verification_valid, true);
```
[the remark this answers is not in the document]
yshekel: you are right but right now it will make CI fail until it is implemented, but this is a PR for API only

icicle/tests/test_hash_api.cpp
```diff
@@ -0,0 +1,164 @@
...
+  constexpr int nof_leaves = 100;
+  uint32_t leaves[nof_leaves];
+  const size_t input_size = sizeof(leaves);
+  for (int i = 0; i < nof_leaves; ++i) {
+    leaves[i] = i;
+  }
+
+  // define the merkle tree
+  auto config = default_merkle_tree_config();
+  auto layer0_hash = HashSumBackend::create(20, 8); // input 20 bytes, output 8 bytes input   400B ->  160B
```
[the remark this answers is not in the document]
yshekel: The api with the secondary did not meet the requirements of mmcs and we decided to not support mmcs right now.

icicle/include/icicle/backend/hash/hash_backend.h
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * @brief Get the output size in bytes for a single hash chunk.
+     * @return The output size in bytes.
+     */
+    uint64_t output_size() const { return m_output_size; }
+
+  protected:
+    const uint64_t m_output_size;              ///< The number of output bytes produced by the hash.
+    const uint64_t m_default_input_chunk_size; ///< Expected input chunk size for hashing operations.
```
Koren-Brand, line 63: Why is the terminology different instead of just input_size? I don't understand what chunk implies

icicle/include/icicle/backend/hash/hash_backend.h
```diff
@@ -0,0 +1,74 @@
...
+
+    inline uint64_t get_single_chunk_size(uint64_t size_or_zero) const
```
Koren-Brand, line 65: what is the use of size_or_zero? it seems like the function should just return the default type and the child of the hash should decide whether it needs the default size or use a size variable from someplace else

icicle/include/icicle/backend/hash/hash_backend.h
```diff
@@ -0,0 +1,74 @@
...
+    {
+      auto size = (size_or_zero == 0) ? input_default_chunk_size() : size_or_zero;
+      ICICLE_ASSERT(size > 0) << "Cannot infer hash size. Make sure to pass it to hasher.hash(...size...) or have "
```
Koren-Brand, line 68: This assertion is only valid if input_default_chunk_size has not been set in the constructor (as it is unsigned i.e. >= 0) so shouldn't this assertion be implemented in the getter of input_default_chunk_size above?

icicle/include/icicle/backend/hash/hash_backend.h
yshekel: when you hash you typically get the input size as parameter. Hashes like keccak can accept any input size and output 256B for example.  Hashes such as poseidon have a few variants but per each, the input size is fixed. This is the chunk size.
When building a tree you want to define the chunk size even for a keccak hash (for example 1KB or whatever). Then given the chunk size and the size of the leaves you can pad and compute the hash on the leaves, where each chunk of data (the 1KB) is used to generate one tree node.

icicle/include/icicle/backend/hash/hash_backend.h
yshekel: You can ignore it if you don't think it's useful for the backend but I guess this logic will be implemented everywhere otherwise. In case of tree you need to use the default chunk size but for general hashing this is generally not the case. so if a user (or tree builder) passes size=0 then use the default and otherwise what he speicifed.

icicle/include/icicle/backend/hash/hash_backend.h
yshekel: no, it's fine to not have a default chunk size. For example you want to hash a string of size 189 bytes. Then you can use the hasher again for another data of size 259 bytes for example. In that case there is not default chunk size.
For tree builder though this is important to have a default chunk size so in this case the assertion will fail.

Koren-Brand: I understand that the default is not required.
However, if the user calls input_default_chunk_size() I assume he assumes there is a default size, and in that case, an assertion should be made, shouldn't it?

icicle/include/icicle/backend/hash/hash_backend.h
```diff
@@ -0,0 +1,74 @@
...
+   *
+   */
+  class HashBackend
+  {
+  public:
+    /**
+     * @brief Constructor for the HashBackend class.
+     *
+     * @param output_size The size of the output in bytes.
+     * @param default_input_chunk_size The size of a single input chunk. Useful for Merkle trees.
```
mickeyasa, line 24: in bytes

icicle/include/icicle/backend/hash/hash_backend.h
yshekel: I think that we should avoid assertions. The user can deal with it. No fatal error occured

icicle/include/icicle/backend/hash/hash_backend.h
yshekel: fixing

icicle/include/icicle/backend/hash/hash_backend.h
Koren-Brand: Then it can be removed from here as well doesn't it?

yshekel: in that case the user can handle it I think.

icicle/include/icicle/backend/merkle/merkle_tree_backend.h
```diff
@@ -0,0 +1,126 @@
...
+      uint64_t element_idx,
+      const MerkleTreeConfig& config,
+      MerklePath& merkle_path /*output*/) const = 0;
+
+    /**
+     * @brief Get the hash functions used for each layer of the Merkle tree.
+     *
+     * @return A vector of Hash objects representing the hash function for each layer.
+     */
+    const std::vector<Hash>& get_layer_hashes() const { return m_layer_hashes; }
```
mickeyasa, line 73: why do we need that?

icicle/include/icicle/backend/merkle/merkle_tree_backend.h
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * @brief Get the size of each leaf element in bytes.
+     *
+     * @return Size of each leaf element in bytes.
+     */
+    uint64_t get_leaf_element_size() const { return m_leaf_element_size; }
```
mickeyasa, line 80: same

icicle/include/icicle/merkle/merkle_path.h
```diff
@@ -0,0 +1,115 @@
...
+    template <typename T>
+    T* access_at_offset(uint64_t offset)
+    {
+      if (offset >= m_size) { throw std::out_of_range("Offset out of bounds"); }
+      return reinterpret_cast<T*>(m_path_data.get() + offset);
+    }
+
+  private:
+    std::unique_ptr<std::byte[]> m_path_data; ///< Unique ownership of the path data.
+    uint64_t m_size;                          ///< Size of the path data.
```
mickeyasa, line 111: don't you want to keep is as a vector? or the rust transactions are going to be problematic?

icicle/include/icicle/backend/merkle/merkle_tree_backend.h
yshekel: The tree needs it to verify.

icicle/include/icicle/backend/merkle/merkle_tree_backend.h
yshekel: same reason, for verify

icicle/include/icicle/merkle/merkle_path.h
yshekel: I could keep it as a vector and access the data like that https://cplusplus.com/reference/vector/vector/data/
Not sure which is better. Do you prefer vector?

mickeyasa (approved):
nice

mickeyasa (approved):

mickeyasa (approved):

PR #612, opened 2024-09-12
[CI] Fix device race in golang tests

jeremyfelder:
## Describe the changes

This PR updates golang tests to include a suite that allows us to wrap each test in a function that will handle calling runtime.SetDevice

### DIFF
--- a/.github/workflows/golang.yml
+++ b/.github/workflows/golang.yml
@@ -36,6 +36,9 @@ jobs:
     with:
       pr-number: ${{ github.event.pull_request.number }}
 
+  # TODO - add runtime tests to the workflow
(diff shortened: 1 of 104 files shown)


yshekel (approved):

nonam3e (commented):
Nice fix 💪
LGTM but bn254 golang test failed 😢

LeonHibnik (approved):

jeremyfelder:
> Nice fix 💪 LGTM but bn254 golang test failed 😢

Not sure why that failed. From the values of the failure it looks like the functions didn't run and just returned the default values. Everything passes locally consistently.

I added a proper exit code for the tests so this would be easier to catch in the future

PR #610, opened 2024-09-08
merge Blake2s and Keccak to HASH

aviadingo:
## Describe the changes

added integration for Keccak256 Keccak512 Sha3_256 Sha3_512 and Blake2s
using new Hash API

This PR overrides old PR that was directly to main.

minor changes regarding last PR, due to integration mostly.

will be merged with the hash branch and then all together to main

PR #616, opened 2024-09-23
blake2s cpp and rust API

emirsoyturk:
## Changes

This PR updates the existing Blake2s implementation by integrating a C++ API and a Rust wrapper API.

### DIFF
--- a/icicle/src/hash/CMakeLists.txt
+++ b/icicle/src/hash/CMakeLists.txt
@@ -1,5 +1,10 @@
 set(TARGET icicle_hash)
 
-add_library(${TARGET} STATIC keccak/extern.cu)
+set(SRC ${CMAKE_SOURCE_DIR}/src/hash)
+
+set(HASH_SOURCE ${SRC}/keccak/extern.cu)
+list(APPEND HASH_SOURCE ${SRC}/blake2s/extern.cu)
+
+add_library(${TARGET} STATIC ${HASH_SOURCE})
 target_include_directories(${TARGET} PUBLIC ${CMAKE_SOURCE_DIR}/include/)
 set_target_properties(${TARGET} PROPERTIES OUTPUT_NAME "ingo_hash")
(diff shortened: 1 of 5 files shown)


icicle/src/hash/CMakeLists.txt
```diff
@@ -1,5 +1,10 @@
 set(TARGET icicle_hash)
 
-add_library(${TARGET} STATIC keccak/extern.cu)
+set(SRC ${CMAKE_SOURCE_DIR}/src/hash)
+
+set(HASH_SOURCE ${SRC}/keccak/extern.cu)
```
yshekel, line 5: minor but you could do both in same line rather then use append

wrappers/rust/icicle-hash/src/blake2s/tests.rs
```diff
@@ -0,0 +1,50 @@
...
+        let height = 22;
+        let input_block_len = 136;
+        let leaves = vec![1u8; (1 << height) * input_block_len];
+        let mut digests = vec![0u64; merkle_tree_digests_len((height + 1) as u32, 2, 1)];
+
+        let leaves_slice = HostSlice::from_slice(&leaves);
+        let digests_slice = HostSlice::from_mut_slice(&mut digests);
+
+        build_blake2s_merkle_tree(leaves_slice, digests_slice, height, input_block_len, &config).unwrap();
+        println!("Root: {:?}", digests_slice[0]);
```
yshekel, line 48: note that both tests don't test correctness

yshekel (approved):

jeremyfelder:
@emirsoyturk @yuvalingo It seems like the base branch is from V2. Is this PR meant to be against V3? I think this [PR](https://github.com/ingonyama-zk/icicle/pull/610)/branch is the latest for blake2s on V3

emirsoyturk:
@jeremyfelder right. Maybe we can merge it into the `aviad/blake2s` branch, and then I can open another branch for V3. That way, we would have Blake2s for both V2 and V3.

PR #617, opened 2024-09-23
Stas/vecop/reduction

svpolonsky:
## Describe the changes

This PR adds vector reduction operations (sum of elements, product of elements) to CPU backend. 

Specific changes:

1. Debugged CPU reduction operations 
2. Registered at the frontend
3. New C++ example for vector operations
4. Added rust, golang warppers


examples/c++/vector-api/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,25 @@
+# Make sure NVIDIA Container Toolkit is installed on your host
+
+# Use the specified base image
+FROM nvidia/cuda:12.0.0-devel-ubuntu22.04
+
+# Update and install dependencies
+RUN apt-get update && apt-get install -y \
```
yshekel, line 7: please remove this file. We don't recommend it anymore

examples/c++/vector-api/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,22 @@
+{
+    "name": "Icicle Examples: polynomial multiplication",
```
yshekel, line 2: same

examples/c++/vector-api/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+cmake_minimum_required(VERSION 3.18)
+
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
+
+project(example)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g")
```
yshekel, line 7: is that intentional in debug?

examples/c++/vector-api/README.md
```diff
@@ -0,0 +1,30 @@
...
+ 
+ All these operations can be performed on a host or device both synchronously and asynchronously.
+
+## Running the example
+
+```sh
+# for CPU
+./run.sh -d CPU
+# for CUDA
+./run.sh -d CUDA -b /path/to/cuda/backend/install/dir
```
yshekel, line 21: this won't work right? I mean since it's not implemented for CUDA

examples/c++/vector-api/example.cpp
```diff
@@ -0,0 +1,188 @@
...
+#include <cassert>
+
+#include "icicle/runtime.h"
+#include "icicle/api/bn254.h"
+#include "icicle/utils/log.h"
+
+
+// SP: I undertstand this code is auto-generated, but I can't get scrip/gen_c_api.py to work. 
+
+extern "C" eIcicleError bn254_vector_product(
```
yshekel, line 13: I think it's better to use the template function in vec_ops.h rather than those C apis that are for FFI.

icicle/backend/cpu/src/field/cpu_vec_ops.cpp
```diff
@@ -154,17 +155,17 @@ class VectorOpTask : public TaskBase
   // Single worker functionality to execute sum(vector)
   void vector_sum()
   {
-    *m_output = m_op_a[0];
-    for (uint64_t i = 1; i < m_nof_operations; ++i) {
-      *m_output = *m_output + m_op_a[i];
+    m_intermidiate_res = T::zero();
```
yshekel, line 158: why? was there a bug?

icicle/include/icicle/vec_ops.h
```diff
@@ -46,6 +46,37 @@ namespace icicle {
...
+   * @tparam T Type of the elements in the vector.
+   * @param vec_a Input vector.
+   * @param n Number of elements in the vector.
+   * @param config Configuration for the operation.
+   * @param output Output scalar to store the result.
+   * @return eIcicleError Error code indicating success or failure.
+   */
+
+  template <typename T>
+  eIcicleError vector_product(const T* vec_a, uint64_t n, const VecOpsConfig& config, T* output, uint64_t offset, uint64_t  stride);
```
yshekel, line 63: The order of params should be consistent with other APIs: first inputs and non-optional params, then config, then output. You need to move offset and stride to be before the config (see slice as reference).
In addition, maybe offset is unnecessary since a user can easily do it with pointer arithmetics.

examples/c++/vector-api/.devcontainer/devcontainer.json
svpolonsky: Done. I also removed .devcontainer/

examples/c++/vector-api/README.md
svpolonsky: no CUDA yet. I removed CUDA in README. I will add later, when we have it in vector ops.

icicle/backend/cpu/src/field/cpu_vec_ops.cpp
svpolonsky: yes. For reduction ops we should use m_intermidiate_res and not m_output. m_output will aggregate m_intermidiate_res from different threads.

icicle/include/icicle/vec_ops.h
svpolonsky: Perfect suggestion! Let's first figure out how to use `vec_ops.h` in c++ examples. Next I will reorder parameters accordingly.

svpolonsky:
I updated the branch based on Yuval's comments.

PR #618, opened 2024-09-24
Add FRI folding for M31

jeremyfelder:
This PR adds FRI folding for M31 field.


icicle/src/fri/fri.cu
```diff
@@ -0,0 +1,148 @@
...
+        ibutterfly(f0_px, f1_px, S::inverse(domain_ys[idx / 2]));
+        E f_prime = f0_px + alpha * f1_px;
+        auto folded_eval_idx = idx / 2;
+        folded_eval[folded_eval_idx] = folded_eval[folded_eval_idx] * alpha_sq + f_prime;
+      }
+    }
+  } // namespace
+
+  template <typename S, typename E>
+  cudaError_t fold_line(E* eval, S* domain_xs, E alpha, E* folded_eval, int n, FriConfig& cfg)
```
yshekel, line 50: is int enough for size? for ntt we had this bug for large ntts in babybear.

icicle/src/fri/fri.cu
```diff
@@ -0,0 +1,148 @@
...
+
+    int num_threads = 256;
+    int num_blocks = (n / 2 + num_threads - 1) / num_threads;
+    fold_line_kernel<<<num_blocks, num_threads, 0, stream>>>(d_eval, d_domain_xs, alpha, d_folded_eval, n);
+
+    // Move folded_eval back to host if requested
+    if (!cfg.are_results_on_device) {
+      CHK_IF_RETURN(cudaMemcpyAsync(folded_eval, d_folded_eval, sizeof(E) * n / 2, cudaMemcpyDeviceToHost, stream));
+    }
+
```
yshekel, line 91: need to release the allocated gpu memory

icicle/src/fri/fri.cu
```diff
@@ -0,0 +1,148 @@
...
+    int num_blocks = (n / 2 + num_threads - 1) / num_threads;
+    fold_circle_into_line_kernel<<<num_blocks, num_threads, 0, stream>>>(
+      d_eval, d_domain_ys, alpha, alpha_sq, d_folded_eval, n);
+
+    // Move folded_evals back to host if requested
+    if (!cfg.are_results_on_device) {
+      CHK_IF_RETURN(cudaMemcpyAsync(folded_eval, d_folded_eval, sizeof(E) * n / 2, cudaMemcpyDeviceToHost, stream));
+    }
+
+    // Sync if stream is default stream
```
yshekel, line 143: same comment about memory and the int type for n.

wrappers/rust/icicle-fields/icicle-m31/Cargo.toml
```diff
@@ -17,3 +17,4 @@ cmake = "0.1.50"
 [features]
 default = []
 devmode = ["icicle-core/devmode"]
+fri = []
```
yshekel, line 20: why does it default to off?

yshekel (approved):
looks good to me. Maybe would be nice to test against stwo rather than hardcoded values.

jeremyfelder:
> Maybe would be nice to test against stwo rather than hardcoded values.

@yshekel These hard coded values are taken from Stwo's unit tests and are deterministic values. I can change the tests to use Stwo's crate to generate all of these values instead of using the hardcoded ones (though they will be the same). is that what you mean?

ChickenLover (approved):

PR #619, opened 2024-09-26
Arkworks conversion example

yshekel:
This PR:
(1) introduces a new Rust example for converting Arkworks types to ICICLE and back. It is meant to be used as reference for users using ICICLE.
(2) update to/from Montgomery trait and implementations to support HostOrDeviceSlice instead of DeviceSlice only so that it can work for host data as well.
(3) choose better c for cpu msm for way faster cpu (for non precompute case)


examples/rust/arkworks/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "arkworks"
```
vhnatyk, line 2: a bit confusing name

examples/rust/arkworks/src/main.rs
```diff
@@ -0,0 +1,333 @@
...
+
+    // convert the ICICLE result back to Ark projective and compare
+    let ark_res_from_icicle = icicle_to_ark_projective_points(&icicle_msm_result);
+    assert_eq!(ark_res_from_icicle[0], ark_msm_result);
+
+    //============================================================================================//
+    //======================= Part 5b: compute MSM on device memory  =============================//
+    //============================================================================================//
+
+    let is_device_shraing_host_mem = icicle_runtime::runtime::get_device_properties()
```
krakhit: is_device_sharing_host_mem

examples/rust/arkworks/src/main.rs
```diff
@@ -0,0 +1,333 @@
...
+        .unwrap()
+        .using_host_memory;
+
+    if is_device_shraing_host_mem {
```
krakhit: is_device_sharing_host_mem

examples/rust/arkworks/Cargo.toml
yshekel: replaced with arkworks_icicle_conversions

examples/rust/arkworks/src/main.rs
```diff
@@ -1,54 +1,333 @@
...
     icicle_scalars
 }
 
+fn ark_to_icicle_scalars_async<T, I>(ark_scalars: &[T], stream: &IcicleStream) -> DeviceVec<I>
+where
+    T: PrimeField,
+    I: FieldImpl + MontgomeryConvertible,
+{
+    // SAFETY: Reinterpreting Arkworks field elements as Icicle-specific scalars
+    let icicle_scalars = unsafe { &*(&ark_scalars[..] as *const _ as *const [I]) };
```
krakhit: Rust recommended to remove the double slicing here, since ark_scalars is already a slice. I think it allocates memory again when creating this extra slice. Minor change. 

fn ark_to_icicle_scalars_async<T, I>(ark_scalars: &[T], stream: &IcicleStream) -> DeviceVec<I>
where
    T: PrimeField,
    I: FieldImpl + MontgomeryConvertible,
{
    // SAFETY: Reinterpreting Arkworks field elements as Icicle-specific scalars
    let icicle_scalars = unsafe { &*(ark_scalars as *const _ as *const [I]) };

    // Create a HostSlice from the mutable slice
    let icicle_host_slice = HostSlice::from_slice(icicle_scalars);

    let mut icicle_scalars = DeviceVec::<I>::device_malloc_async(ark_scalars.len(), stream).unwrap();
    icicle_scalars
        .copy_from_host(icicle_host_slice)
        .unwrap();

    // Convert from Montgomery representation using the Icicle type's conversion method
    I::from_mont(&mut icicle_scalars, stream);
    icicle_scalars
}

examples/rust/arkworks/src/main.rs
```diff
@@ -1,54 +1,333 @@
...
+
+    // convert the ICICLE result back to Ark projective and compare
+    let ark_res_from_icicle = icicle_to_ark_projective_points(&icicle_msm_result);
+    assert_eq!(ark_res_from_icicle[0], ark_msm_result);
+
+    //============================================================================================//
+    //======================= Part 5b: compute MSM on device memory  =============================//
+    //============================================================================================//
+
+    let is_device_shraing_host_mem = icicle_runtime::runtime::get_device_properties()
```
krakhit: nitpick:  `is_device_sharing_host_mem`

examples/rust/arkworks/src/main.rs
```diff
@@ -1,54 +1,333 @@
...
+        .unwrap()
+        .using_host_memory;
+
+    if is_device_shraing_host_mem {
```
krakhit: nitpick:  `is_device_sharing_host_mem`

examples/rust/arkworks/src/main.rs
yshekel: fixed

examples/rust/arkworks/src/main.rs
yshekel: updated but it doesn't seem to be faster

krakhit (approved):

PR #620, opened 2024-09-29
Miki/MTnewAPI

mickeyasa:
merkle tree according to the new API


icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
```diff
@@ -9,39 +16,404 @@ namespace icicle {
...
+          << "Each layer output size must divide the above layer input size. Otherwise its not a tree.\n"
+          << "Layer " << layer_idx << " input size = " << layer_hashes[layer_idx + 1].input_default_chunk_size() << "\n"
+          << "Layer " << layer_idx + 1 << " output size = " << layer_hashes[layer_idx].output_size() << "\n";
+
+        // initialize m_layers with hashes
+        m_layers[layer_idx].m_hash = layer_hashes[layer_idx];
+
+        // Calculate m_path_size
+        if (0 < layer_idx) {
+          m_prooned_path_size +=
```
yshekel, line 39: pruned, not prooned

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
```diff
@@ -9,39 +16,404 @@ namespace icicle {
...
+
+      copy_to_path_from_store_min_layer(input_chunk_offset, is_pruned, path);
+      // print_proof(merkle_proof);
+      return eIcicleError::SUCCESS;
+    }
+
+    // Debug functions
+    eIcicleError print_tree(const std::byte* leaves, uint64_t leaves_size) const
+    {
+      ICICLE_ASSERT(m_tree_already_built) << "Tree cannot print before built";
```
yshekel, line 171: You can return an error instead of crashing in such cases

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
```diff
@@ -9,39 +16,404 @@ namespace icicle {
...
     }
 
-    eIcicleError build(const std::byte* leaves, uint64_t size, const MerkleTreeConfig& config) override
+    // TODO: handle size
+    eIcicleError build(const std::byte* leaves, uint64_t leaves_size, const MerkleTreeConfig& config) override
     {
-      ICICLE_LOG_DEBUG << "CPU CPUMerkleTreeBackend::build() called with " << size << " bytes of leaves";
-      return eIcicleError::SUCCESS; // TODO: Implement tree-building logic
+      TasksManager<HashTask> task_manager(get_nof_workers(config)); // Run workers.
+      ICICLE_ASSERT(!m_tree_already_built) << "Tree cannot be built more than one time";
```
yshekel, line 50: why? I think it's worth an error message telling the user to construct another tree and return an error rather than crashing.
The problem with crashing is mainly that you crash in C++ and it's really unclear in rust/Go what happened. With an error message Rust can panic and I think it's clearer for the user where it crashed and then he also sees an error message using ICICLE_LOG_ERROR << "..."

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
```diff
@@ -9,39 +16,404 @@ namespace icicle {
...
+          auto cur_segment_iter = m_map_segment_id_2_inputs.find(cur_segment_id);
+          cur_segment_iter->second->m_nof_inputs_ready +=
+            m_layers[completed_layer_idx].m_hash.output_size() * NOF_OPERATIONS_PER_TASK;
+
+          // check if cur segment is ready to be executed
+          const Hash cur_hash = m_layers[cur_layer_idx].m_hash;
+          const uint64_t nof_hashes_in_seg = std::min(
+            m_layers[cur_layer_idx].m_nof_hashes - cur_segment_idx * NOF_OPERATIONS_PER_TASK,
+            uint64_t(NOF_OPERATIONS_PER_TASK));
+          ICICLE_ASSERT(nof_hashes_in_seg > 0) << "Edge case negative number of hashes";
```
yshekel, line 93: what is this edge case? bug? input is wrong? for the latter, can you identify it earlier and return a clear error?

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
```diff
@@ -9,39 +16,404 @@ namespace icicle {
...
       MerkleProof& merkle_proof) const override
     {
-      ICICLE_LOG_DEBUG << "CPU CPUMerkleTreeBackend::get_merkle_proof() called for element index " << element_idx;
-      // Dummy implementation. TODO implement
-      uint64_t root = 123;
-      merkle_proof.allocate(
-        false /*pruned*/, element_idx, leaves + element_idx * get_leaf_element_size(), get_leaf_element_size(),
-        (const std::byte*)&root, sizeof(root));
-      return eIcicleError::SUCCESS; // TODO: Implement proof generation logic
+      ICICLE_ASSERT(m_tree_already_built) << "Tree cannot generate a merkle proof before built";
```
yshekel, line 129: same regarding error

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
```diff
@@ -9,39 +16,404 @@ namespace icicle {
...
+        const std::vector<Hash> sub_tree_layer_hashes(
+          m_layer_hashes.begin(), m_layer_hashes.begin() + m_output_store_min_layer + 1);
+        CPUMerkleTreeBackend sub_tree(sub_tree_layer_hashes, m_leaf_element_size, 0);
+
+        // build the sub tree
+        const uint64_t sub_tree_leaves_size =
+          l0_total_input_size * m_layers[0].m_nof_hashes / m_layers[m_output_store_min_layer].m_nof_hashes;
+        const uint64_t sub_tree_leaves_offset = (element_offset / sub_tree_leaves_size) * sub_tree_leaves_size;
+        const std::byte* sub_tree_leaves = &(leaves[sub_tree_leaves_offset]);
+        sub_tree.build(sub_tree_leaves, sub_tree_leaves_size, config);
```
yshekel, line 156: is this computing the minimal work that is required for the path?

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
```diff
@@ -9,39 +16,404 @@ namespace icicle {
...
+
+        // retrieve from the sub tree the path and increment path
+        const uint64_t sub_tree_input_chunk_offset = element_offset % sub_tree_leaves_size;
+        sub_tree.copy_to_path_from_store_min_layer(sub_tree_input_chunk_offset, is_pruned, path);
+      }
+
+      copy_to_path_from_store_min_layer(input_chunk_offset, is_pruned, path);
```
yshekel, line 163: did you mean to copy to path both here and above? shouldn't they write to different offsets?

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
```diff
@@ -9,39 +16,404 @@ namespace icicle {
...
+
+      // Set task next segment to handle return data
+      task->m_next_segment_idx = next_segment_idx;
+
+      // dispatch task
+      task->dispatch();
+    }
+
+    // restore the proof path from the tree
+    void copy_to_path_from_store_min_layer(const uint64_t input_chunk_offset, bool is_prooned, std::byte*& path) const
```
yshekel, line 384: the reference here is really confusing in terms of API. I would suggest returning the size as return value and let the caller increment the path pointer.

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
```diff
@@ -9,39 +16,404 @@ namespace icicle {
...
+        std::cout << std::dec << "Layer " << layer_idx << ": " << m_layers[layer_idx].m_hash.input_default_chunk_size()
+                  << " -> " << m_layers[layer_idx].m_hash.output_size() << std::endl;
+        print_bytes(
+          m_layers[layer_idx].m_results.data(), m_layers[layer_idx].m_nof_hashes,
+          m_layers[layer_idx].m_hash.output_size());
+      }
+      return eIcicleError::SUCCESS;
+    }
+
+    eIcicleError print_proof(const MerkleProof& proof) const
```
yshekel, line 185: maybe better to accept an ostream so that you can write to file/network too

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
```diff
@@ -9,39 +16,404 @@ namespace icicle {
...
+      }
+      auto [root, root_size] = proof.get_root();
+      std::cout << "Layer " << m_layers.size() - 1 << "(Root):" << std::endl;
+      print_bytes(root, 1, root_size);
+      return eIcicleError::SUCCESS;
+    }
+
+    eIcicleError get_hash_result(int layer_index, int hash_index, const std::byte*& hash_result /*OUT*/) const
+    {
+      ICICLE_ASSERT(m_tree_already_built) << "Tree cannot get hash result before built";
```
yshekel, line 210: if you anyway assert or return success then just return the pointer or null with an ICICLE_LOG_ERROR.
this reference to pointer is less clear when reading the top level only

icicle/include/icicle/fields/quartic_extension.h
```diff
@@ -241,7 +241,7 @@ class QuarticExtensionField
       FF::reduce(
         (CONFIG::nonresidue_is_negative
            ? (FF::mul_wide(xs.real, x0) + FF::template mul_unsigned<CONFIG::nonresidue>(FF::mul_wide(xs.im2, x2)))
-           : (FF::mul_wide(xs.real, x0)) - FF::template mul_unsigned<CONFIG::nonresidue>(FF::mul_wide(xs.im2, x2)))),
+           : (FF::mul_wide(xs.real, x0))-FF::template mul_unsigned<CONFIG::nonresidue>(FF::mul_wide(xs.im2, x2)))),
```
yshekel, line 244: that's weird that the formatter did that

icicle/include/icicle/fields/storage.h
```diff
@@ -40,8 +38,7 @@ struct
 #ifdef __CUDA_ARCH__
   __align__(LIMBS_ALIGNMENT(LIMBS_COUNT))
 #endif
-    storage
-{
+    storage {
```
yshekel, line 41: you format rules are somehow different

icicle/include/icicle/merkle/merkle_tree.h
```diff
@@ -116,17 +142,58 @@ namespace icicle {
...
+            return eIcicleError::SUCCESS;
+          }
+          // run current layer hash
+          hash_results.resize(hash_output_size);
+          layer_hashes[layer_idx].hash(path, hash_input_size, hash_config, hash_results.data());
+          path += hash_input_size;
+        }
+      }
+
+      // Copare the last hash result with the root
```
yshekel, line 194: typo Compare

icicle/run
```diff
@@ -0,0 +1,4 @@
+rm -rf build
```
yshekel, line 1: did you commit that by mistake?

icicle/tests/test_hash_api.cpp
```diff
@@ -149,22 +148,25 @@ TEST_F(HashApiTest, MerkleTree)
   auto layer2_hash = HashSumBackend::create(20, 8); // input 20 bytes, output 8 bytes         20B  ->  8B    output
   auto leaf_element_size = 4;
   auto merkle_tree =
-    MerkleTree::create({layer0_hash, layer1_hash, layer2_hash}, leaf_element_size, 2 /*min level to store*/);
+    MerkleTree::create({layer0_hash, layer1_hash, layer2_hash}, leaf_element_size, 1 /*min level to store*/);
 
   // build tree
-  ICICLE_CHECK(merkle_tree.build(leaves, input_size, config));
+  ICICLE_CHECK(merkle_tree.build(leaves, nof_leaves, config));
```
yshekel, line 154: we can time this. Interesting to see it compared to CUDA

icicle/src/hash/merkle_c_api.cpp
```diff
@@ -119,14 +119,16 @@ const std::byte* icicle_merkle_tree_get_root(icicle::MerkleTree* tree, size_t* o
 eIcicleError icicle_merkle_tree_get_proof(
   icicle::MerkleTree* tree,
   const std::byte* leaves,
+  uint64_t leaves_size,
```
yshekel, line 122: this is breaking rust right?

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
mickeyasa: כןסקג

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
mickeyasa: Koren added that.
Removed until full verification once he'll be back

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
mickeyasa: thats a denbug function to be used by me - no one can access it

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
mickeyasa: Its a debug function - no one can access it

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
mickeyasa: the path from the sub tree is the begining of the full path.
Its very comfortable that the path is incremented so it will be ready for the rest of the path concatention from the original tree

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
mickeyasa: fixed

icicle/src/hash/merkle_c_api.cpp
mickeyasa: right.
get proof must recieve leaves_size for padding

icicle/run
mickeyasa: yes

icicle/tests/test_hash_api.cpp
mickeyasa: done

icicle/include/icicle/merkle/merkle_tree.h
mickeyasa: fixed

icicle/include/icicle/fields/storage.h
mickeyasa: strange - but they look better

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
```diff
@@ -9,39 +16,418 @@ namespace icicle {
...
+          // Calculate path_size
+          m_pruned_path_size +=
+            layer_hashes[layer_idx].input_default_chunk_size() - layer_hashes[layer_idx - 1].output_size();
+          m_full_path_size += layer_hashes[layer_idx].input_default_chunk_size();
+        }
       }
     }
 
-    eIcicleError build(const std::byte* leaves, uint64_t size, const MerkleTreeConfig& config) override
+    // TODO: handle size
```
yshekel, line 53: is this handled?

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
```diff
@@ -9,39 +16,418 @@ namespace icicle {
...
+    eIcicleError build(const std::byte* leaves, uint64_t leaves_size, const MerkleTreeConfig& config) override
     {
-      ICICLE_LOG_DEBUG << "CPU CPUMerkleTreeBackend::build() called with " << size << " bytes of leaves";
-      return eIcicleError::SUCCESS; // TODO: Implement tree-building logic
+      TasksManager<HashTask> task_manager(get_nof_workers(config)); // Run workers.
+      if (m_tree_already_built) {
+        ICICLE_LOG_ERROR << "Tree cannot be built more than one time\n";
```
yshekel, line 58: (1) no need to add \n since there is a newline anyway.
(2) maybe it's a warning?

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
```diff
@@ -9,39 +16,418 @@ namespace icicle {
...
+        return eIcicleError::INVALID_ARGUMENT;
+      }
+      m_tree_already_built = true; // Set the tree status as built
+      uint64_t l0_segment_idx = 0; // Index for the segment of hashes from layer 0 to send
+      const int nof_layers = m_layers.size();
+      init_layers_db(config);
+      const uint64_t nof_segments_at_l0 =
+        (m_layers[0].m_nof_hashes + NOF_OPERATIONS_PER_TASK - 1) / NOF_OPERATIONS_PER_TASK;
+
+      // run until the root is processed
```
yshekel, line 68: don't you assume input size is smaller than some size? you don't seem to check it

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
```diff
@@ -9,39 +16,418 @@ namespace icicle {
...
     {
-      ICICLE_LOG_DEBUG << "CPU CPUMerkleTreeBackend::get_merkle_proof() called for element index " << element_idx;
-      // Dummy implementation. TODO implement
-      uint64_t root = 123;
-      merkle_proof.allocate(
-        false /*pruned*/, element_idx, leaves + element_idx * get_leaf_element_size(), get_leaf_element_size(),
-        (const std::byte*)&root, sizeof(root));
-      return eIcicleError::SUCCESS; // TODO: Implement proof generation logic
+      if (!m_tree_already_built) {
+        ICICLE_LOG_ERROR << "Tree cannot generate a merkle proof before built\n";
```
yshekel, line 140: again maybe this is a warning and not error

icicle/include/icicle/hash/hash.h
```diff
@@ -22,7 +22,7 @@ namespace icicle {
      *
      * @param backend A shared pointer to a backend hash implementation.
      */
-    explicit Hash(std::shared_ptr<HashBackend> backend) : m_backend(std::move(backend)) {}
+    explicit Hash(std::shared_ptr<HashBackend> backend = nullptr) : m_backend(std::move(backend)) {}
```
yshekel, line 25: maybe better let the caller pass nullptr rather than default value so that this is explicit and not implicit.

icicle/include/icicle/hash/keccak.h
```diff
@@ -15,7 +15,7 @@ namespace icicle {
    */
   Hash create_keccak_256_hash(uint64_t input_chunk_size = 0);
   struct Keccak256 {
-    inline static Hash create(uint64_t input_chunk_size = 0) { return create_keccak_256_hash(); }
+    inline static Hash create(uint64_t input_chunk_size = 0) { return create_keccak_256_hash(input_chunk_size); }
```
yshekel, line 18: maybe better rename it default_input_chunk_size?

icicle/include/icicle/merkle/merkle_tree.h
```diff
@@ -116,17 +142,58 @@ namespace icicle {
...
+      const uint nof_layers = layer_hashes.size();
+      auto [leaf, hash_input_size, leaf_idx] = merkle_proof.get_leaf();
+      HashConfig hash_config;
 
-      ICICLE_LOG_DEBUG << "In MerkleTree::verify()";
+      // offset in bytes to the selected leaf at the full tree
+      uint64_t leaf_element_start = leaf_idx * m_backend->get_leaf_element_size();
+
+      // Hash the leaves into hash_results
+      uint hash_output_size = layer_hashes[0].output_size();
```
yshekel, line 154: better use front() or at least assert it is not empty to avoid invalid memory access

icicle/tests/test_hash_api.cpp
```diff
@@ -313,38 +316,210 @@ class HashSumBackend : public HashBackend
...
 
-TEST_F(HashApiTest, MerkleTree)
+void assert_valid_tree(
+  const MerkleTree& tree,
+  int nof_input_byes,
+  const std::byte* inputs,
+  const std::vector<Hash>& hashes,
+  const MerkleTreeConfig& config)
+{
+  int nof_outputs = nof_input_byes * hashes[0].output_size() / hashes[0].input_default_chunk_size();
```
yshekel, line 326: again must check size before accessing [0]

icicle/tests/test_hash_api.cpp
```diff
@@ -313,38 +316,210 @@ class HashSumBackend : public HashBackend
...
+  std::byte* layer_in = new std::byte[nof_input_byes]; // Going layer by layer - having the input layer as the largest
```
yshekel, line 327: you should always avoid new/delete in modern c++.
Use
`auto layer_in = std::make_unique<std::byte[]>(nof_inout_bytes); `
instead and no need to release. It also handles the case of exceptions

icicle/tests/test_hash_api.cpp
```diff
@@ -313,38 +316,210 @@ class HashSumBackend : public HashBackend
...
+
+template <typename in_limb>
+void assert_valid_tree(
+  const MerkleTree& tree,
+  int nof_inputs,
+  const in_limb* inputs,
+  const std::vector<Hash>& hashes,
+  const MerkleTreeConfig& config)
+{
+  return assert_valid_tree(
```
yshekel, line 371: return type is void so no need the return statement.
Maybe better to return a bool and not assert though.

icicle/tests/test_hash_api.cpp
```diff
@@ -313,38 +316,210 @@ class HashSumBackend : public HashBackend
...
+TEST_F(HashApiTest, MerkleTreeBasic)
 {
-  ICICLE_CHECK(icicle_set_device(s_reference_target)); // TODO CUDA too
   // define input
-  constexpr int nof_leaves = 100;
-  uint32_t leaves[nof_leaves];
-  const size_t input_size = sizeof(leaves);
-  for (int i = 0; i < nof_leaves; ++i) {
-    leaves[i] = i;
+  int leaf_size = sizeof(merkel_hash_t);
```
yshekel, line 378: it's not a hash so the type is confusing

icicle/tests/test_hash_api.cpp
```diff
@@ -313,38 +316,210 @@ class HashSumBackend : public HashBackend
...
+  output_store_min_layer = output_store_min_layer & 3; // Ensure index is in a valid 0-3 range
+  ICICLE_LOG_DEBUG << "Min store layer:\t" << output_store_min_layer;
+
+  auto prover_tree = MerkleTree::create(hashes, sizeof(merkel_hash_t), output_store_min_layer);
+
+  auto verifier_tree = MerkleTree::create(hashes, sizeof(merkel_hash_t), output_store_min_layer);
+
+  // build tree
+
+  // START_TIMER(MerkleTree_build)
```
yshekel, line 406: you cann keep the timer. Just remove the std::cout since the END_TIMER is already printing that

icicle/tests/test_hash_api.cpp
```diff
@@ -313,38 +316,210 @@ class HashSumBackend : public HashBackend
...
+
+    // Test invalid proof (By adding random data to the proof)
+    verification_valid = true;
+    ICICLE_CHECK(prover_tree.get_merkle_proof(other_leaves, nof_leaves, leaf_idx, true, config, merkle_proof));
+    ICICLE_CHECK(verifier_tree.verify(merkle_proof, verification_valid));
+    ASSERT_FALSE(verification_valid);
   }
+}
+
+TEST_F(HashApiTest, MerkleTree)
```
yshekel, line 449: what is the difference between the two tests? it's almost the same.
Can you make a test that scales with input size? So that we can easily test execution time on CPU and GPU.

wrappers/rust/icicle-core/src/merkle/mod.rs
```diff
@@ -263,11 +265,14 @@ impl MerkleTree {
...
 
         let proof = MerkleProof::new().unwrap();
+        let byte_size = (leaves.len() * std::mem::size_of::<T>()) as u64;
         let result = unsafe {
             icicle_merkle_tree_get_proof(
                 self.handle,
                 leaves.as_ptr() as *const u8,
+                byte_size,
                 leaf_idx as u64,
+                false,
```
yshekel, line 275: why false? you need to take this argument

yshekel (changes requested):
I did not review the CPU backend implementation in depth. @Koren-Brand  did you?

a few general notes:
(1) it seems that the tests are only for CPU and when we add CUDA we will need another PR to add tests for CUDA too.
(2) I think we need a test fir large input and preferably one where input size can be modified and it works correctly (by computing the number of layers correctly etc.) It will be also used for benchmarking.
(3) You only the test the happy path. But I don't see any tests where inputs are incorrect. For example what happens when input is too large? or something is ill-defined?
(4) padding is not tested

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
mickeyasa: not yet - will be at 3.2

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
mickeyasa: \n fixed
its not a warning. I didn't check but it will probably crush if you rebuild

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
mickeyasa: currently size is ignored.
I'll add an error

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
mickeyasa: same

icicle/include/icicle/merkle/merkle_tree.h
mickeyasa: how front() helps?

icicle/include/icicle/hash/keccak.h
mickeyasa: done

icicle/include/icicle/hash/hash.h
mickeyasa: done

wrappers/rust/icicle-core/src/merkle/mod.rs
mickeyasa: you wrote this file, no?

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
mickeyasa: done

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
mickeyasa: done

icicle/include/icicle/merkle/merkle_tree.h
yshekel: actually it doesn't. It is also undefined behavior if the vector is empty. 
The safe way seems to be https://en.cppreference.com/w/cpp/container/vector/at
or simply assert the vector size is large enough

wrappers/rust/icicle-core/src/merkle/mod.rs
yshekel: Yes but Koren added this argument. I will fix it

yshekel (approved):
updated tests and rust api to accept the pruned-path argument when computing proofs

mickeyasa:
Added merkle tree support with CPU backend

PR #622, opened 2024-10-07
Keccak CPU backend

yshekel:
Implement Keccak and SHA-3 for CPU backend, including tests.

cuda-backend-branch: yshekel/keccak_cuda


icicle/backend/cpu/src/hash/cpu_keccak.cpp
```diff
@@ -1,119 +1,316 @@
...
-      for (int i = 0; i < output_size() * config.batch; ++i) {
-        output[i] = std::byte(i % 256);
+      const auto digest_size_in_bytes = output_size();
+      const auto single_input_size = get_single_chunk_size(
+        size); // if size==0 using default input chunk size. This is useful for Merkle-Tree constructions
+      ICICLE_LOG_DEBUG << name() << "::hash() called, batch=" << config.batch << ", input-size=" << size << " bytes";
+
+      for (unsigned batch_idx = 0; batch_idx < config.batch; ++batch_idx) {
+        eIcicleError err = sha3_hash_buffer(
+          8 * digest_size_in_bytes, m_is_keccak, input + batch_idx * single_input_size, single_input_size,
```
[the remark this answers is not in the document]
yshekel: done

icicle/backend/cpu/src/hash/cpu_keccak.cpp
```diff
@@ -1,119 +1,319 @@
...
+        if (err != eIcicleError::SUCCESS) { return err; }
       }
       return eIcicleError::SUCCESS;
     }
+
+  private:
+    static const uint64_t s_keccakf_rndc[24];
+    static const unsigned s_keccakf_rotc[24];
+    static const unsigned s_keccakf_piln[24];
+    const bool m_is_keccak;
```
mickeyasa, line 57: what is that memeber for?

yshekel: The implementation that Aviad used needs to know that since it's both for keccak and sha3.
Since the same object implements keccak and sha3, it is stored. maybe an enum is better.

icicle/include/icicle/utils/modifiers.h
```diff
@@ -26,3 +26,9 @@
 #define __host__
 #define __device__
 #endif
```
mickeyasa, line 28: we are not compatible with windows anyway.

icicle/tests/test_hash_api.cpp
```diff
@@ -134,6 +247,7 @@ class HashSumBackend : public HashBackend
 
 TEST_F(HashApiTest, MerkleTree)
 {
+  ICICLE_CHECK(icicle_set_device(s_reference_target)); // TODO CUDA too
```
mickeyasa, line 250: what about the TODO

icicle/tests/test_hash_api.cpp
```diff
@@ -61,34 +68,140 @@ class HashApiTest : public ::testing::Test
...
+  ICICLE_CHECK(keccakCUDA.hash(d_input, chunk_size, config, d_output));
+  END_TIMER(cuda_timer_device_mem, "CUDA Keccak large time (on device memory)", true);
+  ICICLE_CHECK(icicle_copy(output_main_case_2.get(), d_output, output_size * config.batch));
+  ASSERT_EQ(0, memcmp(output_main_case_2.get(), output_ref.get(), output_size * config.batch));
 }
 
-// TODO: add tests for all hashes
+TEST_F(HashApiTest, sha3)
+{
+  auto config = default_hash_config();
```
mickeyasa, line 179: why are we testing it only on CPU?

icicle/include/icicle/utils/modifiers.h
yshekel: I know but the source was like that. Do you prefer deleting it? no way we support windows?

icicle/tests/test_hash_api.cpp
yshekel: there is no merkle tree for GPU yet so the TODO is for Roman I guess :)

mickeyasa (approved):
maybe I didn;t understand but tests should also run on GPU

icicle/tests/test_hash_api.cpp
yshekel: we test all devices including CUDA in the loop below

PR #626, opened 2024-10-10
Emir/poseidon 

emirsoyturk:


icicle/include/icicle/hash/poseidon.h
```diff
@@ -16,8 +16,43 @@ namespace icicle {
   // Poseidon constants are being initialized.
   template <typename S>
   struct PoseidonConstantsInitOptions {
-    // TODO: Define the struct with fields such as arity, alpha, nof_rounds, mds_matrix, etc.
+    unsigned int  arity;                  ///< Arity of a hash (number of inputs of the single hash).
+    unsigned int  alpha;                  ///< Sbox power.
+    unsigned int  partial_rounds;         ///< Number of upper full rounds of a single hash.
+    unsigned int  full_rounds_half;       ///< Number of partial rounds of a single hash.
+    S*            round_constants;        ///< Round constants (both of the full and the partial rounds). The order of the constants in the memory is according to the rounds order.
```
yshekel, line 23: is the matrices dims known and fixed? if yes, can't we simply define them and then copy the entire struct to GPU as is without an additional indirection

icicle/tests/test_hash_api.cpp
```diff
@@ -165,32 +165,1090 @@ TEST_F(HashApiTest, MerkleTree)
...
+  auto input = std::make_unique<scalar_t[]>(arity);
+  scalar_t output = scalar_t::from(0);
+  // Randomize the input array
+  input[0] = scalar_t::from(1);
+  input[1] = scalar_t::from(1);
+  input[2] = scalar_t::from(1);
+
+  PoseidonConstantsInitOptions<scalar_t> options = {};
+
+  unsigned char poseidon_constants_2[] = {
```
yshekel, line 212: can you maybe randomize it? if we compare CPU to CUDA it should work with any constants right? it should be a more robust test too.

PR #633, opened 2024-10-22
perf: window-method for EC scalar multiplication

yshekel:


icicle/include/icicle/curves/projective.h
```diff
@@ -178,13 +179,32 @@ class Projective
 
   friend HOST_DEVICE Projective operator*(SCALAR_FF scalar, const Projective& point)
   {
+    // Precompute points: P, 2P, ..., (2^window_size - 1)P
+    constexpr unsigned window_size =
+      4; // 4 seems fastest. Optimum is minimizing EC add and depends on the field size. for 256b it's 4.
+    constexpr unsigned table_size = (1 << window_size); // 2^window_size
+    std::array<Projective, table_size> table;
+    table[0] = point;
+    for (int i = 1; i < table_size; ++i) {
```
mickeyasa, line 188: you can run until table_size-1

icicle/include/icicle/curves/projective.h
```diff
@@ -178,13 +179,32 @@ class Projective
 
   friend HOST_DEVICE Projective operator*(SCALAR_FF scalar, const Projective& point)
   {
+    // Precompute points: P, 2P, ..., (2^window_size - 1)P
+    constexpr unsigned window_size =
+      4; // 4 seems fastest. Optimum is minimizing EC add and depends on the field size. for 256b it's 4.
+    constexpr unsigned table_size = (1 << window_size); // 2^window_size
+    std::array<Projective, table_size> table;
```
mickeyasa, line 186: table_size-1

icicle/include/icicle/curves/projective.h
```diff
@@ -178,13 +179,32 @@ class Projective
...
-    UNROLL
-#endif
-    for (int i = 0; i < SCALAR_FF::NBITS; i++) {
-      if (i > 0) { res = dbl(res); }
-      if (scalar.get_scalar_digit(SCALAR_FF::NBITS - i - 1, 1)) { res = res + point; }
+
+    const int nof_windows = (SCALAR_FF::NBITS + window_size - 1) / window_size;
+    for (int w = nof_windows - 1; w >= 0; w -= 1) {
+      // Extract the next window_size bits from the scalar
+      unsigned window = scalar.get_scalar_digit(w, window_size);
```
mickeyasa, line 197: isn't it more efficient to change scalar via << instead of calculating the bits every time?

mickeyasa (approved):
worth checking the optimzations I mentioned before pushing

icicle/include/icicle/curves/projective.h
```diff
@@ -178,13 +179,32 @@ class Projective
 
   friend HOST_DEVICE Projective operator*(SCALAR_FF scalar, const Projective& point)
   {
+    // Precompute points: P, 2P, ..., (2^window_size - 1)P
+    constexpr unsigned window_size =
+      4; // 4 seems fastest. Optimum is minimizing EC add and depends on the field size. for 256b it's 4.
```
Koren-Brand, line 184: Should it be part of the curve config? I assume the difference between 256 and 384 curves is significant

icicle/include/icicle/curves/projective.h
yshekel: It is but I tried and it didn't seem to make any difference overall so I reused this code.

icicle/include/icicle/curves/projective.h
yshekel: I used this and both are optimized for 4:

2^w + (256/w) * (2^w-1) / (2^w)
2^w + (384/w) * (2^w-1) / (2^w)

w - the window bit size

icicle/include/icicle/curves/projective.h
```diff
@@ -178,13 +179,32 @@ class Projective
...
-      if (scalar.get_scalar_digit(SCALAR_FF::NBITS - i - 1, 1)) { res = res + point; }
+
+    const int nof_windows = (SCALAR_FF::NBITS + window_size - 1) / window_size;
+    for (int w = nof_windows - 1; w >= 0; w -= 1) {
+      // Extract the next window_size bits from the scalar
+      unsigned window = scalar.get_scalar_digit(w, window_size);
+
+      // Double the result window_size times
+      for (int j = 0; j < window_size; ++j) {
+        res = dbl(res); // Point doubling
```
Koren-Brand, line 201: Maybe add some condition to skip it the first time when result is sill zero?

icicle/tests/test_curve_api.cpp
```diff
@@ -189,40 +190,21 @@ TEST_F(CurveApiTest, ecntt)
...
-  // }
-  // const NTTDir dir = static_cast<NTTDir>(rand() % 2); // 0: forward, 1: inverse
-  // const int log_coset_stride = rand() % 3;
-  // scalar_t coset_gen;
-  // if (log_coset_stride) {
-  //   coset_gen = scalar_t::omega(logn + log_coset_stride);
-  // } else {
-  //   coset_gen = scalar_t::one();
-  // }
+  // (TODO) Randomize configuration
```
Koren-Brand, line 205: You missed a TODO

icicle/tests/test_curve_api.cpp
```diff
@@ -236,57 +218,55 @@ TEST_F(CurveApiTest, ecntt_new)
...
-  // run(s_ref_target, out_ref.get(), "ecntt", VERBOSE /*=measure*/, 10 /*=iters*/);
-  // run(s_main_target, out_main.get(), "ecntt", VERBOSE /*=measure*/, 10 /*=iters*/);
-  // ASSERT_EQ(0, memcmp(out_main.get(), out_ref.get(), total_size * sizeof(scalar_t))); // FIXME
+
+  run(s_main_target, out_main.get(), "ecntt", false /*=measure*/, 1 /*=iters*/); // warmup
+  run(s_ref_target, out_ref.get(), "ecntt", VERBOSE /*=measure*/, 1 /*=iters*/);
+  run(s_main_target, out_main.get(), "ecntt", VERBOSE /*=measure*/, 1 /*=iters*/);
+  // note that memcmp is tricky here because projetive points can have many representations
+  for (uint64_t i = 0; i < N; ++i) {
+    ASSERT_FALSE(projective_t::is_zero(out_ref[i]));
```
Koren-Brand, line 267: Is there a chance for a valid calculation to result in zero?

icicle/tests/test_curve_api.cpp
```diff
@@ -319,6 +300,26 @@ TYPED_TEST(CurveSanity, CurveSanityTest)
...
+
+  auto expected_mult = TypeParam::zero();
+  START_TIMER(ref)
+  for (int i = 0; i < scalar_t::NBITS; i++) {
+    if (i > 0) { expected_mult = TypeParam::dbl(expected_mult); }
+    if (scalar.get_scalar_digit(scalar_t::NBITS - i - 1, 1)) { expected_mult = expected_mult + point; }
+  }
+  END_TIMER(ref, "scalar mult double-and-add", true);
+
+  ASSERT_EQ(mult, expected_mult);
```
Koren-Brand, line 320: Is there a comparison to be made between CPU and GPU like the other tests?

icicle/include/icicle/curves/projective.h
yshekel: good point

icicle/tests/test_curve_api.cpp
yshekel: It's for Shanie, I just removed the commented out code she left

icicle/tests/test_curve_api.cpp
yshekel: Input is 0 so if we do nothing inside the backend the tests passes otherwise so I want to avoid this case.

icicle/tests/test_curve_api.cpp
yshekel: no it's just a test for me to be sure I did not mess the logic. Because both CUDA and CPU use the same source code for EC*scalar so they will agree even if the computation is wrong.

Then it was useful to compare to the old method perfromance.

PR #637, opened 2024-10-29
add tests for device merkle tree

ChickenLover:
cuda-backend-branch: merkle_tree


yshekel:
I think you should also make the existing tests (currently CPU only) execute on CUDA backend too (C++, rust and go tests). Should be very simple.

ChickenLover:
@yshekel I agree. Added some device tests to rust and go. Also added cuda support to one of our cpu tests in c++

yshekel (approved):

yshekel:
I updated the rust example and C++ tests to test all devices

LeonHibnik (approved):

PR #645, opened 2024-11-06
support column batch for polynomial division CPU backend

ShanieWinitz:
cpu_poly_divide with columns_batch=true fixed

### DIFF
--- a/icicle/backend/cpu/src/field/cpu_vec_ops.cpp
+++ b/icicle/backend/cpu/src/field/cpu_vec_ops.cpp
@@ -803,12 +803,19 @@ REGISTER_SLICE_EXT_FIELD_BACKEND("CPU", cpu_slice<extension_t>);
 
 /*********************************** Highest non-zero idx ***********************************/
 template <typename T>
-eIcicleError cpu_highest_non_zero_idx(
-  const Device& device, const T* input, uint64_t size, const VecOpsConfig& config, int64_t* out_idx /*OUT*/)
+eIcicleError cpu_highest_non_zero_idx_internal(
+  const Device& device,
+  const T* input,
+  uint64_t size,
+  const VecOpsConfig& config,
+  int64_t* out_idx /*OUT*/,
+  int32_t idx_in_batch_to_calc)
 {
   ICICLE_ASSERT(input && out_idx && size != 0) << "Error: Invalid argument";
   uint64_t stride = config.columns_batch ? config.batch_size : 1;
-  for (uint64_t idx_in_batch = 0; idx_in_batch < config.batch_size; ++idx_in_batch) {
+  uint32_t start_idx = (idx_in_batch_to_calc == -1) ? 0 : idx_in_batch_to_calc;
+  uint32_t end_idx = (idx_in_batch_to_calc == -1) ? config.batch_size : idx_in_batch_to_calc + 1;
+  for (uint64_t idx_in_batch = start_idx; idx_in_batch < end_idx; ++idx_in_batch) {
     out_idx[idx_in_batch] = -1; // zero vector is considered '-1' since 0 would be zero in vec[0]
     const T* curr_input =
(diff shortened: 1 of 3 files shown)


icicle/backend/cpu/src/field/cpu_vec_ops.cpp
```diff
@@ -803,12 +803,18 @@ REGISTER_SLICE_EXT_FIELD_BACKEND("CPU", cpu_slice<extension_t>);
 
 /*********************************** Highest non-zero idx ***********************************/
 template <typename T>
-eIcicleError cpu_highest_non_zero_idx(
-  const Device& device, const T* input, uint64_t size, const VecOpsConfig& config, int64_t* out_idx /*OUT*/)
+eIcicleError cpu_highest_non_zero_index(
```
yshekel, line 806: I would rename this to "internal" or maybe prefix with underscore rather than replace idx with index since it's confusing for readers who are not aware of it.

icicle/backend/cpu/src/field/cpu_vec_ops.cpp
```diff
@@ -803,12 +803,18 @@ REGISTER_SLICE_EXT_FIELD_BACKEND("CPU", cpu_slice<extension_t>);
...
+  const T* input,
+  uint64_t size,
+  const VecOpsConfig& config,
+  int64_t* out_idx /*OUT*/,
+  int32_t idx_in_batch_to_calc)
 {
   ICICLE_ASSERT(input && out_idx && size != 0) << "Error: Invalid argument";
   uint64_t stride = config.columns_batch ? config.batch_size : 1;
   for (uint64_t idx_in_batch = 0; idx_in_batch < config.batch_size; ++idx_in_batch) {
+    if (idx_in_batch_to_calc != -1 && idx_in_batch_to_calc != idx_in_batch) { continue; }
```
yshekel, line 817: better compute the start->end indices of the loop to be either a specific index or 0 to batch_size instead of doing many iterations and skipping all except one.

icicle/tests/test_polynomial_api.cpp
```diff
@@ -137,395 +131,446 @@ class PolynomialTest : public ::testing::Test
 
 TEST_F(PolynomialTest, evaluation)
 {
-  const scalar_t coeffs[3] = {one, two, three};
-  auto f = Polynomial_t::from_coefficients(coeffs, 3);
-  scalar_t x = scalar_t::rand_host();
+  for (auto device : s_registered_devices) {
```
yshekel, line 134: In this test you only use a loop to test all devices right?

yshekel (approved):

icicle/tests/test_field_api.cpp
```diff
@@ -805,34 +807,61 @@ TEST_F(FieldApiTestBase, polynomialDivision)
...
 
   for (auto device : s_registered_devices) {
     ICICLE_CHECK(icicle_set_device(device));
     for (int columns_batch = 0; columns_batch <= 1; columns_batch++) {
-      ICICLE_LOG_DEBUG << "testing polynomial division on device " << device << " [column_batch=" << columns_batch
-                       << "]";
+      ICICLE_LOG_INFO << "testing polynomial division on device " << device << " [column_batch=" << columns_batch
+                      << "]";
+
+      // Debug: Initialize each input vector in the batch with the same ascending values
```
HadarIngonyama, line 819: remove comments before merge

icicle/tests/test_field_api.cpp
```diff
@@ -797,6 +797,8 @@ TEST_F(FieldApiTestBase, polynomialEval)
 
 TEST_F(FieldApiTestBase, polynomialDivision)
 {
+  int seed = time(0);
+  srand(seed);
   const uint64_t numerator_size = 1 << 4;
```
HadarIngonyama, line 802: why aren't the sizes randomized like in the other tests?

HadarIngonyama (approved):

PR #647, opened 2024-11-06
Support padding for the leaves of the merkle tree 

mickeyasa:
Support padding for the leaves of the merkle tree


mickeyasa (commented):
Support padding for the leaves of the merkle tree

icicle/tests/test_hash_api.cpp
```diff
@@ -315,54 +322,216 @@ class HashSumBackend : public HashBackend
   }
 };
 
-void assert_valid_tree(
+/**
```
yshekel, line 325: why do you need doxygen documentation for tests? also the doc string is wrong (missing leaf-size).

icicle/tests/test_hash_api.cpp
```diff
@@ -315,54 +322,216 @@ class HashSumBackend : public HashBackend
...
+      std::vector<std::byte> last_leaf(leaf_size);
+      memcpy(last_leaf.data(), inputs + input_size - leaf_size, leaf_size);
+      int nof_leaves_in_padding = (tree_input_size - input_size) / leaf_size;
+      for (int i = 0; i < nof_leaves_in_padding; i++) {
+        memcpy(input_vec.data() + input_size + i * leaf_size, last_leaf.data(), leaf_size);
+      }
+    }
+  }
+
+  // std::cout << "Padded inputs:\n";
```
yshekel, line 369: remove

icicle/tests/test_hash_api.cpp
```diff
@@ -315,54 +322,216 @@ class HashSumBackend : public HashBackend
...
   auto [root, root_size] = tree.get_merkle_root();
-
   for (int i = 0; i < root_size; i++) {
-    ASSERT_EQ(root[i], layer_out[i]) << "Mismatch in root[" << i << "]";
+    if (root[i] != layer_out[i]) { return false; }
   }
+  return true;
 }
 
+/**
```
yshekel, line 418: same about the doxygen doc

icicle/tests/test_hash_api.cpp
```diff
@@ -315,54 +322,216 @@ class HashSumBackend : public HashBackend
...
+ *                             test will fail if this value isn't default (1) and T != std::byte
+ */
+template <typename T>
+void test_merkle_tree(
+  const std::vector<Hash>& hashes,
+  const MerkleTreeConfig& config,
+  const int output_store_min_layer,
+  int nof_leaves,
+  const T* leaves,
+  unsigned explicit_leaf_size = 1)
```
yshekel, line 457: why default value?

icicle/tests/test_hash_api.cpp
```diff
@@ -315,54 +322,216 @@ class HashSumBackend : public HashBackend
...
+
+  // assert that incorrect size fails
+  if (config.padding_policy == PaddingPolicy::None) {
+    ASSERT_NE(prover_tree.build(leaves, nof_leaves * explicit_leaf_size - 1, config), eIcicleError::SUCCESS);
+    ASSERT_NE(prover_tree.build(leaves, nof_leaves * explicit_leaf_size + 1, config), eIcicleError::SUCCESS);
+  }
+  // build tree
+  START_TIMER(MerkleTree_build)
+  ICICLE_CHECK(prover_tree.build(leaves, nof_leaves * explicit_leaf_size, config));
+  END_TIMER(MerkleTree_build, "Merkle Tree CPU", true)
```
yshekel, line 474: is it CPU always? you didn't set the device so it can be anything here

icicle/tests/test_hash_api.cpp
```diff
@@ -628,15 +913,25 @@ TEST_F(HashApiTest, MerkleTreeLeavesOnDeviceTreeOnHost)
   }
 }
 
+void deep_copy_byte_array_to_vec(const std::byte* src, std::size_t size, std::vector<std::byte>& dest)
```
yshekel, line 916: There is already a simpler way to do that : 
'auto dest = std::vector<std::byte>(src, src + size);'

icicle/tests/test_hash_api.cpp
```diff
@@ -391,48 +560,209 @@ TEST_F(HashApiTest, MerkleTreeBasic)
...
 
-  // build tree
-  ICICLE_CHECK(prover_tree.build(leaves, nof_leaves, config));
-  assert_valid_tree<uint32_t>(prover_tree, nof_leaves, leaves, hashes, config);
+TEST_F(HashApiTest, MerkleTreeZeroPadding)
+{
+  const int leaf_size = sizeof(uint32_t);
+  const int nof_leaves = 100;
+  uint32_t leaves[nof_leaves];
+  // randomize(leaves, nof_leaves);
```
yshekel, line 571: should you use randomize instead?

icicle/tests/test_hash_api.cpp
```diff
@@ -391,48 +560,209 @@ TEST_F(HashApiTest, MerkleTreeBasic)
...
 
-  auto prover_tree = MerkleTree::create(hashes, sizeof(uint32_t), output_store_min_layer);
-  auto verifier_tree = MerkleTree::create(hashes, sizeof(uint32_t), output_store_min_layer);
+  test_merkle_tree(hashes, config, output_store_min_layer, nof_leaves, leaves);
+}
 
-  // build tree
-  ICICLE_CHECK(prover_tree.build(leaves, nof_leaves, config));
-  assert_valid_tree<uint32_t>(prover_tree, nof_leaves, leaves, hashes, config);
+TEST_F(HashApiTest, MerkleTreeZeroPadding)
```
yshekel, line 566: you need to test it for all devices with a loop like MerkleTreeMixMediumSize
This is true for all tests except for where you use the dummy HashSum since it's only CPU

icicle/tests/test_hash_api.cpp
```diff
@@ -391,48 +560,209 @@ TEST_F(HashApiTest, MerkleTreeBasic)
...
+
+  // Test invalid proof (By modifying random data in the leaves)
+  verification_valid = true;
+  ICICLE_CHECK(prover_tree.get_merkle_proof(wrong_bytes.get(), byte_size, leaf_idx, true, config, merkle_proof));
+  ICICLE_CHECK(verifier_tree.verify(merkle_proof, verification_valid));
+  ASSERT_FALSE(verification_valid) << "Pruned proof of invalid inputs at index " << leaf_idx
+                                   << " is valid (And should be invalid).";
+}
+
+TEST_F(HashApiTest, MerkleTreeLastValuePadding)
```
yshekel, line 689: need to test all devices

icicle/tests/test_hash_api.cpp
```diff
@@ -841,4 +1152,87 @@ TEST_F(HashApiTest, poseidon3_batch)
...
+
+  std::vector<std::vector<std::byte>> device_roots(s_registered_devices.size());
+  int device_roots_idx = 0;
+
+  for (int i = 0; i < s_registered_devices.size(); i++) {
+    const auto& device = s_registered_devices[i];
+    ICICLE_LOG_INFO << "MerkleTreeDeviceBig on device=" << device;
+    ICICLE_CHECK(icicle_set_device(device));
+
+    // Create a Keccak256 hasher with an arity of 2: every 64B -> 32B
```
yshekel, line 1177: wrong comment

icicle/tests/test_hash_api.cpp
```diff
@@ -841,4 +1152,87 @@ TEST_F(HashApiTest, poseidon3_batch)
   ASSERT_EQ(0, memcmp(output_cpu.get(), output_cuda.get(), config.batch * sizeof(scalar_t)));
 }
 
-#endif // POSEIDON
\ No newline at end of file
+TEST_F(HashApiTest, poseidon_tree)
```
yshekel, line 1155: why don't you use the 'test_merkle_tree' here?

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
```diff
@@ -40,42 +40,41 @@ namespace icicle {
...
+          m_full_path_size += cur_layer_input_size;
         }
       }
     }
 
-    // TODO: handle size
+    ~CPUMerkleTreeBackend()
+    {
+      for (auto& pair : m_map_segment_id_2_inputs) {
+        delete pair.second; // Ensure all dynamically allocated memory is freed
```
yshekel, line 56: modern C++ should avoid new/malloc+delete/free and use smart pointers anyway. This would avoid the leak and also works when exceptions are thrown.
There is usually no reason to use malloc/new

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
```diff
@@ -40,42 +40,41 @@ namespace icicle {
...
-                         << leaves_size << ". Note: Padding is currently not supported but will be soon";
-        return eIcicleError::INVALID_ARGUMENT;
-      }
-      if (leaves_size > expected_input_size) {
-        ICICLE_LOG_ERROR << "CPU Merkle tree: Expecting " << expected_input_size << " bytes in input, got "
-                         << leaves_size << ". Leaves size cannot exceeds tree size.";
+      // build a vector with the leaves that needs padding.
+      std::vector<std::byte> padded_leaves;
+      if (!init_layers_db(config, leaves_size) || !init_padded_leaves(padded_leaves, leaves, leaves_size, config)) {
         return eIcicleError::INVALID_ARGUMENT;
```
yshekel, line 70: why did you remove the ICICLE_LOG_ERROR? every time you return an error you'd better log the error, so the user can understand how to fix his code

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
```diff
@@ -95,7 +94,11 @@ namespace icicle {
 
           // delete completed_segment_id from the map
           const uint64_t completed_segment_id = completed_segment_idx ^ (completed_layer_idx << 56);
-          m_map_segment_id_2_inputs.erase(completed_segment_id);
+          auto segment = m_map_segment_id_2_inputs.find(completed_segment_id);
+          if (segment != m_map_segment_id_2_inputs.end()) {
+            delete segment->second;
```
yshekel, line 99: same comment here. With smart pointers you only need to erase

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
mickeyasa: thre error messages remain. They moved inside the functions init_layers_db, init_padded_leaves

icicle/tests/test_hash_api.cpp
Koren-Brand: Because it's not the test itself but a function used by the tests so I added documentation to it
I'll add leaf size to the doc

icicle/tests/test_hash_api.cpp
Koren-Brand: Nope, it's a remanent from a previous version of the test file (when it was for CPU only). I will remove it.

icicle/tests/test_hash_api.cpp
Koren-Brand: Didn't know - will do!

icicle/tests/test_hash_api.cpp
Koren-Brand: Correct me if I'm wrong but padding is not yet supported on GPU
But I do need to add a set_device to reference_device (CPU) in those tests

icicle/tests/test_hash_api.cpp
Koren-Brand: I can add it but here there is already comparison between cpu and gpu. I felt that adding the reference tree test redundant.

icicle/tests/test_hash_api.cpp
Koren-Brand: This value is only needed in cases where the actual leave size is different than sizeof(T) (For example MerkleTreeLarge test) to avoid having to fill it everytime, a default value was set. Given that this value is used multiplicatively inside the function, a value of 1 is set to be the default.

icicle/tests/test_hash_api.cpp
Koren-Brand: Again I think it is not yet implemented on GPU

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
mickeyasa: fixed

icicle/backend/cpu/src/hash/cpu_merkle_tree.cpp
mickeyasa: fixed

icicle/CMakeLists.txt
```diff
@@ -18,6 +18,11 @@ endif()
 # Print the selected build type
 message(STATUS "Build type: ${CMAKE_BUILD_TYPE}")
 
+# Prevent build if both SANITIZE and CUDA_BACKEND are enabled
+if(SANITIZE AND CUDA_BACKEND)
```
danny-shterman, line 22: Why SANITIZE AND CUDA_BACKEND couldn't both be true?
What happens in tests that CPU result if compared against GPU?

mickeyasa: sanitizer and cuda does not work together

danny-shterman (approved):

Koren-Brand (approved):

mickeyasa:
Support padding for the leaves of the merkle tree

PR #652, opened 2024-11-10
verify and assert correct t-parameter for Poseidon hasher

yshekel:


icicle/src/hash/poseidon.cpp
```diff
@@ -9,6 +9,12 @@ namespace icicle {
   template <>
   Hash create_poseidon_hash<scalar_t>(unsigned t, const scalar_t* domain_tag)
   {
+    // Assert that t is valid. Ideally would like to return an eIcicleError but the API doesn't let us do it
+    constexpr std::array<int, 4> validTValues = {3, 5, 9, 12};
+    const bool is_valid_t = std::find(validTValues.begin(), validTValues.end(), t) != validTValues.end();
+    ICICLE_LOG_ERROR << "Poseidon only supports t values of 3, 5, 9, or 12.";
```
mickeyasa, line 15: why do you always print that

mickeyasa (changes requested):

icicle/src/hash/poseidon.cpp
yshekel: excellent question :)

yshekel: fixed

mickeyasa (approved):

danny-shterman (approved):

PR #654, opened 2024-11-11
updates for metal backend

yshekel:
- add script to pull metal backend
- update cmake to build metal backend
- update rust wrappers with feature to build metal backend
- update examples to support any backend
- updated C++ test suites to share logic for selecting main device


icicle/CMakeLists.txt
```diff
@@ -37,7 +37,9 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 # Build options
 option(BUILD_TESTS "Build unit tests. Default=OFF" OFF)
 option(CPU_BACKEND "Build CPU backend. Default=ON" ON)
+# TODO Yuval: consider decoupling backends from frontend build. Is that worth the effort?
```
mickeyasa, line 40: is it?
Over time the cpu backend compilation time will be less efficient

icicle/CMakeLists.txt
```diff
@@ -130,6 +132,31 @@ if (CUDA_BACKEND)
     endif()
 endif()
 
+if (METAL_BACKEND)
+  string(TOLOWER "${METAL_BACKEND}" CUDA_BACKEND_LOWER)
+  if (CUDA_BACKEND_LOWER STREQUAL "local")
```
mickeyasa, line 137: why metal cares about cuda?

mickeyasa (approved):

icicle/CMakeLists.txt
yshekel: good catch

icicle/CMakeLists.txt
yshekel: I referred to cuda and metal. I did not intend to split the CPU from this repo or build it separately.
We can discuss it but it's not related to this PR.
About metal and cuda, we could separate them entirely from the frontend cmake but it will complicate the backend. I am not sure if it is really worth it at this point although it is probably better to decouple the backends build entirely if we can.

scripts/pull_cuda_backend.sh
```diff
@@ -1,24 +1,41 @@
...
+# Check if BACKEND_DIR exists
+if [ ! -d "${BACKEND_DIR}" ]; then
+    echo "Error: Directory '${BACKEND_DIR}' does not exist."
+    exit 1
+fi
+# Get the absolute path of the backend directory
+ABS_CUDA_DIR=$(realpath ${BACKEND_DIR})/cuda
 
-    echo "Trying to pull CUDA backend commit '${CUDA_BACKEND}' to '${ABS_BACKEND_DIR}'"
+echo "Trying to pull Metal backend commit '${CUDA_BACKEND}' to '${ABS_CUDA_DIR}'"
```
ShanieWinitz, line 14: Probably should be CUDA here

yshekel: fixed

jeremyfelder (approved):

wrappers/rust/icicle-runtime/Cargo.toml
```diff
@@ -9,10 +9,16 @@ repository.workspace = true
...
 cmake = "0.1.50"
 
 [features]
+# Enabling backend features means building them with the ICICLE frontend.
+# Sources must be either local (under icicle/backend) or pulled during the build process.
+# Note: Backend repositories are private and not meant for user workflows. Users should install backends directly instead.
 cuda_backend = []
 pull_cuda_backend = []
+metal_backend = []
+pull_metal_backend = []
```
jeremyfelder, line 24: Is there a way to prevent these features from being used outside of tests, examples, or debug builds?

yshekel: not that I am aware of, if you have a solution that would be great

.github/workflows/cpp.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       uses: actions/checkout@v4
     - name: Check clang-format
       if: needs.check-changed-files.outputs.cpp == 'true'
-      run: if [[ $(find ./ \( -path ./icicle/build -prune -o -path ./**/target -prune -o -path ./examples -prune \) -iname *.h -or -iname *.cuh -or -iname *.cu -or -iname *.c -or -iname *.cpp | xargs clang-format --dry-run -ferror-limit=1 -style=file 2>&1) ]]; then echo "Please run clang-format"; exit 1; fi
+      run: ./scripts/format_all.sh . --check
```
jeremyfelder, line 28: Can you also update the [hooks script](https://github.com/ingonyama-zk/icicle/blob/main/scripts/hooks/pre-push#L13-L28) to use this newer script

yshekel: done

PR #664, opened 2024-11-24
Program-support

mickeyasa:
## Describe the changes

This PR...

## Linked Issues

Resolves #


icicle/include/icicle/program/symbol.h
```diff
@@ -0,0 +1,160 @@
+#pragma once
+
+#include <memory>
+#include <vector>
+
+namespace icicle {
+
+  /**
+   * @brief Enum to represent all the operation supported by the Symbol class.
```
danny-shterman, line 9: operations

icicle/include/icicle/program/symbol.h
```diff
@@ -0,0 +1,160 @@
...
+
+    OP_INPUT,
+    OP_CONST,
+  };
+
+  /**
+   * @brief a node at the data-flow grapsh (DFG) that represent a single operation
+   *
+   * This class contains a data about an operation done by the Symbol class.
+   * It saves oll the data to describe the operation done and point to its operands.
```
danny-shterman, line 28: points

icicle/include/icicle/program/symbol.h
```diff
@@ -0,0 +1,160 @@
...
+   * At any given time a Sybol contains an Operation instance that represent all the
```
danny-shterman, line 29: Symbol

icicle/include/icicle/program/symbol.h
```diff
@@ -0,0 +1,160 @@
...
+
+    // Constructor
+    Operation<S>(
+      OpCode opcode,
+      std::shared_ptr<Operation<S>> operand1,
+      std::shared_ptr<Operation<S>> operand2 = nullptr,
+      std::unique_ptr<S> constant = nullptr,
+      int mem_addr = -1)
+        : m_opcode(opcode), m_operand1(operand1), m_operand2(operand2), m_mem_addr(mem_addr),
+          m_constant(std::move(constant))
```
danny-shterman, line 55: why not just
m_constant(constant)
?

icicle/include/icicle/program/program.h
```diff
@@ -0,0 +1,140 @@
...
+        break;
+      default:
+        ICICLE_LOG_ERROR << "Illegal opcode: " << int(pre_def);
+      }
+      m_nof_outputs = 1;
+      int instruction = int(OpCode::NOF_OPERATIONS) + int(pre_def);
+      m_instructions.push_back(instruction);
+    }
+
+    // run over all symbols at the vector and set there gen process to OP_INPUT
```
danny-shterman, line 55: inputs instead of symbols?
change "there" to "their"

icicle/backend/cpu/src/field/cpu_vec_ops.cpp
```diff
@@ -368,7 +368,7 @@ class VectorOpTask : public TaskBase
 public:
   T m_intermidiate_res;    // pointer to the output. Can be a vector or scalar pointer
   uint64_t m_idx_in_batch; // index in the batch. Used in intermediate res tasks
-};                         // class VectorOpTask
+}; // class VectorOpTask
```
yshekel, line 371: this so annoying but I think that mac and linux format it differently and this is why we face those issues all the time

icicle/include/icicle/program/symbol.h
```diff
@@ -0,0 +1,160 @@
...
+#include <memory>
+#include <vector>
+
+namespace icicle {
+
+  /**
+   * @brief Enum to represent all the operation supported by the Symbol class.
+   *
+   */
+  enum OpCode {
```
yshekel, line 12: maybe once more namespace or rename? it's maybe a little too general and might conflict in the future

icicle/include/icicle/program/symbol.h
```diff
@@ -0,0 +1,160 @@
...
+   * At any given time a Sybol contains an Operation instance that represent all the
+   * calculation history done so far to reach the current value.
+   */
+
+  template <typename S>
+  class Operation
+  {
+  public:
+    OpCode m_opcode;
+    std::shared_ptr<Operation<S>> m_operand1; // 1st operand if exist
```
yshekel, line 38: (1) does this mean we won't ever have ops with more than 2 operands?
(2) this template won't support cases like op1 is scalar and op2 is EC point. Is that intentional? not interesting?

icicle/include/icicle/program/symbol.h
```diff
@@ -0,0 +1,160 @@
...
+    std::shared_ptr<Operation<S>> m_operand2; // 2nd operand if exist
+
+    // optinal parameters:
+    std::unique_ptr<S> m_constant; // for OP_CONST: const value
+
+    // implementation:
+    int m_mem_addr; // location at the input vectors
```
yshekel, line 45: should this be here? this seems like it is coupling this class to something else. Also int may not be enough for large data. probably u64 is better

icicle/include/icicle/program/symbol.h
yshekel: you cannot copy unique_ptr, since they have a single owner so you must move

icicle/include/icicle/program/symbol.h
```diff
@@ -0,0 +1,160 @@
...
+      std::shared_ptr<Operation<S>> operand1,
+      std::shared_ptr<Operation<S>> operand2 = nullptr,
+      std::unique_ptr<S> constant = nullptr,
+      int mem_addr = -1)
+        : m_opcode(opcode), m_operand1(operand1), m_operand2(operand2), m_mem_addr(mem_addr),
+          m_constant(std::move(constant))
+    {
+    }
+
+    bool was_visited(bool set_as_visit)
```
yshekel, line 59: maybe is_visited? all this grammar in code is sometimes weird.
BTW this is reminding me something :)

icicle/include/icicle/program/symbol.h
```diff
@@ -0,0 +1,160 @@
...
+
+  /**
+   * @brief The basic variable used to describe a program.
+   *
+   * This class is used by the end user to describe a functionality later on provided
+   * to execute by different backends.
+   * This enables the user to constructs lamda.
+   */
+  template <typename S>
+  class Symbol
```
yshekel, line 81: consider rename to SymbolicVariable or SymVar since symbol is typically the name for variables and functions in the context of linkage

icicle/include/icicle/program/program.h
```diff
@@ -0,0 +1,140 @@
...
+
+#include <vector>
+#include <functional>
+#include "icicle/program/symbol.h"
+
+namespace icicle {
+
+  using InstructionType = uint32_t;
+
+  enum PreDefinedPrograms { AB_MINUS_C, EQ_X_AB_MINUS_C };
```
yshekel, line 11: why EQ_X but AB? I mean maybe it's E times Q? the name convention is not really clear. I do understand from the context but still

icicle/backend/cpu/include/cpu_program_executor.h
```diff
@@ -0,0 +1,96 @@
...
+        m_variable_ptrs[variable_ptrs_idx++] = &(m_intermidites[idx]);
+      }
+    }
+
+    // execute the program an return a program to the result
+    void execute()
+    {
+      const std::byte* instruction;
+      for (InstructionType instruction : m_program.m_instructions) {
+        const int func_select = (instruction & 0xFF);
```
danny-shterman, line 32: func_select unused

icicle/include/icicle/program/program.h
```diff
@@ -0,0 +1,140 @@
...
+    void set_as_inputs(std::vector<Symbol<S>>& combine_inputs)
+    {
+      m_nof_inputs = combine_inputs.size();
+      for (int input_idx = 0; input_idx < m_nof_inputs; input_idx++) {
+        combine_inputs[input_idx].set_as_input(input_idx);
+      }
+    }
+
+    // run over the DFG held by result and gemerate the program
+    void generate_program(Symbol<S>& result)
```
yshekel, line 65: I think it would be better to remove this logic from this class.  I would make this class only a container for the compiled format.

In compilers like llvm they typically have passes on the IR (symbols graph here). So one pass will allocate constants for example, another will generate the code, another will print it to file etc. Then a passManager is able to execute all the passes in a correct order based on dependencies and finally last pass is code generation which should fill this program object. 
I think it's a better design be cause you generally don't want to modify this code and risk breaking it when adding new optimizations for example.

icicle/backend/cpu/include/cpu_program_executor.h
```diff
@@ -0,0 +1,96 @@
...
+#include "icicle/program/symbol.h"
+#include "icicle/program/program.h"
+
+namespace icicle {
+
+  template <typename S>
+  class CpuProgramExecutor
+  {
+  public:
+    CpuProgramExecutor(Program<S>& program)
```
yshekel, line 14: I think you need to either take is as const reference or by value. Why would you modify the input program and make it not work on other backends possible?

icicle/backend/cpu/include/cpu_program_executor.h
yshekel, line 32: consider adding macros to extract fields and make the code more readable

icicle/backend/cpu/include/cpu_program_executor.h
```diff
@@ -0,0 +1,96 @@
...
+        const int func_select = (instruction & 0xFF);
+        (this->*m_function_arr[instruction & 0xFF])(instruction);
+      }
+    }
+
+    std::vector<S*> m_variable_ptrs;
+
+  private:
+    Program<S> m_program;
+    std::vector<S> m_intermidites;
```
yshekel, line 41: typo I think? intermediates

danny-shterman (approved):

icicle/include/icicle/program/symbol.h
```diff
@@ -0,0 +1,160 @@
...
+      const bool was_visited = (m_visit_idx == s_last_visit);
+      if (set_as_visit) { m_visit_idx = s_last_visit; }
+      return was_visited;
+    }
+
+    static void reset_visit() { s_last_visit++; }
+
+  private:
+    unsigned int m_visit_idx = 0;
+    static inline unsigned int s_last_visit = 1;
```
danny-shterman, line 70: Just curious. Why it starts with s_ when you always insist that members will start with m_

icicle/include/icicle/program/program.h
yshekel: also this aligns with the fact that you have m_mem_addr inside the symbol where technically the symbol has nothing to do with this. Ideally some pass needs to take care of it and store the mapping itself. Then the final code generation pass would use this information for generating the code.

Alternatively maybe you could add an attributes map to the Symbol object and store general things on it. Typically passes would write this and later passes read it. But anyway coupling symbol to the program implementation should be avoided.

icicle/backend/cpu/include/cpu_program_executor.h
mickeyasa: done

icicle/include/icicle/fields/quartic_extension.h
```diff
@@ -241,7 +241,7 @@ class QuarticExtensionField
       FF::reduce(
         (CONFIG::nonresidue_is_negative
            ? (FF::mul_wide(xs.real, x0) + FF::template mul_unsigned<CONFIG::nonresidue>(FF::mul_wide(xs.im2, x2)))
-           : (FF::mul_wide(xs.real, x0)) - FF::template mul_unsigned<CONFIG::nonresidue>(FF::mul_wide(xs.im2, x2)))),
+           : (FF::mul_wide(xs.real, x0))-FF::template mul_unsigned<CONFIG::nonresidue>(FF::mul_wide(xs.im2, x2)))),
```
LeonHibnik, line 244: formatter changed this?

icicle/include/icicle/program/program.h
```diff
@@ -0,0 +1,140 @@
...
+      std::vector<Symbol<S>> program_inputs(nof_inputs);
+      set_as_inputs(program_inputs);
+      Symbol<S> result = program_func(program_inputs);
+      generate_program(result);
+    }
+
+    // Generate a program based on a PreDefinedPrograms
+    Program(PreDefinedPrograms pre_def)
+    {
+      switch (pre_def) {
```
LeonHibnik, line 40: Is it sufficient to use only m_nof_imputs here to distinguish between cases?

mickeyasa: this are the predefined functions. The functionality is defined at the executors.
Only the number of inputs and outputs are missing

icicle/include/icicle/program/program.h
mickeyasa: done

icicle/include/icicle/program/program.h
mickeyasa: lets discuss

icicle/include/icicle/program/symbol.h
mickeyasa: (1) for now that's more than enough. If we'll need more, can be easily changed to a vector. For now, the operations are very simple
(2) Also here, I see no reason to support. No primitive in the near future that I know of will need that.

icicle/include/icicle/program/symbol.h
mickeyasa: The name might be misleading. 
This is the idx in the variable vector that this operation is going to store the result.
This is relevant to Program and not the executor.

icicle/include/icicle/program/symbol.h
mickeyasa: its a static memeber

icicle/include/icicle/program/symbol.h
mickeyasa: This name is used by the end user so I need somthing short and simple.
I agree Symbol is not good but not sure about your. Please come up with better suggestions.

icicle/include/icicle/fields/quartic_extension.h
mickeyasa: Looks like we have an issue with formatting on mac vs intel.
We need to check

icicle/backend/cpu/src/field/cpu_vec_ops.cpp
mickeyasa: Maybe we need the CI to fix the code and not us?

icicle/include/icicle/fields/quartic_extension.h
LeonHibnik: It's just weird it affected the - sign and not others

icicle/backend/cpu/src/field/cpu_vec_ops.cpp
yshekel: it will have to make a new commit but maybe it's possible

icicle/include/icicle/program/program.h
mickeyasa: added in comment clarification about those macros

icicle/backend/cpu/include/cpu_program_executor.h
mickeyasa: fixed

icicle/backend/cpu/include/cpu_program_executor.h
mickeyasa: fixed

mickeyasa: done

yshekel (approved):

PR #665, opened 2024-11-25
Replace random function to make CI reproducible

Koren-Brand:
## Describe the changes

This PR changes random generation to be based on rand such that bugs could be effectively reproduced.
Ctest (CI) Now outputs test log on test failure

## Linked Issues

Resolves #


icicle/include/icicle/fields/field.h
```diff
@@ -691,12 +691,9 @@ class Field
 
   static HOST_INLINE Field rand_host()
   {
-    std::random_device rd;
-    std::mt19937_64 generator(rd());
-    std::uniform_int_distribution<unsigned> distribution;
     Field value{};
     for (unsigned i = 0; i < TLC; i++)
-      value.limbs_storage.limbs[i] = distribution(generator);
+      value.limbs_storage.limbs[i] = rand();
```
mickeyasa, line 696: the uniform distribution does nothing?

mickeyasa (approved):
good initiative

icicle/include/icicle/fields/field.h
yshekel, line 696: maybe assert that limbs is 4 bytes here because rand() returns int

yshekel (approved):
Approved but please see the comment

icicle/include/icicle/fields/field.h
Koren-Brand: The implementation of limbs_storage in field is uint32 so it is an exact fit

Koren-Brand: Apparently, rand has a maximal value of 2^15 - 1 so another solution might be better

icicle/include/icicle/fields/field.h
```diff
@@ -689,14 +689,18 @@ class Field
     return rv;
   }
 
+  inline static std::mt19937 field_rand_generator = std::mt19937{};
```
yshekel, line 692: maybe better if you moved all of this to utils and use it here from utils.

Koren-Brand: that's a good idea

icicle/include/icicle/fields/field.h
```diff
@@ -24,6 +24,7 @@
 #endif // __CUDACC__
 
 #include "icicle/errors.h"
+#include "icicle/utils/rand_gen.cpp"
```
yshekel, line 27: you need to rename it to rand_gen.h

icicle/include/icicle/utils/rand_gen.cpp
```diff
@@ -0,0 +1,5 @@
+#pragma once
```
yshekel, line 1: it's not a c++ file right? rename to .h

Koren-Brand: changed

PR #667, opened 2024-11-27
fix 404 status URL

wangjingcun:
## Describe the changes

The previous link has expired and has been replaced with the latest address.

## Linked Issues

Resolves #


jeremyfelder:
@wangjingcun Thanks for pointing that out!

The link to the 1.5.1 release is not the correct line. Can you update the link to https://github.com/ingonyama-zk/icicle/blob/a02459c64d33ffe867ec633b771ee9ffca808c29/icicle/appUtils/ntt/ntt.cu#L456; this is the commit the function was introduced in

PR #675, opened 2024-12-05
add a test to check cpu program executor

mickeyasa:
Add a test to check cpu program executor

### DIFF
--- a/icicle/tests/test_field_api.cpp
+++ b/icicle/tests/test_field_api.cpp
@@ -11,6 +11,10 @@
 #include "icicle/utils/log.h"
 #include "icicle/backend/ntt_config.h"
 
+
+#include "icicle/program/symbol.h"
+#include "icicle/program/program.h"
+#include "../../icicle/backend/cpu/include/cpu_program_executor.h"
 #include "test_base.h"
 
 using namespace field_config;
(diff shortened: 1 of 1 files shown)


icicle/tests/test_field_api.cpp
```diff
@@ -11,6 +11,9 @@
 #include "icicle/utils/log.h"
 #include "icicle/backend/ntt_config.h"
 
+#include "icicle/program/symbol.h"
+#include "icicle/program/program.h"
+#include "../../icicle/backend/cpu/include/cpu_program_executor.h"
```
yshekel, line 16: maybe we should move tests like this one to the backend? this is kind of dirty and works for CPU only anyway.
If we define another test executable in the backend then we will test them only when the backend is compiled and avoid this uglyness

icicle/tests/test_field_api.cpp
```diff
@@ -939,6 +942,128 @@ TYPED_TEST(FieldApiTest, ntt)
 }
 #endif // NTT
 
+// define program
+using MlePoly = Symbol<scalar_t>;
+MlePoly combine_func(std::vector<MlePoly>& inputs)
```
yshekel, line 947: why not const?

icicle/tests/test_field_api.cpp
```diff
@@ -939,6 +942,128 @@ TYPED_TEST(FieldApiTest, ntt)
...
+  const MlePoly& A = inputs[0];
+  const MlePoly& B = inputs[1];
+  const MlePoly& C = inputs[2];
+  const MlePoly& EQ = inputs[3];
+  return (EQ * (A * B - C));
+}
+
+TEST_F(FieldApiTestBase, CpuProgramExecutor)
+{
+  int seed = time(0);
```
yshekel, line 958: why not rely on the common test seed that Koren did?

icicle/tests/test_field_api.cpp
```diff
@@ -939,6 +942,128 @@ TYPED_TEST(FieldApiTest, ntt)
...
+  scalar_t::rand_host_many(in_b.get(), total_size);
+  auto in_c = std::make_unique<scalar_t[]>(total_size);
+  scalar_t::rand_host_many(in_c.get(), total_size);
+  auto in_eq = std::make_unique<scalar_t[]>(total_size);
+  scalar_t::rand_host_many(in_eq.get(), total_size);
+
+  //----- element wise operation ----------------------
+  auto out_element_wise = std::make_unique<scalar_t[]>(total_size);
+  std::ostringstream oss;
+  oss << "Strait forward function (Element wise) time: ";
```
yshekel, line 976: typo 'Straight'
BTW no need for string stream here but ok

icicle/tests/test_field_api.cpp
```diff
@@ -939,6 +942,128 @@ TYPED_TEST(FieldApiTest, ntt)
...
+
+  START_TIMER(element_wise_op)
+  for (int i = 0; i < 100000; ++i) {
+    out_element_wise[i] = in_eq[i] * (in_a[i] * in_b[i] - in_c[i]);
+  }
+  // std::cout << "Strait forward function (Element wise) time: ";
```
yshekel, line 982: remove

icicle/tests/test_field_api.cpp
```diff
@@ -939,6 +942,128 @@ TYPED_TEST(FieldApiTest, ntt)
...
+  END_TIMER(element_wise_op, oss.str().c_str(), true);
+
+  //----- written program ----------------------
```
yshekel, line 985: what do you mean written?

icicle/tests/test_field_api.cpp
```diff
@@ -939,6 +942,128 @@ TYPED_TEST(FieldApiTest, ntt)
...
+
+  CpuProgramExecutor<scalar_t> prog_exe_written(program_written);
+  auto out_written_program = std::make_unique<scalar_t[]>(total_size);
+
+  // init program
+  prog_exe_written.m_variable_ptrs[0] = in_a.get();
+  prog_exe_written.m_variable_ptrs[1] = in_b.get();
+  prog_exe_written.m_variable_ptrs[2] = in_c.get();
+  prog_exe_written.m_variable_ptrs[3] = in_eq.get();
+  prog_exe_written.m_variable_ptrs[4] = out_written_program.get();
```
yshekel, line 997: is it implicit that last one is output? what if there are multiple outputs?

icicle/tests/test_field_api.cpp
```diff
@@ -939,6 +942,128 @@ TYPED_TEST(FieldApiTest, ntt)
...
+
+  START_TIMER(element_wise_op)
+  for (int i = 0; i < 100000; ++i) {
+    out_element_wise[i] = in_eq[i] * (in_a[i] * in_b[i] - in_c[i]);
+  }
+  // std::cout << "Strait forward function (Element wise) time: ";
+  END_TIMER(element_wise_op, oss.str().c_str(), true);
+
+  //----- written program ----------------------
+  Program<scalar_t> program_written(&combine_func, 4);
```
yshekel, line 986: compilation occurs here?

icicle/tests/test_field_api.cpp
```diff
@@ -939,6 +942,128 @@ TYPED_TEST(FieldApiTest, ntt)
...
+  // init program
+  prog_exe_predef.m_variable_ptrs[0] = in_a.get();
+  prog_exe_predef.m_variable_ptrs[1] = in_b.get();
+  prog_exe_predef.m_variable_ptrs[2] = in_c.get();
+  prog_exe_predef.m_variable_ptrs[3] = in_eq.get();
+  prog_exe_predef.m_variable_ptrs[4] = out_predef_program.get();
+
+  oss.str("");
+  oss.clear();
+  oss << "Program predefined time: ";
```
yshekel, line 1033: minor: note you can simply put this string directly in END_TIMER in this case

yshekel (commented):
left a few small comments. Mainly maybe we should consider moving this test to the backend and this should allow us to do it for any backend if it's not a frontend api.

icicle/tests/test_field_api.cpp
mickeyasa: we talked about that yesterday
We said we'll start with puting it in field and see what is the volume.
Add another suite for backend is fine

icicle/tests/test_field_api.cpp
mickeyasa: true.
Took it before Koren's push

icicle/tests/test_field_api.cpp
mickeyasa: the program can be written using a function or predefined 
specifically in this case, I coded the predefined function for comaprison

icicle/tests/test_field_api.cpp
mickeyasa: if by compilation you mean building the instructions, then yes.

icicle/tests/test_field_api.cpp
mickeyasa: currently only 1 output.
Next step is to support multiple outputs.
Anyway the outputs arrive after the inputs

icicle/tests/test_field_api.cpp
mickeyasa: its already there. 
I don't follow.

icicle/tests/test_field_api.cpp
yshekel: I meant sumcheck in field, not directly testing backends here

icicle/tests/test_field_api.cpp
yshekel: so written mean the user is writing it?

icicle/tests/test_field_api.cpp
yshekel: I mean you don't need oss for a simple string but never mind

icicle/tests/test_field_api.cpp
mickeyasa: yeap

idanfr-ingo (commented):
add few comments.
I think we should add tests for programs with constants and programs with multiple outputs (if the support of such program was added)

icicle/tests/test_field_api.cpp
```diff
@@ -939,6 +942,128 @@ TYPED_TEST(FieldApiTest, ntt)
...
+  scalar_t::rand_host_many(in_a.get(), total_size);
+  auto in_b = std::make_unique<scalar_t[]>(total_size);
+  scalar_t::rand_host_many(in_b.get(), total_size);
+  auto in_c = std::make_unique<scalar_t[]>(total_size);
+  scalar_t::rand_host_many(in_c.get(), total_size);
+  auto in_eq = std::make_unique<scalar_t[]>(total_size);
+  scalar_t::rand_host_many(in_eq.get(), total_size);
+
+  //----- element wise operation ----------------------
+  auto out_element_wise = std::make_unique<scalar_t[]>(total_size);
```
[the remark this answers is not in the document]
mickeyasa: explicit it is

icicle/tests/test_field_api.cpp
```diff
@@ -939,6 +942,128 @@ TYPED_TEST(FieldApiTest, ntt)
...
+MlePoly combine_func(std::vector<MlePoly>& inputs)
+{
+  const MlePoly& A = inputs[0];
+  const MlePoly& B = inputs[1];
+  const MlePoly& C = inputs[2];
+  const MlePoly& EQ = inputs[3];
+  return (EQ * (A * B - C));
+}
+
+TEST_F(FieldApiTestBase, CpuProgramExecutor)
```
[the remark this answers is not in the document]
mickeyasa: multiple outputs not yet supported - will be soon.
I'll add the test for constants together with that

icicle/tests/test_field_api.cpp
mickeyasa: to be resolved once Aviad will make order in the CI and the test suits

yshekel (approved):

idanfr-ingo (approved):

PR #690, opened 2024-12-10
[FIX]: Add nodejs setup for release script

jeremyfelder:
## Describe the changes

This PR:
- Fixes the release script by ensuring nodejs is installed before trying to bump the docs version
- Moves building docker images for releases into a separate workflow as they only need to be rebuilt if prerequisite packages change
- Updates the release workflow and the build_all script to pull and use the latest docker images

### DIFF
--- a/.github/workflows/build-release-docker-img.yml
+++ b/.github/workflows/build-release-docker-img.yml
@@ -1,0 +1,34 @@
+name: Build Release Docker Images
+
+on: workflow_dispatch
+    
+jobs:
+  build-and-publish:
+    name: Build and Publish Docker Images
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        images:
+          - ubuntu22
+          - ubuntu20
+          - ubi8
+          - ubi9
+    steps:
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Build, tag, and push docker image
+        uses: docker/build-push-action@v6
+        with:
+          # this label connects the icicle repo with the built package
+          # See more: https://docs.github.com/en/packages/learn-github-packages/connecting-a-repository-to-a-package
+          labels: |
+            org.opencontainers.image.source=https://github.com/ingonyama-zk/icicle
+          file: ./Dockerfile.${{ matrix.images }}
+          tags: ghcr.io/ingonyama-zk/icicle-release-${{ matrix.images }}-cuda122:latest
+          push: true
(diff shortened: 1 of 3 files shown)


scripts/release/build_all.sh
```diff
@@ -71,9 +66,13 @@ docker run --rm --gpus all                  \
             -v ./scripts:/scripts           \
             icicle-release-ubi8-cuda122 bash /scripts/release/build_release_and_tar.sh icicle_$version ubi8 cuda122 &
 
+# NOTE: The last command should not be run in the background,
```
yshekel, line 69: you are correct here but what happens if this one is finished but the previous ones still did not? maybe need some way to wait for all.

scripts/release/build_all.sh
```diff
@@ -29,16 +29,11 @@ if [[ ! -d "./icicle" || ! -d "./scripts" ]]; then
...
-echo "Building Docker images..."
-# Ubuntu 22.04, CUDA 12.2.2
-docker build -t icicle-release-ubuntu22-cuda122 -f ./scripts/release/Dockerfile.ubuntu22 .
-# Ubuntu 20.04, CUDA 12.2.2
-docker build -t icicle-release-ubuntu20-cuda122 -f ./scripts/release/Dockerfile.ubuntu20 .
-# ubi8 (rhel compatible), CUDA 12.2.2
-docker build -t icicle-release-ubi8-cuda122 -f ./scripts/release/Dockerfile.ubi8 .
-# ubi9 (rhel compatible), CUDA 12.2.2
-docker build -t icicle-release-ubi9-cuda122 -f ./scripts/release/Dockerfile.ubi9 .
+# Download latest build images
```
yshekel, line 32: I assume you built and pushed them manually? can you commit a script for doing so as well? so we can easily build and push when updating the dockers

jeremyfelder: @yshekel There is a github workflow for it https://github.com/ingonyama-zk/icicle/blob/13220b56671f9d48caeac1c60ccff5e9dc9a0497/.github/workflows/build-release-docker-img.yml.

I can make it a script if you want so its possible to run outside of github actions

scripts/release/build_all.sh
jeremyfelder: Fixed in commit [d8dfba9](https://github.com/ingonyama-zk/icicle/pull/690/commits/d8dfba99740a15eda87ba14a81fa61483018f2ea)

.github/workflows/build-release-docker-img.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Build, tag, and push docker image
+        uses: docker/build-push-action@v6
+        with:
+          # this label connects the icicle repo with the built package
+          # See more: https://docs.github.com/en/packages/learn-github-packages/connecting-a-repository-to-a-package
+          labels: |
+            org.opencontainers.image.source=https://github.com/ingonyama-zk/icicle
+          file: ./Dockerfile.${{ matrix.images }}
```
LeonHibnik, line 32: Update the path to the Dockerfiles

yshekel (approved):

LeonHibnik (approved):

PR #692, opened 2024-12-11
add sumcheck frontend

mickeyasa:
add the classes for sumcheck frontend


icicle/include/icicle/sumcheck/sumcheck_proof.h
```diff
@@ -0,0 +1,51 @@
...
+   * @tparam S Type of the field element (e.g., prime field or extension field elements).
+   */
+
+  template <typename S>
+  class SumCheckProof
+  {
+  public:
+    // Constructor
+    SumCheckProof(uint nof_round_polynomials, uint round_polynomial_degree)
+        : m_round_polynomilas(nof_round_polynomials, std::vector<S>(round_polynomial_degree + 1))
```
krakhit, line 25: m_round_polynomials

icicle/include/icicle/sumcheck/sumcheck_proof.h
```diff
@@ -0,0 +1,51 @@
...
+   *
+   * @tparam S Type of the field element (e.g., prime field or extension field elements).
+   */
+
+  template <typename S>
+  class SumCheckProof
+  {
+  public:
+    // Constructor
+    SumCheckProof(uint nof_round_polynomials, uint round_polynomial_degree)
```
krakhit, line 24: Just a confusion here: in the general proof struct:
SumcheckProof<F: Field> {
   num_vars: usize,
   degree: usize,
   round_polynomials: Vec<Vec<F>>,
}

The num_vars : X_1,X_2....,X_n (size of boolean hypercube) where is it accessed from? is it inferred from input poly or user provides it?

mickeyasa: the polynomials are populated later using the set_round_polynomial_value function

icicle/backend/cpu/CMakeLists.txt
```diff
@@ -18,6 +18,9 @@ if (FIELD)
     if (POSEIDON2)
         target_sources(icicle_field PRIVATE src/hash/cpu_poseidon2.cpp)
     endif()
+    if(SUMCHECK OR HASH)
```
yshekel, line 21: this is strange but we discussed it

icicle/backend/cpu/include/cpu_sumcheck.h
```diff
@@ -0,0 +1,32 @@
...
+  class CpuSumcheckBackend : public SumcheckBackend<F>
+  {
+  public:
+    CpuSumcheckBackend(const F& claimed_sum, SumcheckTranscriptConfig<F>&& transcript_config)
+        : SumcheckBackend<F>(claimed_sum, std::move(transcript_config))
+    {
+    }
+
+    eIcicleError get_proof(
+      const std::vector<std::vector<F>*>& input_polynomials,
```
yshekel, line 21: (1) maybe rename to mle?
(2) pointer to vector is strange. Why do you need that?

icicle/backend/cpu/include/cpu_sumcheck.h
```diff
@@ -0,0 +1,32 @@
...
+      const CombineFunction<F>& combine_function,
+      SumCheckConfig& config,
```
yshekel, line 23: const?

icicle/backend/cpu/include/cpu_sumcheck.h
```diff
@@ -0,0 +1,32 @@
...
+      SumCheckProof<F>& sumcheck_proof /*out*/) const override
+    {
+      return eIcicleError::API_NOT_IMPLEMENTED;
+    }
+
```
yshekel, line 28: since this is not implemented, isn't it better to avoid registration of the backend? it could confuse users, although still not released.

icicle/cmake/fields_and_curves.cmake
```diff
@@ -2,17 +2,17 @@
 # Define available fields with an index and their supported features
 # Format: index:field:features
 set(ICICLE_FIELDS
-  1001:babybear:NTT,EXT_FIELD,POSEIDON,POSEIDON2
-  1002:stark252:NTT,POSEIDON,POSEIDON2
-  1003:m31:EXT_FIELD,POSEIDON,POSEIDON2
+  1001:babybear:NTT,EXT_FIELD,POSEIDON,POSEIDON2,SUMCHECK
```
yshekel, line 5: should remove from small fields

icicle/include/icicle/backend/sumcheck_backend.h
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * @brief Calculate the sumcheck based on the inputs and retrieve the Sumcheck proof.
+     * @param input_polynomials a vector of MLE polynomials to process
+     * @param combine_function a program that define how to fold all MLS polynomials into the round polynomial.
+     * @param config Configuration for the Sumcheck operation.
+     * @param sumcheck_proof Reference to the SumCheckProof object where all round polynomials will be stored.
+     * @return Error code of type eIcicleError.
+     */
+    virtual eIcicleError get_proof(
+      const std::vector<std::vector<F>*>& input_polynomials,
```
yshekel, line 51: same comment about pointer to vector. Also note this can be really annoying for FFI

icicle/include/icicle/sumcheck/sumcheck.h
```diff
@@ -1,13 +1,32 @@
 #pragma once
 
+#include "icicle/errors.h"
+#include "icicle/program/returning_value_program.h"
+#include "icicle/sumcheck/sumcheck_proof.h"
+#include "icicle/sumcheck/sumcheck_config.h"
 #include "icicle/sumcheck/sumcheck_transcript_config.h"
-
+#include "icicle/backend/sumcheck_backend.h"
```
yshekel, line 8: This is not supposed to be included here. Only backends are supposed to include this to register

icicle/include/icicle/sumcheck/sumcheck.h
```diff
@@ -1,13 +1,32 @@
...
+  /**
+   * @brief Static factory method to create a Sumcheck instance.
+   *
+   * @param claimed_sum The total sum of the values of a multivariate polynomial f(x₁, x₂, ..., xₖ)
+   * when evaluated over all possible Boolean input combinations
+   * @param transcript_config Configuration for encoding and hashing prover messages.
+   * @return A Sumcheck object initialized with the specified backend.
+   */
+  template <typename F>
+  Sumcheck<F> create_sumcheck(const F& claimed_sum, SumcheckTranscriptConfig<F>&& transcript_config);
```
yshekel, line 22: you have both create and create_sumcheck in this class and they are the same.

icicle/include/icicle/sumcheck/sumcheck_proof.h
```diff
@@ -0,0 +1,51 @@
...
+        : m_round_polynomials(nof_round_polynomials, std::vector<S>(round_polynomial_degree + 1))
+    {
+      if (nof_round_polynomials == 0) {
+        ICICLE_LOG_ERROR << "Number of round polynomials(" << nof_round_polynomials << ") in the proof must be >0";
+      }
+    }
+
+    // set the value of polynomial round_polynomial_idx at x = evaluation_idx
+    void set_round_polynomial_value(int round_polynomial_idx, int evaluation_idx, S& value)
```
yshekel, line 33: const S& ?

icicle/include/icicle/sumcheck/sumcheck_proof.h
yshekel, line 24: what is uint? probably defined in some headers but better stick to standard uint32_t or similar

icicle/src/sumcheck.cpp
```diff
@@ -0,0 +1,20 @@
+// #include "icicle/errors.h"
```
yshekel, line 1: remove file?

icicle/tests/test_field_api.cpp
```diff
@@ -1054,6 +1056,31 @@ TEST_F(FieldApiTestBase, CpuProgramExecutorReturningVal)
...
 }
 
+TEST_F(FieldApiTestBase, Sumcheck)
+{
+  int mle_poly_size = 1 << 13;
+  int nof_mle_poly = 4;
+  scalar_t claimed_sum = scalar_t::from(8);
+
+  // create transcript_config
+  SumcheckTranscriptConfig<scalar_t> transcript_config; // TODO Miki: define labels?
```
yshekel, line 1066: define labels and hash yes

icicle/tests/test_field_api.cpp
```diff
@@ -1054,6 +1056,31 @@ TEST_F(FieldApiTestBase, CpuProgramExecutorReturningVal)
...
+
+  // create sumcheck
+  auto sumcheck = Sumcheck<scalar_t>::create(claimed_sum, std::move(transcript_config));
+
+  // generate inputs
+  std::vector<std::vector<scalar_t>*> mle_polynomials(nof_mle_poly);
+  for (auto& mle_poly_ptr : mle_polynomials) {
+    mle_poly_ptr = new std::vector<scalar_t>(mle_poly_size);
```
yshekel, line 1074: avoid new at all cost. This is a bad thing and indeed your test is leaking memory since you don't call delete.
ALways prefer smart pointer and even better if you can make the vector pointer jsut a vector, and move it into the outer vector. It's zero cost almost

icicle/backend/cpu/include/cpu_sumcheck.h
krakhit: recommend: Input MLE's

icicle/include/icicle/sumcheck/sumcheck_proof.h
```diff
@@ -0,0 +1,72 @@
...
+
+namespace icicle {
+
+  /**
+   * @brief Represents a sumcheck proof.
+   *
+   * This class encapsulates the sumcheck proof, contains the evaluations of the round polynomials
+   * for each layer at the sumcheck proof.
+   * Evaluations are at x = 0, 1, 2 ... K
+   * Where K is the degree of the combiine function used at the sumcheck protocol.
```
idanfr-ingo, line 14: typo - "combine"

icicle/backend/cpu/CMakeLists.txt
LeonHibnik: For only HASH defined, why do anyone forced to build sumcheck?

idanfr-ingo (approved):

icicle/backend/cpu/include/cpu_sumcheck.h
```diff
@@ -0,0 +1,159 @@
...
+  public:
+    CpuSumcheckBackend(const F& claimed_sum, SumcheckTranscriptConfig<F>&& transcript_config)
+        : SumcheckBackend<F>(claimed_sum, std::move(transcript_config)),
+          m_cpu_sumcheck_transcript(claimed_sum, std::move(transcript_config))
+    {
+    }
+
+    // Calculate a proof for the mle polynomials
+    eIcicleError get_proof(
+      const std::vector<F*>& mle_polynomials,
```
yshekel, line 24: should this be const F*? I assume you don't mean to write the mles? right?

icicle/backend/cpu/include/cpu_sumcheck.h
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    // calculate alpha for the next round based on the round_polynomial of the current round
+    F get_alpha(std::vector<F>& round_polynomial) override
+    {
+      return m_cpu_sumcheck_transcript.get_alpha(round_polynomial);
+    }
+
+    // Reset the sumcheck transcript before a new proof generation or verification
+    void reset_transcript(const uint32_t mle_polynomial_size, const uint32_t combine_function_poly_degree) override
```
yshekel, line 88: why isn't this private? do you expect users to call it? Can't you call it when calling get_proof()? and not expose this to users

icicle/backend/cpu/include/cpu_sumcheck_transcript.h
```diff
@@ -0,0 +1,129 @@
...
+    const int nof_bytes_to_copy = std::min(sizeof(alpha), hash_result.size());
+    std::memcpy(&alpha, hash_result.data(), nof_bytes_to_copy);
+    alpha = alpha * S::one();
+  }
+
+  // append an integer uint32_t to hash input
+  void append_u32(std::vector<std::byte>& byte_vec, const uint32_t data)
+  {
+    const std::byte* data_bytes = reinterpret_cast<const std::byte*>(&data);
+    byte_vec.insert(byte_vec.end(), data_bytes, data_bytes + sizeof(uint32_t));
```
yshekel, line 73: Isn't is resizing the vector every time of every few times? Is there a way to pre allocate it and avoid it?

icicle/cmake/fields_and_curves.cmake
```diff
@@ -2,18 +2,18 @@
 # Define available fields with an index and their supported features
 # Format: index:field:features
 set(ICICLE_FIELDS
-  1001:babybear:NTT,EXT_FIELD,POSEIDON,POSEIDON2
-  1002:stark252:NTT,POSEIDON,POSEIDON2
-  1003:m31:EXT_FIELD,POSEIDON,POSEIDON2
-  1004:koalabear:NTT,EXT_FIELD,POSEIDON,POSEIDON2
+  1001:babybear:NTT,EXT_FIELD,POSEIDON,POSEIDON2,SUMCHECK
+  1002:stark252:NTT,POSEIDON,POSEIDON2,SUMCHECK
+  1003:m31:EXT_FIELD,POSEIDON,POSEIDON2,SUMCHECK
```
yshekel, line 7: I think small fields (m31 and babybear, koalabear) are not supposed to be using this sumcheck. @krakhit ?

icicle/include/icicle/backend/sumcheck_backend.h
```diff
@@ -0,0 +1,109 @@
...
+     * @brief Calculate the sumcheck based on the inputs and retrieve the Sumcheck proof.
+     * @param mle_polynomials a vector of MLE polynomials to process
+     * @param mle_polynomial_size the size of each MLE polynomial
+     * @param combine_function a program that define how to fold all MLS polynomials into the round polynomial.
+     * @param config Configuration for the Sumcheck operation.
+     * @param sumcheck_proof Reference to the SumcheckProof object where all round polynomials will be stored.
+     * @return Error code of type eIcicleError.
+     */
+    virtual eIcicleError get_proof(
+      const std::vector<F*>& mle_polynomials,
```
yshekel, line 52: I commented on this already but I think it should be const F*, since you should not modify the input mles

icicle/include/icicle/backend/sumcheck_backend.h
```diff
@@ -0,0 +1,109 @@
...
+      const CombineFunction<F>& combine_function,
+      const SumcheckConfig& config,
+      SumcheckProof<F>& sumcheck_proof /*out*/) = 0;
+
+    /**
+     * @brief Initialize the transcript for the upcoming calculation of the FIat shamir.
+     * @param mle_polynomial_size the number of elements in an MLE polynomial
+     * @param combine_function_poly_degree the degree of the combine function
+     */
+    virtual void reset_transcript(const uint32_t mle_polynomial_size, const uint32_t combine_function_poly_degree) = 0;
```
yshekel, line 63: I think this should be implicit when calling get_proof() and not exposed to users. It's strange and bug prone.

icicle/include/icicle/backend/sumcheck_backend.h
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * @brief Calculate the sumcheck based on the inputs and retrieve the Sumcheck proof.
+     * @param mle_polynomials a vector of MLE polynomials to process
+     * @param mle_polynomial_size the size of each MLE polynomial
+     * @param combine_function a program that define how to fold all MLS polynomials into the round polynomial.
+     * @param config Configuration for the Sumcheck operation.
+     * @param sumcheck_proof Reference to the SumcheckProof object where all round polynomials will be stored.
+     * @return Error code of type eIcicleError.
+     */
+    virtual eIcicleError get_proof(
```
yshekel, line 51: minor but: I know it's consistent with merkle tree so probably should remain as is but I think renaming to prove() is pretier.

icicle/include/icicle/program/returning_value_program.h
```diff
@@ -24,9 +24,27 @@ namespace icicle {
       this->set_as_inputs(program_parameters);
       program_parameters[nof_inputs] = program_func(program_parameters); // place the output after the all inputs
       this->generate_program(program_parameters);
+      m_poly_degree = program_parameters[nof_inputs].m_operation->m_poly_degree;
     }
 
     // Generate a program based on a PreDefinedPrograms
-    ReturningValueProgram(PreDefinedPrograms pre_def) : Program<S>(pre_def) {}
+    ReturningValueProgram(PreDefinedPrograms pre_def) : Program<S>(pre_def)
```
yshekel, line 31: This is really a strange name 🤣

icicle/include/icicle/backend/sumcheck_backend.h
yshekel: I now noticed it's in the backend code and not exposed to the users so we can live with that but still can we avoid this and have get_proof() do it instead?

icicle/include/icicle/sumcheck/sumcheck.h
```diff
@@ -16,18 +35,104 @@ namespace icicle {
...
+      const CombineFunction<F>& combine_function,
+      const SumcheckConfig& config,
+      SumcheckProof<F>& sumcheck_proof /*out*/) const
     {
+      return m_backend->get_proof(mle_polynomials, mle_polynomial_size, combine_function, config, sumcheck_proof);
     }
 
-    // Add public methods for protocol operations, e.g., prove, verify.
+    /**
+     * @brief Verify an element against the Sumcheck round polynomial.
```
yshekel, line 65: It's not verifying the last step right? Maybe we should mention this in the comment.

icicle/include/icicle/sumcheck/sumcheck_proof.h
```diff
@@ -0,0 +1,72 @@
...
+    {
+      for (auto& round_poly : m_round_polynomials) {
+        for (auto& element : round_poly) {
+          element = S::zero();
+        }
+      }
+    }
+
+  private:
+    std::vector<std::vector<S>> m_round_polynomials; // logN vectors of round_poly_degree elements
```
yshekel, line 55: those are always on host right? even if using another backend like CUDA

icicle/include/icicle/sumcheck/sumcheck_proof.h
```diff
@@ -0,0 +1,72 @@
...
+
+  public:
+    // for debug
+    void print_proof()
+    {
+      std::cout << "Sumcheck Proof :" << std::endl;
+      for (int round_poly_i = 0; round_poly_i < m_round_polynomials.size(); round_poly_i++) {
+        std::cout << "  Round polynomial " << round_poly_i << ":" << std::endl;
+        ;
```
yshekel, line 64: remove this ;

icicle/include/icicle/sumcheck/sumcheck_transcript_config.h
```diff
@@ -46,10 +46,12 @@ namespace icicle {
   template <typename F>
   class SumcheckTranscriptConfig
   {
-  public:
   public:
     // Default Constructor
-    SumcheckTranscriptConfig() : m_little_endian(true), m_seed_rng(0) {}
+    SumcheckTranscriptConfig()
```
yshekel, line 51: Is it a good idea to have this default? Why not require users to specify the seed and hash function?

icicle/tests/test_field_api.cpp
```diff
@@ -1188,6 +1190,55 @@ TEST_F(FieldApiTestBase, ProgramExecutorVecOpDataOnDevice)
...
+
+  // create transcript_config
+  SumcheckTranscriptConfig<scalar_t> transcript_config; // default configuration
+
+  // generate inputs
+  std::vector<scalar_t*> mle_polynomials(nof_mle_poly);
+  for (int poly_i = 0; poly_i < nof_mle_poly; poly_i++) {
+    mle_polynomials[poly_i] = new scalar_t[mle_poly_size];
+    for (int element_i = 0; element_i < mle_poly_size; element_i++) {
+      mle_polynomials[poly_i][element_i] = scalar_t::from(poly_i * 10 + element_i + 1);
```
yshekel, line 1207: maybe randomize?

icicle/tests/test_field_api.cpp
```diff
@@ -1188,6 +1190,55 @@ TEST_F(FieldApiTestBase, ProgramExecutorVecOpDataOnDevice)
...
+    const scalar_t c = mle_polynomials[2][element_i];
+    const scalar_t eq = mle_polynomials[3][element_i];
+    claimed_sum = claimed_sum + (a * b - c) * eq;
+  }
+
+  // ===== Prover side ======
+  // create sumcheck
+  auto prover_sumcheck = create_sumcheck<scalar_t>(claimed_sum, std::move(transcript_config));
+
+  CombineFunction<scalar_t> combine_func(EQ_X_AB_MINUS_C);
```
yshekel, line 1225: A little confused, is there a difference between CombineFunction and program?

icicle/tests/test_field_api.cpp
```diff
@@ -1188,6 +1190,55 @@ TEST_F(FieldApiTestBase, ProgramExecutorVecOpDataOnDevice)
...
+  SumcheckConfig config;
+  SumcheckProof<scalar_t> sumcheck_proof(log_mle_poly_size, nof_mle_poly - 1);
```
yshekel, line 1227: Can you avoid it by having an empty constructor and let the prove function resize?

icicle/tests/test_field_api.cpp
```diff
@@ -1188,6 +1190,55 @@ TEST_F(FieldApiTestBase, ProgramExecutorVecOpDataOnDevice)
...
+
+  prover_sumcheck.get_proof(mle_polynomials, mle_poly_size, combine_func, config, sumcheck_proof);
```
yshekel, line 1229: missing ICICLE_CHECK() around this to make sure it suceeds

icicle/tests/test_field_api.cpp
```diff
@@ -1188,6 +1190,55 @@ TEST_F(FieldApiTestBase, ProgramExecutorVecOpDataOnDevice)
...
+  for (auto& mle_poly_ptr : mle_polynomials) {
+    delete[] mle_poly_ptr;
+  }
+
+  // ===== Verifier side ======
+  // create sumcheck
+  auto verifier_sumcheck = create_sumcheck<scalar_t>(claimed_sum, std::move(transcript_config));
+  bool verification_pass;
+  verifier_sumcheck.verify(sumcheck_proof, verification_pass);
```
yshekel, line 1238: here as well. Also, better initialize verification_pass=false to make sure it is set to true inside.

yshekel (changes requested):
Overall looks good but I left some comments

icicle/cmake/fields_and_curves.cmake
krakhit: For small fields: it is not recommended to use it, it would still work though for testing etc. 

1. We need Fiat Shamir hash to output randomness in extension field, which we dont have at this point.  The algorithm would still work, but would not have the required soundness. 

2. The current algorithm that we use in the accumulation phase, when used with extension fields, would be only doing extension-extension multiplications/additions after round 1, and we lose the "small fields" compute advantage. 

3. The precompute algorithm for the accumulation phase, https://eprint.iacr.org/2024/1046  uses precomputed tables, for minimising, extension-extension, and base-extension multiplications. It is ideal for small fields, since we can afford more memory, and also take advantage of efficient base field arithmetic.

icicle/include/icicle/sumcheck/sumcheck.h
```diff
@@ -16,18 +33,117 @@ namespace icicle {
...
+     * First round polynomial is verified against the claimed sum.
+     * Last round polynomial is not verified.
+     * @param sumcheck_proof The SumcheckProof object includes the round polynomials.
+     * @param valid output valid bit. True if the Proof is valid, false otherwise.
+     * @return Error code of type eIcicleError indicating success or failure.
+     */
+    eIcicleError verify(
+      SumcheckProof<F>& sumcheck_proof,
```
yshekel, line 78: why not const proof?

yshekel: @mickeyasa

icicle/include/icicle/sumcheck/sumcheck.h
yshekel: @mickeyasa

icicle/backend/cpu/include/cpu_sumcheck.h
yshekel: @mickeyasa

icicle/include/icicle/backend/sumcheck_backend.h
yshekel: @mickeyasa

icicle/include/icicle/sumcheck/sumcheck_transcript_config.h
yshekel: @mickeyasa

yshekel (approved):

PR #699, opened 2024-12-23
Add capability to set a default device for all threads

jeremyfelder:
## Describe the changes

This PR adds the capability to set a default device for all threads. Setting a default device does not override a thread's explicitly set device

### DIFF
--- a/icicle/src/device_api.cpp
+++ b/icicle/src/device_api.cpp
@@ -58,6 +58,15 @@ namespace icicle {
 
     const Device& get_default_device() { return m_default_device; }
 
+    eIcicleError set_default_device(const Device& dev)
+    {
+      if (!is_device_registered(dev.type)) {
+        THROW_ICICLE_ERR(eIcicleError::INVALID_DEVICE, "Device type " + std::string(dev.type) + " has not been registered");
+      }
+      m_default_device = dev;
+      return eIcicleError::SUCCESS;
+    }
+
     std::vector<std::string> get_registered_devices_list()
     {
       std::vector<std::string> registered_devices;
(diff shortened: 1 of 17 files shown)


icicle/src/device_api.cpp
```diff
@@ -58,6 +58,16 @@ namespace icicle {
 
     const Device& get_default_device() { return m_default_device; }
 
+    eIcicleError set_default_device(const Device& dev)
+    {
+      if (!is_device_registered(dev.type)) {
+        THROW_ICICLE_ERR(
```
yshekel, line 64: Why not return an error and do nothing? instead of crashing

icicle/src/device_api.cpp
```diff
@@ -58,6 +58,16 @@ namespace icicle {
...
+          eIcicleError::INVALID_DEVICE, "Device type " + std::string(dev.type) + " has not been registered");
+      }
+      m_default_device = dev;
```
yshekel, line 67: note that this is not thread safe. Not sure if it's a problem though, since users are not supposed to call this from multiple threads.

yshekel (approved):
Nice !!

icicle/src/device_api.cpp
LeonHibnik: I assume that the user doesn't have a set_device on each thread so returning an error here will result on making the program run on cpu instead of what the user wanted

LeonHibnik (approved):

icicle/src/device_api.cpp
jeremyfelder: I will add docs for this and will mention it there

icicle/src/device_api.cpp
yshekel: Then he should not ignore errors. But ok

jeremyfelder: I prefer not throwing here and allow the user to decide to terminate or fallback to CPU as the original default.

There are other areas in our code I think we should make this change but I was following what we did elsewhere

PR #707, opened 2024-12-25
register program execution backends

idanfr-ingo:
- registration of program execution backend
- registration guidelines readme

cuda-backend-branch: idan/program-support


icicle/include/icicle/backend/README.md
```diff
@@ -0,0 +1,76 @@
+# How to Register a New Function
+
+For users to be able to use a backend function (whenever it's a CPU/GPU one), the function should be exposed to the user and be register to the matching backend. We do not let the user call the backend function directly, instead a wrapping function, with identical signature to all backend, is created as a user-facing function. Then, the backend specific functions are registered to their respective beckends in such a way that a dispatcher will know to call the appropriate function (acoorfing to the device used) whenever the user facing function is used.
```
mickeyasa, line 3: be register -> be registered

mickeyasa: Fixed "beckends" typo to "backends"

icicle/include/icicle/vec_ops.h
```diff
@@ -367,11 +367,16 @@ namespace icicle {
    * @param program a class that describes the functionality to run on each set of entries at the vectors.
    * @param size Size of each arrays. The program will be executed size times
    * @param config Configuration for the operation.
+   * @note Using the CUDA backend you can use config.ext to tell the function where each vector resides (device or
```
mickeyasa, line 370: Using the CUDA backend... -> When using the CUDA backend...
be including-->include
"you should have a key "is_<id>_on_device" for the #id vector in data_vec "  - not clear

mickeyasa (approved):
Some typos.
Do we have a clear document about the program and how to use it?

icicle/include/icicle/vec_ops.h
idanfr-ingo: Change the unclear part, is it clear now?

icicle/include/icicle/backend/README.md
```diff
@@ -0,0 +1,76 @@
+# How to Register a New Function
```
LeonHibnik, line 1: We should have this in the dev docs

LeonHibnik (approved):

icicle/include/icicle/backend/vec_ops_backend.h
```diff
@@ -53,6 +53,13 @@ namespace icicle {
   using scalarHighNonZeroIdxOpImpl = std::function<eIcicleError(
     const Device& device, const scalar_t* input, uint64_t size, const VecOpsConfig& config, int64_t* out_idx)>;
 
+  using programExecutionImpl = std::function<eIcicleError(
+    const Device& device,
+    std::vector<scalar_t*>& data,
```
yshekel, line 58: const

icicle/include/icicle/backend/README.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+### Define the registration function
+The registration function (`register_execute_program` at the example above) should be defined just before the macro:
+```cpp
+  void register_execute_program(const std::string& deviceType, programExecutionImpl);
+```
+It should have `void` as it return type and its name should be `register_<user-facing-function's-name>` (it can be inferred that the user facing function of the example above is `execute_program`). Wrong function name here will cause a linking error at compilation time.
+
+### Define function signature
```
yshekel, line 34: it's declare, not define

icicle/include/icicle/backend/README.md
```diff
@@ -0,0 +1,76 @@
+# How to Register a New Function
+
+For users to be able to use a backend function (whenever it's a CPU/GPU one), the function should be exposed to the user and be registered to the matching backend. We do not let the user call the backend function directly, instead a wrapping function, with identical signature to all backend, is created as a user-facing function. Then, the backend specific functions are registered to their respective backends in such a way that a dispatcher will know to call the appropriate function (according to the device used) whenever the user facing function is used.
+Here is a step-by-step explanation of how to do that registration process right.
+
+## The backend specific function
+The first step is to write the backend specific function. You should notice that its first argument must be `Device`. For example:
```
yshekel, line 7: I want to suggest a different order of things that makes more sense to me. You can ignore if you disagree:
1. first you define a template frontend API (this is used by the user directly and this is all he has to know) and config. **This is done in vec_ops.h for example.**
2. Then you declare the type that the backend needs to register and a register function. **This is in vec_ops_backend.h**
3. Now you need to instantiate the dispatcher (using ICICLE_DISPATCHER_INST) which also defines the register function (from 2). **This must be in vec_ops.cpp since this is compiled**
4. Finally a backend is implementing a function that matches the prototype from step 2 and uses the REGISTER macro to register it.

icicle/include/icicle/backend/vec_ops_backend.h
idanfr-ingo: data contains inputs and outputs therefore it cannot be const

yshekel (approved):

PR #723, opened 2025-01-02
Hadar/b const mult

HadarIngonyama:
This PR optimizes the multiplication by the constant 3b in the EC addition.
It seems to significantly improve the CPU MSM performance and the CUDA MSM performance a tiny bit.


icicle/include/icicle/curves/params/bls12_377.h
```diff
@@ -25,6 +25,11 @@ namespace bls12_377 {
     static constexpr point_field_t weierstrass_b = {0x00000001, 0x00000000, 0x00000000, 0x00000000,
                                                     0x00000000, 0x00000000, 0x00000000, 0x00000000,
                                                     0x00000000, 0x00000000, 0x00000000, 0x00000000};
+    static constexpr point_field_t weierstrass_3b = {0x00000003, 0x00000000, 0x00000000, 0x00000000,
```
mickeyasa, line 28: do we really need that extra definition?
using constexpr in the code does not yield the same performance?

icicle/include/icicle/fields/field.h
```diff
@@ -875,6 +875,22 @@ class Field
...
   friend HOST_DEVICE bool operator!=(const Field& xs, const Field& ys) { return !(xs == ys); }
 
+  template <typename Gen>
+  static HOST_DEVICE_INLINE Field mul_weierstrass_3b(const Field& xs)
+  {
+    Field r = {};
+    if constexpr (Gen::is_b_u32) {
+      r = mul_unsigned<Field{Gen::weierstrass_3b}.limbs_storage.limbs[0], Field>(xs);
+      if constexpr (Gen::is_b_neg)
+        return neg(r);
```
mickeyasa, line 885: I think it will be more efficient to write here another multiplication by neg(b) which can be calculated in constexpr

icicle/include/icicle/fields/complex_extension.h
```diff
@@ -158,6 +158,52 @@ class ComplexExtensionField
...
   }
 
+  template <typename Gen>
+  static HOST_DEVICE_INLINE FF mul_weierstrass_3b_real(const FF& xs)
+  {
+    FF r = {};
+    if constexpr (Gen::is_b_u32_g2_re) {
+      r = FF::template mul_unsigned<FF{Gen::weierstrass_3b_g2_re}.limbs_storage.limbs[0], FF>(xs);
+      if constexpr (Gen::is_b_neg_g2_re)
+        return FF::neg(r);
```
mickeyasa, line 168: I think it will be more efficient to write here another multiplication by neg(b) which can be calculated in constexpr

mickeyasa (commented):
some questions and optimizations

icicle/include/icicle/fields/field.h
HadarIngonyama: I don't understand.. xs is known only at runtime, this cannot be calculated in constexpr

icicle/include/icicle/fields/complex_extension.h
HadarIngonyama: same here

icicle/include/icicle/curves/params/bls12_377.h
HadarIngonyama: it probably will, i can try

icicle/include/icicle/fields/field.h
mickeyasa: according to your code,  the neg(r) is calculated in runtime
you can write instead of "return neg(r)" 
return mul_unsigned<neg(Field{Gen::weierstrass_3b}.limbs_storage.limbs[0], Field)>(xs);
that way the neg operation will be calculated at compile time

HadarIngonyama: that will ruin the whole point of the optimization... 3b is stored in absolute value because it's small that way

yshekel (approved):
Agree with miki that we should compute weistrass_3b at compile time instead. Other than that LGTM

mickeyasa:
.

mickeyasa (approved):
V

icicle/include/icicle/curves/params/grumpkin.h
```diff
@@ -16,7 +16,13 @@ namespace grumpkin {
     static constexpr point_field_t gen_y = {0x823f272c, 0x833fc48d, 0xf1181294, 0x2d270d45,
                                             0x6a45d63,  0xcf135e75, 0x00000002, 0x00000000};
 
-    static constexpr point_field_t weierstrass_b = {0xeffffff0, 0x43e1f593, 0x79b97091, 0x2833e848,
-                                                    0x8181585d, 0xb85045b6, 0xe131a029, 0x30644e72};
+    // actual value:
+    // static constexpr point_field_t weierstrass_b = {0xeffffff0, 0x43e1f593, 0x79b97091, 0x2833e848,
```
yshekel, line 20: remove

PR #731, opened 2025-01-12
fix bug in CPU vec ops regarding nof workers

yshekel:
- CPU vec-ops are using two threads less than available. 
- The intention was to reserve one for the main thread but this is redundant here since main thread is doing very little work.
- In addition, for a single core machine this is a bug, where 0 workers are used.


mickeyasa (approved):
great

LeonHibnik (approved):

PR #734, opened 2025-01-13
Hadar/base math

HadarIngonyama:
This PR separates cpu (host) math and gpu (device) math from the field class itself. Now the field calls the relevant base math according to the context.

## CUDA backend branch
# cuda-backend-branch: hadar/base-math

### DIFF
--- a/icicle/include/icicle/fields/field.h
+++ b/icicle/include/icicle/fields/field.h
@@ -116,15 +116,7 @@ public:
   static HOST_DEVICE_INLINE Field inv_log_size(uint32_t logn)
   {
     if (logn == 0) { return Field{CONFIG::one}; }
-#ifndef __CUDA_ARCH__ // A device function cannot call the icicle error function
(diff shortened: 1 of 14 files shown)


icicle/include/icicle/fields/field.h
```diff
@@ -37,6 +39,12 @@
 
 using namespace icicle;
 
+#ifdef __CUDA_ARCH__
+namespace base_math = device_math;
+#else
```
LeonHibnik, line 44: We should be able to extend this to Metal and Vulkan as well

HadarIngonyama: yes, probably

yshekel (approved):
I assume no performance impact right?

icicle/include/icicle/fields/field.h
LeonHibnik: @yshekel I meant the namespace

LeonHibnik (approved):

PR #735, opened 2025-01-13
Support android and vulkan

yshekel:
- updated cmake for android cross compilation, including vulkan
- updated log.h for android

### DIFF
--- a/icicle/cmake/setup.cmake
+++ b/icicle/cmake/setup.cmake
@@ -1,0 +1,71 @@
+
+# Option to cross-compile for Android
+option(BUILD_FOR_ANDROID "Cross-compile for Android" OFF)
+
+if (BUILD_FOR_ANDROID)
+    message(STATUS "Configuring for Android ARM64...")
+
+    set(ANDROID_MIN_API 29)
+    set(CMAKE_SYSTEM_NAME Android CACHE STRING "Target system name for cross-compilation")
+    set(ANDROID_ABI arm64-v8a CACHE STRING "Default Android ABI")
+    set(CMAKE_ANDROID_ARCH_ABI "${ANDROID_ABI}" CACHE STRING "Target ABI for Android") # deprecated??
+    set(ANDROID_PLATFORM "android-${ANDROID_MIN_API}" CACHE STRING "Android API level")
+    set(CMAKE_ANDROID_NDK /home/administrator/users/yuvals/icicle/icicle/backend/vulkan/android/cmdline-tools/ndk CACHE PATH "Path to Android NDK")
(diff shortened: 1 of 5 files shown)


icicle/cmake/setup.cmake
```diff
@@ -0,0 +1,79 @@
...
+        set(CMAKE_ANDROID_NDK ${ANDROID_NDK})
+    elseif (DEFINED ENV{ANDROID_NDK})
+        set(CMAKE_ANDROID_NDK $ENV{ANDROID_NDK})
+    endif()
+
+    # Debugging message for NDK path
+    message(STATUS "Using Android NDK: ${CMAKE_ANDROID_NDK}")
+
+    # Set toolchain and other options
+    set(ANDROID_MIN_API 29) # Minimum API (29 is android 10)
```
jeremyfelder, line 24: Whats the reason 29 is the min? I think a few years ago the recommended min was 23, not sure if thats still the case

jeremyfelder (approved):
🔥 🚀 
Left a couple clarifying questions but otherwise looks great

icicle/include/icicle/utils/log.h
```diff
@@ -3,6 +3,10 @@
 #include <iostream>
 #include <sstream>
```
mickeyasa, line 5: Looks like you can avoid all those ifdefs in this file and just add another seoparate file to android.

mickeyasa (approved):
one comment abou the log file

icicle/cmake/setup.cmake
```diff
@@ -0,0 +1,79 @@
...
+    set(CMAKE_SYSTEM_NAME Android CACHE STRING "Target system name for cross-compilation")
+    set(ANDROID_ABI arm64-v8a CACHE STRING "Default Android ABI")
```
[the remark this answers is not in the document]
yshekel: yes it should be able to also use arm v7 (32b) and x86. arm v7 won't compile the C{U backend though since it's using uint128.
Having said that I did not try

icicle/cmake/setup.cmake
yshekel: I will try 23

icicle/include/icicle/utils/log.h
yshekel: I will try to separate it

icicle/cmake/setup.cmake
yshekel: update: 23 did not work with the vulkan version but 24 did so updated to 24 (Android 7.0). It should cover >97% of devices according to chatgpt.

PR #737, opened 2025-01-14
msm-avoid-double-thread-launch-on-phase1

mickeyasa:
remove the thread launch that collapse all workers buckets into 1
move the collapse work to phase 2


yshekel:
Do you have numbers regarding performance with this PR?

mickeyasa:
Not based on the benchmark but looks like 10-15%  faster

yshekel (approved):

LeonHibnik (approved):

PR #739, opened 2025-01-15
Fix get_device_count() in Rust release mode

yshekel:
Fix bug in icicle_runtime::get_device_count() where returned value is 0 in release mode.


LeonHibnik (approved):

jeremyfelder (commented):
Lets add a small test for this?

yshekel:
> Lets add a small test for this?

It's a good point, except the bug is only in release mode and tests are executed in debug mode.
We can either (1) test in release mode or(2) add a test file that is tested in release mode. @jeremyfelder what do you think?

jeremyfelder:
> > Lets add a small test for this?
> 
> It's a good point, except the bug is only in release mode and tests are executed in debug mode. We can either (1) test in release mode or(2) add a test file that is tested in release mode. @jeremyfelder what do you think?

🤔 I don't see that we even test the icicle-runtime crate in the CI... but we do test in release mode already.

yshekel:
> > > Lets add a small test for this?
> > 
> > 
> > It's a good point, except the bug is only in release mode and tests are executed in debug mode. We can either (1) test in release mode or(2) add a test file that is tested in release mode. @jeremyfelder what do you think?
> 
> 🤔 I don't see that we even test the icicle-runtime crate in the CI... but we do test in release mode already.

Yes you are right. I wonder what happened in tests like this then https://github.com/ingonyama-zk/icicle/blob/d4a8153848d17425c3fc345882768c773a7b9407/wrappers/rust/icicle-core/src/ntt/tests.rs#L379
I was able to replicate this bug easily and see that I got count=0 in release mode so this is a little worrying

yshekel:
> > > > Lets add a small test for this?
> > > 
> > > 
> > > It's a good point, except the bug is only in release mode and tests are executed in debug mode. We can either (1) test in release mode or(2) add a test file that is tested in release mode. @jeremyfelder what do you think?
> > 
> > 
> > 🤔 I don't see that we even test the icicle-runtime crate in the CI... but we do test in release mode already.
> 
> Yes you are right. I wonder what happened in tests like this then
> 
> https://github.com/ingonyama-zk/icicle/blob/d4a8153848d17425c3fc345882768c773a7b9407/wrappers/rust/icicle-core/src/ntt/tests.rs#L379
> 
> 
> I was able to replicate this bug easily and see that I got count=0 in release mode so this is a little worrying

@jeremyfelder Actually we do have tests for icicle-runtime crate. I added a test for get_device_count()

jeremyfelder (approved):

PR #741, opened 2025-01-15
Vecops's Rust wrapper fix for batch>1

Koren-Brand:
## Describe the changes

This PR fixes the Vecops rust wrapper to work with batch > 1 (Which affected most Vector operations besides vector vector operations).
Added assertions when calling vecops functions in rust to prevent invalid configurations.
Minor change in CLI workflow to avoid failures when loading the CUDA backend.


wrappers/rust/icicle-core/src/vec_ops/tests.rs
```diff
@@ -165,16 +165,16 @@ pub fn check_vec_ops_scalars_sum<F: FieldImpl>(test_size: usize)
 where
     <F as FieldImpl>::Config: VecOps<F> + GenerateRandom<F>,
 {
+    let cfg = VecOpsConfig::default();
```
yshekel, line 168: let's define batch>1 then if it should work

wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -156,7 +156,113 @@ fn check_vec_ops_args<'a, F, T>(
...
+fn check_vec_ops_args_slice<F>(
+    input: &(impl HostOrDeviceSlice<F> + ?Sized),
+    offset: u64,
+    stride: u64,
+    size_in: u64,
+    size_out: u64,
+    output: &(impl HostOrDeviceSlice<F> + ?Sized),
+    cfg: &VecOpsConfig,
+) -> VecOpsConfig {
+    if input.len() as u64 != size_in * cfg.batch_size as u64 {
```
yshekel, line 231: you could remove size_in and check that batch divides input.size() like other apis

wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -156,7 +156,113 @@ fn check_vec_ops_args<'a, F, T>(
...
+    }
+    if output.len() as u64 != size_out * cfg.batch_size as u64 {
+        panic!(
+            "Output size, size_out and batch_size {}; {}; {} do not match (output.len() != size_out * cfg.batch_size)",
+            output.len(),
+            size_out,
+            cfg.batch_size
+        );
+    }
+    if offset + (size_out - 1) * stride >= size_in {
```
yshekel, line 247: isn't it 'offset+size*stride-1'?

wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -156,7 +156,113 @@ fn check_vec_ops_args<'a, F, T>(
...
+    }
+    setup_config(a, b, result, cfg)
+}
+
+fn check_vec_ops_args_reduction_ops<F>(
+    input: &(impl HostOrDeviceSlice<F> + ?Sized),
+    result: &(impl HostOrDeviceSlice<F> + ?Sized),
+    cfg: &VecOpsConfig,
+) -> VecOpsConfig {
+    if result.len() != cfg.batch_size as usize || input.len() % cfg.batch_size as usize != 0 {
```
yshekel, line 185: I think you cannot assing batch_size in the rust api and instead we use result.len() as the batch size.

wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -156,7 +156,113 @@ fn check_vec_ops_args<'a, F, T>(
...
+    setup_config(a, b, result, cfg)
+}
+
+fn check_vec_ops_args_scalar_ops<F, T>(
+    a: &(impl HostOrDeviceSlice<F> + ?Sized),
+    b: &(impl HostOrDeviceSlice<T> + ?Sized),
+    result: &(impl HostOrDeviceSlice<F> + ?Sized),
+    cfg: &VecOpsConfig,
+) -> VecOpsConfig {
+    if a.len() != cfg.batch_size as usize || b.len() != result.len() {
```
yshekel, line 168: why would a.len() be the batch?

wrappers/rust/icicle-core/src/vec_ops/mod.rs
Koren-Brand: I think this (current multiplication) is less taxing than a modulo operation.

wrappers/rust/icicle-core/src/vec_ops/mod.rs
Koren-Brand: You're right I'll fix it

wrappers/rust/icicle-core/src/vec_ops/mod.rs
Koren-Brand: This is the api for the cpp function

wrappers/rust/icicle-core/src/vec_ops/mod.rs
yshekel: It's minor and we better have all apis aligned

wrappers/rust/icicle-core/src/vec_ops/mod.rs
yshekel: assign*

wrappers/rust/icicle-core/src/vec_ops/mod.rs
Koren-Brand: But size_in is part of the Rust api (see the function that calls this function) and it should be checked so this is the logical way to check it together with cfg.batch_size
Unless you talk about removing size_in from the API as a whole which is another thing

wrappers/rust/icicle-core/src/vec_ops/mod.rs
Koren-Brand: Ok do we expect the user not to touch cfg? If so why is it part of the function inputs?

wrappers/rust/icicle-core/src/vec_ops/mod.rs
Koren-Brand: Having batch_size be implictly defined by division led to modulo be used instead of multiplication

wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -156,7 +156,119 @@ fn check_vec_ops_args<'a, F, T>(
...
+    cfg: &VecOpsConfig,
+) -> VecOpsConfig {
+    if b.len() != result.len() {
+        panic!(
+            "b.len() and result.len() do not match {} != {}",
+            b.len(),
+            result.len()
+        );
+    }
+    if b.len() % a.len() != 0 {
```
yshekel, line 175: shouldn't a.len() be exactly 1? This is the case of scalar+vector right?

Koren-Brand: Same as the previous - the vecops API specifies that the size scalar array (a) is batch size

yshekel: right

yshekel (approved):

jeremyfelder (approved):

PR #747, opened 2025-01-23
inverse cpu and rust wrappers

emirsoyturk:
cuda-backend-branch: emir/inverse


icicle/backend/cpu/src/field/cpu_vec_ops.cpp
```diff
@@ -24,6 +24,7 @@ enum VecOperation {
   VECTOR_SUB,
   VECTOR_MUL,
   VECTOR_DIV,
+  VECTOR_INV,
```
LeonHibnik, line 27: Can break code compatibility? should add as last and a TODO to re-order on next major

LeonHibnik (commented):
Looks good, left comment, also add golang wrapper

icicle/backend/cpu/src/field/cpu_vec_ops.cpp
yshekel: It's internal code of CPU backend. I don't think anyone is using it, especially since it is a cpp file, not a header.

wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -69,6 +69,12 @@ pub trait VecOps<F> {
         cfg: &VecOpsConfig,
     ) -> Result<(), eIcicleError>;
 
+    fn inv(
+        a: &(impl HostOrDeviceSlice<F> + ?Sized),
```
yshekel, line 73: suggest renaming a->input and result->output or something other than a,result

wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -371,6 +377,19 @@ where
     <<F as FieldImpl>::Config as VecOps<F>>::div(a, b, result, &cfg)
 }
 
+pub fn inv_scalars<F>(
+    a: &(impl HostOrDeviceSlice<F> + ?Sized),
```
yshekel, line 381: same here

yshekel (approved):

LeonHibnik (approved):

LeonHibnik (approved):

yshekel (approved):

PR #761, opened 2025-02-06
Idan/sumcheck test

idanfr-ingo:
Add test for the Cuda's sumcheck implementation

cuda-backend-branch: idan/sumcheck


icicle/tests/test_field_api.cpp
```diff
@@ -1280,28 +1276,49 @@ TEST_F(FieldApiTestBase, Sumcheck)
     claimed_sum = claimed_sum + (a * b - c) * eq;
   }
 
-  // ===== Prover side ======
-  // create sumcheck
-  auto prover_sumcheck = create_sumcheck<scalar_t>();
+  auto run = [&](
```
yshekel, line 1279: is this the only Sumcheck test? I think we must test everything before release!

mickeyasa (approved):

yshekel (approved):

PR #766, opened 2025-02-10
Feat/blake pow

nonam3e:
This PR adds POW algorithm

cuda-backend-branch: feat/blake-pow


icicle/include/icicle/hash/pow.h
```diff
@@ -0,0 +1,47 @@
...
+    icicleStreamHandle stream = nullptr; /**< Stream for asynchronous execution. Default is nullptr. */
+    bool is_challenge_on_device =
+      false; /**< True if challenge reside on the device (e.g., GPU), false if on the host (CPU). Default is false. */
+    bool is_result_on_device =
+      false; /**< True if challenge reside on the device (e.g., GPU), false if on the host (CPU). Default is false. */
+    bool is_async =
+      false; /**< True to run the pow solver asynchronously, false to run synchronously. Default is false. */
+    ConfigExtension* ext = nullptr; /**< Pointer to backend-specific configuration extensions. Default is nullptr. */
+  };
+  const uint8_t BLOCK_LEN = 64;
```
yshekel, line 26: why not in the config struct?

icicle/include/icicle/hash/pow.h
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * @brief Returns the default value of PowConfig for the POW function.
+   *
+   * @return Default value of PowConfig.
+   */
+  static PowConfig default_pow_config() { return PowConfig(); }
+
+  eIcicleError pow_solver(
+    Hash hasher,
+    uint8_t* challenge,
```
yshekel, line 39: for consistency maybe make it std::byte* like hash APIs

icicle/include/icicle/hash/pow.h
```diff
@@ -0,0 +1,47 @@
...
+    uint32_t challenge_size,
+    uint32_t padding_size,
+    uint8_t bits,
+    const PowConfig& config,
+    bool* found,
```
yshekel, line 44: Other ICICLE APIs in C++ use reference for outputs, not pointers so probably better to make it consistent.

icicle/include/icicle/hash/pow.h
nonam3e: oh this value is unused, forgot to delete it

icicle/include/icicle/hash/pow.h
yshekel: (1) I now noticed that you support them being device memory. Also the found boolean? This is a little confusing I think.
(2) Please add description to the API. It's unclear what should be the size of nonce for example etc.

wrappers/rust/icicle-hash/src/pow.rs
```diff
@@ -0,0 +1,85 @@
+use icicle_core::hash::{Hasher, HasherHandle};
+use icicle_runtime::{config::ConfigExtension, errors::eIcicleError, memory::HostOrDeviceSlice, IcicleStreamHandle};
+
+const PADDING_SIZE: u32 = 24;
```
yshekel, line 4: this should be in the config I think

wrappers/rust/icicle-hash/src/tests.rs
```diff
@@ -276,4 +277,40 @@ mod tests {
...
+        let mut gpu_found = false;
+        let mut gpu_nonce = 0;
+        let mut gpu_mined_hash = 0;
+        let found_host = HostSlice::from_mut_slice(std::slice::from_mut(&mut gpu_found));
+        let nonce_host = HostSlice::from_mut_slice(std::slice::from_mut(&mut gpu_nonce));
+        let mined_hash_host = HostSlice::from_mut_slice(std::slice::from_mut(&mut gpu_mined_hash));
+
+        let hasher = Blake3::new(0).unwrap();
+
+        let err = pow_solver(&hasher, input_host, BITS, &cfg, found_host, nonce_host, mined_hash_host);
```
yshekel, line 298: (1) I am not sure I understand this API. How do you define the threshold value for the hash result?
(2) can you verify that with the nonce you found, the hashing indeed succeeds? in this test I mean

yshekel (changes requested):
Overall looks good, I left a few comments.
Is it equivalent to other implementations? which?

LeonHibnik:
> Overall looks good, I left a few comments. Is it equivalent to other implementations? which?

@yshekel it should be equivalent to [this](https://github.com/arkworks-rs/nimue/blob/main/nimue-pow/src/blake3.rs#L72)

LeonHibnik (changes requested):
left a comment here and couple in cuda impl

nonam3e:
added `proof_of_work_check` to verify mined solutions, resolved comments as well

LeonHibnik (approved):

icicle/include/icicle/hash/pow.h
```diff
@@ -0,0 +1,87 @@
...
+   * @param solution_bits The number of leading zero bits required in the hash solution.
+   * @param config The configuration parameters for the PoW process.
+   * @param found A reference to a boolean flag that is set to true if a valid solution is found.
+   * @param nonce A reference to the nonce value that produces a valid hash (if found).
+   * @param mined_hash A reference to the resulting hash value of the successful PoW attempt.
+   *
+   * @return eIcicleError Error code indicating success or failure of the operation.
+   */
+  eIcicleError pow_solver(
+    Hash& hasher,
```
yshekel, line 52: why hasher and challenge not const?

icicle/include/icicle/hash/pow.h
```diff
@@ -0,0 +1,87 @@
...
+   * @param solution_bits The number of leading zero bits required in the hash solution.
+   * @param config The configuration parameters for the PoW process.
+   * @param nonce The nonce value to be verified.
+   * @param is_correct A reference to a boolean flag set to true if the nonce produces a valid hash.
+   * @param mined_hash A reference to store the computed hash value for verification.
+   *
+   * @return eIcicleError Error code indicating success or failure of the verification process.
+   */
+  eIcicleError pow_check(
+    Hash& hasher,
```
yshekel, line 79: here too

icicle/include/icicle/hash/pow.h
```diff
@@ -0,0 +1,87 @@
...
+   * @param challenge_size The size of the challenge data in bytes.
+   * @param solution_bits The number of leading zero bits required in the hash solution.
+   * @param config The configuration parameters for the PoW process.
+   * @param nonce The nonce value to be verified.
+   * @param is_correct A reference to a boolean flag set to true if the nonce produces a valid hash.
+   * @param mined_hash A reference to store the computed hash value for verification.
+   *
+   * @return eIcicleError Error code indicating success or failure of the verification process.
+   */
+  eIcicleError pow_check(
```
yshekel, line 78: mayber pow_verify? I think clearer than pow_check

yshekel (changes requested):
Left small comments and then I will approve

icicle/include/icicle/hash/pow.h
```diff
@@ -0,0 +1,87 @@
...
+   * @return eIcicleError Error code indicating success or failure of the operation.
+   */
+  eIcicleError pow_solver(
+    Hash& hasher,
+    std::byte* challenge,
+    uint32_t challenge_size,
+    uint8_t solution_bits,
+    const PowConfig& config,
+    bool& found,
+    uint64_t& nonce,
```
yshekel, line 58: is this always on host?

nonam3e: yeah, I changed it because they are just single values

yshekel (approved):

PR #771, opened 2025-02-12
Rust wrapper for porgram functionality

Koren-Brand:
Rust wrapper for program and symbol added
Fixed extension field missing from backend implementations of program executor

CUDA backend branch (Until it will be merged to main)
cuda-backend-branch: koren/register_program_execution_for_extension_field


icicle/src/vec_ops.cpp
```diff
@@ -531,18 +531,56 @@ namespace icicle {
...
+    data_vec.reserve(nof_params);
+    for (uint64_t i = 0; i < nof_params; i++) {
+      if (data_ptr[i] == nullptr) { throw std::invalid_argument("Null pointer found in parameters"); }
+      data_vec.push_back(data_ptr[i]);
+    }
+    return ExtExecuteProgramDispatcher::execute(data_vec, *program, size, config);
+  }
+
+  template <>
+  eIcicleError execute_program(
```
jeremyfelder, line 578: Should this function name be `extension_execute_program`?

wrappers/rust/icicle-core/src/program/tests.rs
jeremyfelder, line 1: Can we remove this file since the tests are in the vec_ops tests?

wrappers/rust/icicle-core/src/program/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+
+      fn initialize() {
+        test_utilities::test_load_and_init_devices();
+        test_utilities::test_set_main_device();
+      }
+
+      
+    }
+  }
+}
```
jeremyfelder, line 232: Can we delete this macro since the tests are in the vec_ops tests?

wrappers/rust/icicle-core/src/program/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+            }
+          }
+        }
+      }
+
+      // Returning Value Program trait implementation
+      impl ProgramBaseTrait<$field, Symbol> for ReturningValueProgram {
+        fn new_predefined(pre_def: PreDefinedProgram) -> Result<Self, eIcicleError> {
+          unsafe {
+            let prog_handle = ffi_create_predefined_program(pre_def);
```
jeremyfelder, line 154: This should be calling `create_predefined_returning_value_program`

wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -261,6 +275,31 @@ fn check_vec_ops_args_slice<F>(
     setup_config(input, input, output, cfg, batch_size)
 }
 
+fn check_execute_program<F, Data>( // COMMENT do we need this check in Rust (or any of the other checks)? seems to bloat rust for stuff not even implemented in cpp.
```
jeremyfelder, line 278: What do you mean by "for stuff not even implemented in cpp"?

icicle/include/icicle/program/returning_value_program.h
```diff
@@ -44,7 +44,21 @@ namespace icicle {
 
     int get_polynomial_degee() const { return m_poly_degree; }
 
+  protected:
+    ReturningValueProgram() {}
+
+    // Friend function for C-api to have access to the default constructor
+    template <typename T>
+    friend ReturningValueProgram<T>* create_empty_returning_value_program();
```
yshekel, line 52: same here

icicle/src/program/program_c_api.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+// Destructor
+eIcicleError delete_program(ProgramHandle program)
+{
+  if (!program) { return eIcicleError::INVALID_POINTER; }
+  delete program;
+  return eIcicleError::SUCCESS;
+}
+
+void CONCAT_EXPAND(FIELD, generate_program)(ProgramHandle program, SymbolHandle* parameters_ptr, int nof_parameters)
```
yshekel, line 32: why no return value here?

icicle/src/program/program_c_api.cpp
```diff
@@ -0,0 +1,122 @@
...
+{
+  std::vector<Symbol<scalar_t>> parameters_vec;
+  parameters_vec.reserve(nof_parameters);
+
+  for (int i = 0; i < nof_parameters; i++) {
+    if (parameters_ptr[i] == nullptr) { throw std::invalid_argument("Null pointer found in parameters"); }
```
yshekel, line 38: why throw? you can return an error to user. Avoid exceptions as they are not caught by Rust/Go and make the user experience worse.

icicle/src/program/program_c_api.cpp
```diff
@@ -0,0 +1,122 @@
...
+    parameters_vec.push_back(*parameters_ptr[i]); // TODO replace with span to avoid copying when we switch to cpp20
+  }
+
+  program->m_nof_parameters = nof_parameters;
+  program->generate_program(parameters_vec);
+}
+
+ReturningValueProgramHandle CONCAT_EXPAND(FIELD, create_empty_returning_value_program)()
```
yshekel, line 46: same comment about empty constructors here too

icicle/src/symbol/symbol_api.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+using namespace field_config;
+using namespace icicle;
+
+typedef Symbol<scalar_t>* SymbolHandle;
+
+extern "C" { // TODO remove program from names
+// Symbol functions
+// Constructors
+SymbolHandle CONCAT_EXPAND(FIELD, create_empty_symbol)() { return new Symbol<scalar_t>(); }
```
yshekel, line 13: Can you avoid empty symbol and create it as input directly? Why would you have an empty symbol in Rust?

icicle/src/vec_ops.cpp
```diff
@@ -531,18 +531,56 @@ namespace icicle {
...
   ICICLE_DISPATCHER_INST(ExecuteProgramDispatcher, execute_program, programExecutionImpl)
 
+  // TODO update documentation
   extern "C" eIcicleError CONCAT_EXPAND(FIELD, execute_program)(
-    std::vector<scalar_t*>& data, const Program<scalar_t>& program, uint64_t size, const VecOpsConfig& config)
+    scalar_t** data_ptr,
+    uint64_t nof_params,
+    const Program<scalar_t>* program,
+    uint64_t size,
+    const VecOpsConfig& config)
```
yshekel, line 540: Use pointers instead of references in FFI. We do it everywhere for config too

icicle/src/vec_ops.cpp
```diff
@@ -531,18 +531,56 @@ namespace icicle {
...
   {
-    return ExecuteProgramDispatcher::execute(data, program, size, config);
+    std::vector<scalar_t*> data_vec;
+    data_vec.reserve(nof_params);
+    for (uint64_t i = 0; i < nof_params; i++) {
+      if (data_ptr[i] == nullptr) { throw std::invalid_argument("Null pointer found in parameters"); }
```
yshekel, line 545: again, avoid exceptions and return error instead

wrappers/rust/icicle-core/src/program/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+where
+  F: FieldImpl,
+  S: SymbolTrait<F>
+{
+  fn new_predefined(pre_def: PreDefinedProgram) -> Result<Self, eIcicleError>;
+
+  fn handle(&self) -> Handle;
+}
+
+pub trait ProgramTrait<F, S>:
```
yshekel, line 25: why another trait?

wrappers/rust/icicle-core/src/program/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+pub trait ReturningValueProgramTrait<F, S>:
+  ProgramBaseTrait<F, S> +
+  Sized
+where
+  F:FieldImpl,
+  S: SymbolTrait<F>
+{
+  fn new(program_func: impl FnOnce(&mut Vec<S>) -> S, nof_parameters: u32) -> Result<Self, eIcicleError>;
+
+  fn get_polynomial_degree(&self) -> i32; // COMMENT do we need this on the rust side?
```
yshekel, line 44: Why is this here? First It's a vector API, so why polynomial? Then if you don't use it then remove it.

wrappers/rust/icicle-core/src/program/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+
+      // Programs structs
+      #[repr(C)]
+      pub struct Program
+      {
+        m_handle: Handle
+      }
+
+      #[repr(C)]
+      pub struct ReturningValueProgram {
```
yshekel, line 73: @mickeyasa why do we have to have ReturningValueProgram separate? what is this returned value anyway? Sometimes this one is used in sumcheck?

wrappers/rust/icicle-core/src/program/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+        m_handle: Handle
+      }
+
+       // Program Operations
+      extern "C" {
+        #[link_name = concat!($field_prefix, "_create_empty_program")]
```
yshekel, line 79: I already commented but I don't think this is required. You should be able to construct the program based on predef or symbols

wrappers/rust/icicle-core/src/program/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+  } => {
+    pub(crate) mod test_program {
+      use super::*;
+      use icicle_runtime::test_utilities;
+      use icicle_runtime::{device::Device, runtime};
+      use std::sync_once;
+
+      fn initialize() {
+        test_utilities::test_load_and_init_devices();
+        test_utilities::test_set_main_device();
```
yshekel, line 226: do you only test main device?

yshekel: actually there are no tests here

wrappers/rust/icicle-core/src/symbol/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+  for<'a> Add<&'a Self, Output = Self> +
+  for<'a> Sub<&'a Self, Output = Self> +
+  for<'a> Mul<&'a Self, Output = Self> +
+  for<'a> AddAssign<&'a Self> +
+  for<'a> SubAssign<&'a Self> +
+  for<'a> MulAssign<&'a Self> +
+  Clone + Sized
+{
+  fn new_empty() -> Result<Self, eIcicleError>;
+  fn new_constant(constant: F) -> Result<Self, eIcicleError>;
```
yshekel, line 22: suggest rename to from_constant()

wrappers/rust/icicle-core/src/symbol/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+  fn copy_symbol(other: &Self) -> Result<Self, eIcicleError>;
+
+  fn inverse(&self) -> Self;
+
+  // Functions necessary for the Symbol backend implementation (irrelevant to the user)
+  fn handle(&self) -> SymbolHandle;
```
yshekel, line 28: All of this code here regarding handle and delete etc. should be removed from here!!!
A trait is a generic behavior and you couple the implementation to the trait!!!! This is not how the abstraction should be

wrappers/rust/icicle-core/src/symbol/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+#[macro_export]
+macro_rules! impl_symbol_field {
+  (
+    $field_prefix:literal,
+    $field_prefix_ident:ident,
+    $field:ident,
+    $field_config:ident
+  ) => {
+    pub mod $field_prefix_ident {
+      // use crate::field::{ScalarCfg, ScalarField};
```
yshekel, line 51: remove comment

wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -1,4 +1,6 @@
 use crate::traits::FieldImpl;
+use crate::program::ProgramBaseTrait;
```
yshekel, line 2: It is mandatory to couple program and vecops? Why is it actually here?

wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -973,6 +1066,46 @@ macro_rules! impl_vec_ops_tests {
                 initialize();
                 check_slice::<$field>()
             }
+
+            #[test]
+            pub fn test_program() {
+                initialize();
+                test_utilities::test_set_ref_device();
+                check_program::<$field, Program, Symbol>();
+                initialize();
```
yshekel, line 1075: why init again?

wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -973,6 +1066,46 @@ macro_rules! impl_vec_ops_tests {
...
+                test_utilities::test_set_main_device();
+                check_program::<$field, Program, Symbol>()
+            }
+
+            #[test]
+            pub fn test_returning_value_program() {
+                initialize();
+                test_utilities::test_set_ref_device();
+                check_program_with_return_value::<$field, ReturningValueProgram, Symbol>();
+                initialize();
```
yshekel, line 1085: here too

wrappers/rust/icicle-core/src/vec_ops/tests.rs
```diff
@@ -457,6 +462,195 @@ where
...
+    Symbol: SymbolTrait<F>,
+    Program: ProgramTrait<F, Symbol>,
+{
+    let lambda_eq_x_ab_minus_c = |vars: &mut Vec<Symbol>| {
+        let a = vars[0].clone();
+        let b = vars[1].clone();
+        let c = vars[2].clone();
+        let eq = vars[3].clone();
+
+        let var4_res = eq * (a.clone() * b.clone() - c.clone());
```
yshekel, line 478: So the solution finally is to clone everywhere? this is kind of ugly

yshekel (changes requested):
Sorry but I think the design is not good, especially the traits but not only.
Left many comments.

UPDATE: @mickeyasa @Koren-Brand @LeonHibnik @ShaniBabayoff we decided to remove the symbols from this PR, release only predefined (which is what our clients need at this moment). Next release will support generic programs.

icicle/include/icicle/program/program.h
```diff
@@ -116,6 +116,10 @@ namespace icicle {
     // default constructor
     Program() {}
 
+    // Friend function for C-api to have access to the default constructor
+    template <typename T>
+    friend Program<T>* create_empty_program();
```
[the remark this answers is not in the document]
Koren-Brand: It's used by rust as the cpp constructor requires passing and running a lambda function which I find to be awkward.
So the Rust constructor starts by calling the empty constructor and implements similar logic to the regular cpp constructor but on the rust side.

icicle/include/icicle/program/returning_value_program.h
Koren-Brand: same answer

icicle/src/program/program_c_api.cpp
Koren-Brand: will do

icicle/src/symbol/symbol_api.cpp
Koren-Brand: I was just propagating the public constructors from the cpp side. @mickeyasa  @idanfr-ingo  do you think it's relevant for a rust user?

wrappers/rust/icicle-core/src/program/mod.rs
Koren-Brand: Program and Returning Value Program have different constructors (different lambdas as inputs) that's the reason for this "OOP"-ish choice

wrappers/rust/icicle-core/src/program/mod.rs
Koren-Brand: Yes it will be removed the program tests are in vecops

wrappers/rust/icicle-core/src/vec_ops/mod.rs
Koren-Brand: these are checks that the cpp backend doesn't do, Miki asked me whether or not do I keep this checks or not

wrappers/rust/icicle-core/src/symbol/mod.rs
Koren-Brand: Split of to Handle trait

wrappers/rust/icicle-core/src/vec_ops/mod.rs
Koren-Brand: removed

wrappers/rust/icicle-core/src/program/mod.rs
Koren-Brand: It should fixing it

wrappers/rust/icicle-core/src/program/mod.rs
Koren-Brand: Yes it's for sumcheck
We talked about it in person

wrappers/rust/icicle-core/src/program/mod.rs
Koren-Brand: It's to avoid using the default cpp constructor requiring passing a lambda function to the constructor (Meaning the rust user would have to create a cpp function in the case of the non-empty constructor)

icicle/src/program/program_c_api.cpp
Koren-Brand: Changed to return eIcicleError (which would return success or invalid argument)

icicle/src/symbol/symbol_api.cpp
Koren-Brand: Changed to a create_input_symbol function instead

wrappers/rust/icicle-core/src/program/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+}
+
+pub trait ProgramBaseTrait<F, S>: Sized
+where
+  F: FieldImpl,
+  S: SymbolTrait<F>
+{
+  fn new_predefined(pre_def: PreDefinedProgram) -> Result<Self, eIcicleError>;
+
+  fn handle(&self) -> Handle;
```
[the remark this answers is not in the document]
Koren-Brand: Moved to traits as Handle trait

wrappers/rust/icicle-core/src/program/mod.rs
Koren-Brand: Removed ProgramBaseTrait after splitting of the Handle trait

wrappers/rust/icicle-core/src/program/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+  ProgramBaseTrait<F, S> +
+  Sized
+where
+  F:FieldImpl,
+  S: SymbolTrait<F>
+{
+  fn new(program_func: impl FnOnce(&mut Vec<S>), nof_parameters: u32) -> Result<Self, eIcicleError>;
+}
+
+pub trait ReturningValueProgramTrait<F, S>:
```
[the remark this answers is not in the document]
Koren-Brand: The two traits (Program and ReturningValueProgram) are split as they are different classes in the backend. Additionally, the 2 are used in separate different places (returning for sumcheck while non-returning for vec-ops execution). So the trait split is done to prevent a wrong backend utilized.

docs/docs/icicle/rust-bindings/program.md
```diff
@@ -0,0 +1,149 @@
...
+Additional trait the struct implements besides the `Symbol<F>` trait.
+#### `Handle` 
+Trait to guarantee linking the symbol to the appropriate cpp backend - providing a function to access the pointer to the backend implementation.
+### `Drop`
+Ensures proper resource management by releasing the backend allocated memory when a `Symbol` instance goes out of scope. This prevents memory leaks and ensures that resources are cleaned up correctly, adhering to Rust's RAII (Resource Acquisition Is Initialization) principles.
+
+# Program
+A configurebale program to be ran on the various Icicle backends. It can be one of the members of `PredefinedProgram` enum. The program adheres to the following trait:
+## `Program` Trait Definition
+The trait defines the base functionality required for the user, which in this case is only creation (The other execution functionality is exposed through Vector Operations).
```
Koren-Brand, line 59: Is there documentation for the rust wrapper of the sumcheck?

wrappers/rust/icicle-core/src/vec_ops/mod.rs
Koren-Brand: I'm keeping these checks currently to align with the vecops checks in Rust for now

docs/docs/icicle/rust-bindings/program.md
```diff
@@ -0,0 +1,162 @@
...
+
+:::note
+Please refer to the [Program overview](../primitives/program.md) page for a deep overview. This section is a brief description of the Rust FFI bindings.
+:::
+
+This documantaion is designed to bring developers up to speed about Rust API for program implemented in the cpp backend.
+
+## Introduction
+
+The Rust FFI bindings for both Program d Symbol serve as a "shallow wrapper" around the underlying C++ implementation. These bindings provide a straightforward Rust interface that directly calls functions from a C++ library, effectively bridging Rust and C++ operations. The Rust layer handles simple interface translations without delving into complex logic or data structures, which are managed on the C++ side. This design ensures efficient data handling, memory management, and execution while utilizing the existing backend directly via C++.
```
yshekel, line 11: typo: 'd symbol'

docs/docs/icicle/rust-bindings/program.md
```diff
@@ -0,0 +1,162 @@
...
+```
+### Traits implemented and key methods
+Additional traits the struct implements to fulfil `Symbol<F>` trait that should be noted.
+#### Arithmetic operations
+Symbol implements addition, subtraction and multiplication (as well as the assign variants of them) with other symbols / references as well as field elements. Applying the operations will generate a new symbol (Or overwrite the existing in the case of the assign operation) representing the arithmetic operations of the two operand symbols. The `inverse` function joins these operations to allow an additional arithmetic operation (division).
+#### `DeletableHandle` 
+Trait to guarantee linking the symbol to the appropriate cpp backend - providing a function to access the pointer to the backend implementation, as well as a method (in rust) to delete the relevant memory on the backend side.
+
+# Program
+A configurebale program to be ran on the various Icicle backends. It can be one of the members of `PredefinedProgram` enum. The program adheres to the following trait:
```
yshekel, line 52: What is a configureable program? I think it's programmable, not configurable. I think just drop the 'configureable'.
Also, you 'execute' a program (to be executed)

docs/docs/icicle/rust-bindings/program.md
```diff
@@ -0,0 +1,162 @@
...
+## `Program` Trait Definition
+The trait defines the base functionality required for the user, which in this case is only creation (The execution functionality is exposed through Vector Operations). It is split to 2 traits:
+- `Program` - a program for running a function that takes a vector of field elements (both inputs and outputs) and has no return value (output is written to the given vector). It is executed through Vector Operations.
+- `ReturningValueProgram` - a program for running a function that takes an vector of inputs and returns a single field element. It is used for Sumcheck.
```
yshekel, line 56: typo: an vector--> a vector

yshekel: But also is it true? it doesn't return a single element. I think the lambda itself returns an element, but the computation is outputting a vector. @mickeyasa ?

yshekel: ALso the program is not clear to me. To my understanding the only difference between those is that program is built from a lambda without return value, takes n vectors and computes arbitrary code elementwise. 
ReturningValue expects the lambda, in addition, to also return a value F and when executed this output goes to an 'n+1' implicit element of the program vector.

I think the docs here are confusing.

docs/docs/icicle/rust-bindings/program.md
```diff
@@ -0,0 +1,162 @@
...
+# Usage
+This section will outline how to use Program and Symbol, mirroring the examples from the [cpp overview](../primitives/program.md). The program use-case splits to three steps:
+1. Defining a function/lambda that describes the program to be ran (or choosing one of the predefined list?).
+2. Creating a new program given the above function.
+3. Executing the program using the Vector Operations API.
+## Defining a Function
+```rust
+example_function<F, S>(vars: &mut Vec<S>)
+where
+  F: FieldImpl, // TODO do we need both in types
```
yshekel, line 105: TODO?

docs/docs/icicle/rust-bindings/program.md
```diff
@@ -0,0 +1,162 @@
...
+  S: Symbol<F>,
+{
+  let a = vars[0];
+  let b = vars[1];
+  let c = vars[2];
+  let eq = vars[3];
+
+  vars[4] = eq * (a * b - c) + F::from_u32(9);
```
yshekel, line 113: add a comment here noting that any expression can be implemented here but not control flow. Actually I think this is lacking: I mean what is allowed here and what is not?

docs/docs/icicle/rust-bindings/program.md
```diff
@@ -0,0 +1,162 @@
...
+  vars[5] = a * b - c.inverse();
+  vars[6] = vars[5];
+  vars[3] = (vars[0] + vars[1]) * F::from_u32(2); // all variables can be both inputs and outputs
+}
+```
+One thing to note when comparing this to the cpp version, is the abundance of `clone` calls - due to Symbol not being copyable and Rust's ownership it is required in most operations.
+## Creating a Program
+Applying the constructor with the lambda:
+```rust
+let program = Program::new(example_lambda, 7 /*nof parameters for lambda = vars.size()*/);
```
yshekel, line 123: should be 'example_function' no? please make sure this compiles

docs/docs/icicle/rust-bindings/program.md
```diff
@@ -0,0 +1,162 @@
...
+```
+## Executing the Program
+Execution is done through the appropriate vecops function.
```
yshekel, line 126: Maybe mention that programs work with other ICICLE APIs too

docs/docs/icicle/rust-bindings/program.md
```diff
@@ -0,0 +1,162 @@
...
+```rust
+pub struct Program {
+  handle: ProgramHandle,
+}
+```
+### Traits implemented and key methods
+Additional traits the struct implements to fulfil `Symbol<F>` trait that should be noted.
+#### Arithmetic operations
+Symbol implements addition, subtraction and multiplication (as well as the assign variants of them) with other symbols / references as well as field elements. Applying the operations will generate a new symbol (Or overwrite the existing in the case of the assign operation) representing the arithmetic operations of the two operand symbols. The `inverse` function joins these operations to allow an additional arithmetic operation (division).
+#### `DeletableHandle`
```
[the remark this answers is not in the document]
Koren-Brand: Part of the traits Symbol requires I agree the user shouldn't use it or care about it, but won't it be odd them seeing it in the trait above with no explanation?

docs/docs/icicle/rust-bindings/program.md
```diff
@@ -0,0 +1,162 @@
...
+where
+    F: FieldImpl,
+    <F as FieldImpl>::Config: VecOps<F>,
+    Parameter: HostOrDeviceSlice<F> + ?Sized,
+    Prog: Program<F> + Handle,
+```
+
+And in total with data setup we would get code like this:
+```rust
+let lambda_d_times_ab_minus_c = |vars: &mut Vec<S>| {
```
yshekel, line 142: the name 'lambda_d_times_ab_minus_c' is maybe old? your function does more than that I mean

docs/docs/icicle/rust-bindings/vec-ops.md
```diff
@@ -106,4 +107,15 @@ pub fn bit_reverse_inplace<F>(
     input: &mut (impl HostOrDeviceSlice<F> + ?Sized),
     cfg: &VecOpsConfig,
 ) -> Result<(), eIcicleError>;
+
+pub fn execute_program<F, Prog, Data>(
+    data: &mut Vec<&Data>,
```
yshekel, line 112: why not HostOrDeviceSlice? this won't work with device memory directly

icicle/cmake/target_editor.cmake
```diff
@@ -7,6 +7,9 @@ function(handle_field TARGET)
       src/fields/ffi_extern.cpp
       src/vec_ops.cpp
       src/matrix_ops.cpp
+      src/program/program_c_api.cpp
+      src/program/release_pool.h
```
yshekel, line 11: header files should not be here

icicle/src/program/release_pool.h
```diff
@@ -0,0 +1,36 @@
+#include <set>
```
yshekel, line 1: missing pragma once for header file

icicle/src/program/release_pool.h
```diff
@@ -0,0 +1,36 @@
...
+
+using namespace icicle;
```
yshekel, line 3: avoid using namespace in header files. It makes implicit using when including this.

icicle/src/program/release_pool.h
```diff
@@ -0,0 +1,36 @@
...
+class ReleasePool
+{
+public:
+  static ReleasePool& instance()
+  {
+    thread_local static ReleasePool<T> instance;
+    return instance;
+  }
+
+  void insert(T* ptr) // TODO change to insert
```
yshekel, line 15: remove todo

icicle/src/program/release_pool.h
```diff
@@ -0,0 +1,36 @@
...
+    m_pool.clear();
+  }
+
+private:
+  ReleasePool() = default;
+  ~ReleasePool() = default;
+  ReleasePool(const ReleasePool&) = delete;
+  ReleasePool& operator=(const ReleasePool&) = delete;
+
+  std::set<T*> m_pool; // TODO thread local variable
```
yshekel, line 35: remove TODO, you already have a ReleasePool per thread. ALternatively make this one thread_local and use a single Pool.

icicle/src/program/release_pool.h
yshekel: and actually put this in a namespace, at least inside icicle namespace to avoid name collisions

docs/docs/icicle/rust-bindings/program.md
Koren-Brand: Will add as part of a short explanation before this function

icicle/src/symbol/symbol_api.cpp
```diff
@@ -0,0 +1,151 @@
...
+  auto symbol_ptr = new Symbol<scalar_t>(*constant);
+  ReleasePool<Symbol<scalar_t>>& pool = ReleasePool<Symbol<scalar_t>>::instance();
+  pool.insert(symbol_ptr);
+  return symbol_ptr;
+}
+SymbolHandle CONCAT_EXPAND(FIELD, copy_symbol)(const SymbolHandle other)
+{
+  auto symbol_ptr = new Symbol<scalar_t>(*other);
+  ReleasePool<Symbol<scalar_t>>& pool = ReleasePool<Symbol<scalar_t>>::instance();
+  pool.insert(symbol_ptr);
```
yshekel, line 33: Is this a clone? Otherwise why insert it again?

icicle/src/symbol/symbol_api.cpp
```diff
@@ -0,0 +1,151 @@
...
+  return symbol_ptr;
+}
+
+// Destructor
+eIcicleError delete_symbol(SymbolHandle symbol)
```
yshekel, line 38: Why do you need have it? The release pool handles it for you right? It would be a bug to call this

icicle/src/vec_ops.cpp
yshekel: Why did you resolve it? you didn't answer nor modify the code

docs/docs/icicle/rust-bindings/program.md
Koren-Brand: I'll add to the start

icicle/src/symbol/symbol_api.cpp
Koren-Brand: Yes it's for clone

wrappers/rust/icicle-core/src/program/mod.rs
```diff
@@ -1,58 +1,209 @@
...
 pub enum PreDefinedProgram {
-    ABminusC = 0,
-    EQtimesABminusC,
+  ABminusC = 0,
+  EQtimesABminusC
 }
 
-pub type ReturningValueProgramHandle = *const c_void;
+pub trait Program<F>: 
+  Sized + Handle
```
yshekel, line 14: why is it Sized? what doesn't work without it?

wrappers/rust/icicle-core/src/program/mod.rs
```diff
@@ -1,58 +1,209 @@
...
+where
+  F:FieldImpl,
+{
+  type ProgSymbol: Symbol<F>;
+
+  fn new(program_func: impl FnOnce(&mut Vec<Self::ProgSymbol>), nof_parameters: u32) -> Result<Self, eIcicleError>;
```
yshekel, line 20: I would rename program_func->expression

wrappers/rust/icicle-core/src/program/mod.rs
```diff
@@ -1,58 +1,209 @@
...
+}
+
+pub trait ReturningValueProgram<F>:
+  Sized + Handle
+where
+  F:FieldImpl,
+{
+  type ProgSymbol: Symbol<F>;
+
+  fn new(program_func: impl FnOnce(&mut Vec<Self::ProgSymbol>) -> Self::ProgSymbol, nof_parameters: u32) -> Result<Self, eIcicleError>;
```
yshekel, line 32: same comments here

wrappers/rust/icicle-core/src/program/mod.rs
```diff
@@ -1,58 +1,209 @@
...
+      use icicle_core::program::{ Program, ReturningValueProgram, PreDefinedProgram, Instruction};
+      use icicle_runtime::errors::eIcicleError;
+      use std::ops::{Add, Sub, Mul, AddAssign, SubAssign, MulAssign};
+      use std::ffi::c_void;
+
+      // Programs structs
+      #[repr(C)]
+      pub struct FieldProgram
+      {
+        m_handle: HandleCPP
```
yshekel, line 59: What is handleCPP?

wrappers/rust/icicle-core/src/program/mod.rs
```diff
@@ -1,58 +1,209 @@
...
+
+            program_func(&mut program_parameters);
+
+            let handles: Vec<*const c_void> = program_parameters.iter().map(|s| s.handle()).collect();
+            
+            let prog_handle = ffi_create_empty_program();
+            if prog_handle.is_null() {
+              return Err(eIcicleError::AllocationFailed);
+            }
+            ffi_generate_program(prog_handle, handles.as_ptr(), program_parameters.len() as u32);
```
yshekel, line 115: You don't check return value here?

wrappers/rust/icicle-core/src/program/mod.rs
```diff
@@ -1,58 +1,209 @@
...
+        pub(crate) fn ffi_get_program_polynomial_degree(program: HandleCPP) -> i32;
+      }
+
+      // Program trait implementation
+      impl Program<$field> for FieldProgram {
+        type ProgSymbol = FieldSymbol;
+
+        fn new(program_func: impl FnOnce(&mut Vec<FieldSymbol>), nof_parameters: u32) -> Result<Self, eIcicleError>
+        {
+          unsafe {
```
yshekel, line 101: try to make the unsafe block smaller, only around the ffi

wrappers/rust/icicle-core/src/program/mod.rs
```diff
@@ -1,58 +1,209 @@
...
+            ffi_clear_symbols();
+
+            Ok(Self { m_handle: prog_handle })
+          }
+        }
+
+        fn new_predefined(pre_def: PreDefinedProgram) -> Result<Self, eIcicleError> {
+          unsafe {
+            let prog_handle = ffi_create_predefined_program(pre_def);
+            ffi_clear_symbols();
```
yshekel, line 126: But there are no symbols here? or are there?

wrappers/rust/icicle-core/src/program/mod.rs
yshekel, line 115: Also here you use ffri_create_empty() but for the predefined you directly call ffi_create_predefined_program() so it should be the same here.

icicle/src/program/release_pool.h
Koren-Brand: It's actually redundant - I'm removing it

icicle/src/symbol/symbol_api.cpp
Koren-Brand: You are right there's a double delete called from both rust += operator (and its likes) and the resource pool - I'll delete it

wrappers/rust/icicle-core/src/traits.rs
```diff
@@ -39,6 +39,12 @@ pub trait Arithmetic: Sized + Add<Output = Self> + Sub<Output = Self> + Mul<Outp
...
 
+pub type HandleCPP = *const c_void;
+
 pub trait Handle {
-    fn handle(&self) -> *const c_void;
+    fn handle(&self) -> HandleCPP;
+}
+
+pub trait DeletableHandle: Handle {
+    fn delete_handle(handle: HandleCPP);
```
yshekel, line 49: I think it should be 
fn delete_handle(&mut self);
Otherwise you do something like a.delete(b) to delete b with the object a.

wrappers/rust/icicle-core/src/symbol/mod.rs
```diff
@@ -0,0 +1,346 @@
...
+
+        #[link_name = "delete_symbol"]
+        pub(crate) fn ffi_delete_symbol(symbol: HandleCPP) -> eIcicleError;
+      }
+
+      // Implement Symbol UI
+      impl Symbol<$field> for FieldSymbol {
+        fn new_input(in_idx: u32) -> Result<Self, eIcicleError> {
+          unsafe {
+            let handle = ffi__input_symbol(in_idx);
```
yshekel, line 79: should it be ffi_input_input_symbol with a single underscore?

wrappers/rust/icicle-core/src/symbol/mod.rs
```diff
@@ -0,0 +1,346 @@
...
+
+      impl DeletableHandle for FieldSymbol {
+        fn delete_handle(handle: HandleCPP) {
+          unsafe {
+            if !handle.is_null()
+            {
+              unsafe {
+                let ffi_status = ffi_delete_symbol(handle);
+                if ffi_status != eIcicleError::Success {
+                  panic!("Couldn't delete symbol, due to {:?}", ffi_status);
```
yshekel, line 189: why panic? you can return the error instead

wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -963,6 +1052,22 @@ macro_rules! impl_vec_ops_tests {
                 initialize();
                 check_slice::<$field>()
             }
+
+            #[test]
+            pub fn test_program() {
+                initialize(); // Sets main device
```
yshekel, line 1058: set main explicitly, don't assume this as it can break your code and you won't know

wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -921,13 +1007,16 @@ macro_rules! impl_vec_ops_mixed_field {
...
     (
+      $field_prefix_ident: ident,
       $field:ident
     ) => {
         pub(crate) mod test_vecops {
             use super::*;
             use icicle_runtime::test_utilities;
             use icicle_runtime::{device::Device, runtime};
             use std::sync::Once;
+            // use crate::symbol::$field_prefix_ident::Symbol;
```
yshekel, line 1018: remove

wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -921,13 +1007,16 @@ macro_rules! impl_vec_ops_mixed_field {
...
+            use crate::program::$field_prefix_ident::{FieldProgram, FieldReturningValueProgram};
```
yshekel, line 1019: Can't you use $field to create a program using the program traits? And avoid explicit field module

wrappers/rust/icicle-core/src/program/mod.rs
yshekel: I saw it is defined in traits, commented about it there. I am not sure it should be defined in traits as it is not a trait

wrappers/rust/icicle-core/src/symbol/mod.rs
Koren-Brand: It might be a typo I didn't notice

wrappers/rust/icicle-core/src/program/mod.rs
Koren-Brand: You are correct it must be something I didn't removed when copying

icicle/src/program/program_c_api.cpp
Koren-Brand: will merge into generate

wrappers/rust/icicle-core/src/program/mod.rs
Koren-Brand: For allowing  Result<Self, eIcicleError> as result requires a known size at compile time

wrappers/rust/icicle-core/src/program/mod.rs
Koren-Brand: I prefer to stick to the same names as the cpp functions

wrappers/rust/icicle-core/src/program/mod.rs
Koren-Brand: same answer

wrappers/rust/icicle-core/src/traits.rs
Koren-Brand: the function is removed

wrappers/rust/icicle-core/src/traits.rs
```diff
@@ -39,6 +39,12 @@ pub trait Arithmetic: Sized + Add<Output = Self> + Sub<Output = Self> + Mul<Outp
     fn pow(self, exp: usize) -> Self;
 }
 
+pub type HandleCPP = *const c_void;
```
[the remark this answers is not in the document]
Koren-Brand: Would you prefer not having the type at all?

wrappers/rust/icicle-core/src/program/mod.rs
Koren-Brand: will fix

wrappers/rust/icicle-core/src/program/mod.rs
Koren-Brand: Do you prefer the type removed or renamed?

wrappers/rust/icicle-core/src/vec_ops/mod.rs
Koren-Brand: The type is defined in the impl_program_field macro which has the $field submodule to distinct between field and extension field. Given that program and returning value program are used in two different places (Sumcheck and Vecops) I can't just move the macro here (And it would also be wrong order-wise to have the program implementation under only one of them even if it were to work), so this is how I got it to work

wrappers/rust/icicle-core/src/vec_ops/mod.rs
Koren-Brand: it is defined 37 lines above (where test_set_main_device is explicitly called) and is used in the same way in every other test in vecops - I think it is understandable enough

docs/docs/icicle/rust-bindings/program.md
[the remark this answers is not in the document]
Koren-Brand: Changed to Handle either way as DeletableHandle is removed

wrappers/rust/icicle-core/src/program/mod.rs
yshekel: right. Maybe need to rename there too. Probably will never do. Not big deal but I think it's less clear

wrappers/rust/icicle-core/src/symbol/mod.rs
Koren-Brand: Function removed

wrappers/rust/icicle-core/src/program/mod.rs
yshekel: I think you can define it in the macros internally and rename yes.

wrappers/rust/icicle-core/src/traits.rs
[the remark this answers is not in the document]
yshekel: yes. Move it to the macros internally like other APIs

wrappers/rust/icicle-core/src/traits.rs
yshekel: You don't need it then?

wrappers/rust/icicle-core/src/vec_ops/mod.rs
yshekel: I understand but I don't see why not set the device explicitly. No real downside

wrappers/rust/icicle-core/src/traits.rs
Koren-Brand: nope the deleting happens in the release pool

wrappers/rust/icicle-core/src/sumcheck/tests.rs
```diff
@@ -51,7 +51,7 @@ where
 pub fn check_sumcheck_simple<SW, P>(hash: &Hasher)
 where
     SW: Sumcheck,
-    P: ReturningValueProgram + Handle,
+    P: ReturningValueProgram<<SW as Sumcheck>::Field>,
```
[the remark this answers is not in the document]
Koren-Brand: Changed ReturningValueProgram to have the "old" API

docs/docs/icicle/rust-bindings/program.md
```diff
@@ -0,0 +1,150 @@
+# Rust FFI Bindings for Program
+
+>**_NOTE:_**
+>Please refer to the [Program overview](../primitives/program.md) page for additional detail. This section is a brief description of the Rust FFI bindings.
+
+This documentation is designed to bring developers up to speed about Rust API for program implemented in the cpp backend.
```
idanfr-ingo, line 6: Maybe "backend" is not the best choice here. Why not just use "implemented in cpp" or "for program's cpp implementation".

docs/docs/icicle/rust-bindings/program.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Introduction
+Program is a class that let users define expressions on vector elements, and have ICICLE compile it for the backends for a fused implementation. This solves memory bottlenecks and also let users customize algorithms such as sumcheck. Program can create only element-wise lambda functions. Program itself works for definition while actual execution is handled through other functionalities like [Vector Operations](./vec-ops.md).
+
+
+The Rust FFI bindings for both Program and Symbol serve as a "shallow wrapper" around the underlying C++ implementation. These bindings provide a straightforward Rust interface that directly calls functions from a C++ library, effectively bridging Rust and C++ operations. The Rust layer handles simple interface translations without delving into complex logic or data structures, which are managed on the C++ side. This design ensures efficient data handling, memory management, and execution while utilizing the existing backend directly via C++.
```
idanfr-ingo, line 12: This is some general explanation about the Rust FFI. It does not seem to belong here

docs/docs/icicle/rust-bindings/program.md
```diff
@@ -0,0 +1,150 @@
...
+The `Symbol` struct is implemented for each of the supported icicle fields, implementing the above trait for the specific field (field distinction is relevant for the input symbols and stored constants in the program). In its core it's just a handle to the cpp implementation.
+```rust
+pub struct Symbol {
+  handle: SymbolHandle,
+}
+```
+### Traits implemented and key methods
+Additional traits the struct implements to fulfil `Symbol<F>` trait that should be noted.
+#### Arithmetic operations
+Symbol implements addition, subtraction and multiplication (as well as the assign variants of them) with other symbols / references as well as field elements. Applying the operations will generate a new symbol (Or overwrite the existing in the case of the assign operation) representing the arithmetic operations of the two operand symbols. The `inverse` function joins these operations to allow an additional arithmetic operation (division).
```
idanfr-ingo, line 48: Or in parentheses should begin with lower case "r".

examples/rust/sumcheck/src/main.rs
```diff
@@ -125,7 +125,7 @@ pub fn main() {
     );
     //try different combine functions!
     let combine_function =
-        <icicle_bn254::program::FieldReturningValueProgram as ReturningValueProgram>::new_predefined(
+        <icicle_bn254::program::bn254::FieldReturningValueProgram as ReturningValueProgram>::new_predefined(
```
mickeyasa, line 128: is that a bug?

icicle/include/icicle/program/program.h
```diff
@@ -206,4 +210,10 @@ namespace icicle {
     }
   };
 
+  // Friend function to access the protected default program constructors
```
mickeyasa, line 213: why do you need the default program constructor from rust?

docs/docs/icicle/rust-bindings/program.md
```diff
@@ -0,0 +1,150 @@
...
+The `Program` struct is implemented for each of the supported icicle fields, implementing the above trait for the specific field (field distinction is relevant for the input symbols and stored constants in the program). In its core it's just a handle to the cpp implementation.
+```rust
+pub struct Program {
+  handle: ProgramHandle,
+}
+```
+
+# Usage
+This section will outline how to use Program and Symbol, mirroring the examples from the [cpp overview](../primitives/program.md). The program use-case splits to three steps:
+1. Defining a function/lambda that describes the program to be ran (or choosing one of the predefined list?).
```
idanfr-ingo, line 77: I think you should at list show how to define a program from the pre-defined enum (the same way I did it in the C++ doc).
Also remove the question mark.

docs/docs/icicle/rust-bindings/program.md
```diff
@@ -0,0 +1,150 @@
...
+    cfg: &VecOpsConfig
+) -> Result<(), eIcicleError>
+where
+    F: FieldImpl,
+    <F as FieldImpl>::Config: VecOps<F>,
+    Parameter: HostOrDeviceSlice<F> + ?Sized,
+    Prog: Program<F> + Handle,
+```
+
+And in total with data setup we would get code like this (Example taken from check_program in vec_ops tests):
```
idanfr-ingo, line 121: Example should being with lower case "e"

icicle/include/icicle/program/program.h
Koren-Brand: We talked about it - the other constructor would have you pass a lambda function to it (Which can't be passed from rust to cpp "forcing" the user to write its lambda in cpp). Instead, the empty constructor is called while the rust constructor imitates the original cpp one - creates a symbol vector, runs the lambda, and then generates the program.

docs/docs/icicle/rust-bindings/program.md
```diff
@@ -0,0 +1,150 @@
+# Rust FFI Bindings for Program
+
+>**_NOTE:_**
+>Please refer to the [Program overview](../primitives/program.md) page for additional detail. This section is a brief description of the Rust FFI bindings.
+
+This documentation is designed to bring developers up to speed about Rust API for program implemented in the cpp backend.
+
+## Introduction
+Program is a class that let users define expressions on vector elements, and have ICICLE compile it for the backends for a fused implementation. This solves memory bottlenecks and also let users customize algorithms such as sumcheck. Program can create only element-wise lambda functions. Program itself works for definition while actual execution is handled through other functionalities like [Vector Operations](./vec-ops.md).
```
idanfr-ingo, line 9: I think that it would be nice to make "sumcheck" a link to the relevant section of the sumcheck doc.

wrappers/rust/icicle-curves/icicle-bls12-377/src/program/mod.rs
```diff
@@ -1,3 +1,7 @@
-use icicle_core::impl_program;
+use crate::curve::{ScalarField, BaseField};
 
-impl_program!("bls12_377");
+use icicle_core::impl_program_field;
+
+impl_program_field!("bls12_377", bls12_377, ScalarField, ScalarCfg);
+#[cfg(feature = "bw6-761")]
+impl_program_field!("bw6_761", bw6_761, BaseField, BaseCfg);
```
mickeyasa, line 7: what 761 got to do with 12_377?

wrappers/rust/icicle-curves/icicle-bls12-377/src/symbol/mod.rs
```diff
@@ -0,0 +1,7 @@
+use crate::curve::{ScalarField, BaseField};
+
+use icicle_core::impl_symbol_field;
+
+impl_symbol_field!("bls12_377", bls12_377, ScalarField, ScalarCfg);
+#[cfg(feature = "bw6-761")]
+impl_symbol_field!("bw6_761", bw6_761, BaseField, BaseCfg);
```
mickeyasa, line 7: same here

wrappers/rust/icicle-curves/icicle-bw6-761/src/program/mod.rs
```diff
@@ -0,0 +1,7 @@
+pub mod bw6_761 {
+  // Link program defined using bls12-377's base field to bw6-761
+  use icicle_bls12_377::program::bw6_761::FieldProgram as bw6_761_program;
```
mickeyasa, line 3: what 377 got to do with?

wrappers/rust/icicle-curves/icicle-bw6-761/src/symbol/mod.rs
```diff
@@ -0,0 +1,5 @@
+pub mod bw6_761 {
+  // Link symbol defined using bls12-377's base field to bw6-761
+  use icicle_bls12_377::symbol::bw6_761::FieldSymbol as bw6_761_symbol;
```
mickeyasa, line 3: 377?

wrappers/rust/icicle-curves/icicle-bls12-377/src/program/mod.rs
Koren-Brand: 761's scalar field is 377's base field.
It's similarly defined in other crates.

mickeyasa (changes requested):
comments inside

docs/docs/icicle/rust-bindings/program.md
Koren-Brand: There's a misc addition for sumcheck clarifying that the combine function implements ReturningValueProgram, and links back here. I think that it's enough (with the sumcheck examples also serving as exapmles for predefined returning value programs)

docs/docs/icicle/rust-bindings/program.md
```diff
@@ -0,0 +1,150 @@
...
+
+The following would list the implemented Rust functionality with some examples paralleling those given in the [original program overview](../primitives/program.md).
+# Symbol
+Symbol is the basic (template) class that allow users to define their own program, representing an arithmetic operation. The [function](#defining-a-function-for-program) the user define will operate on symbols.
+## `Symbol` Trait Definition
+The trait defines the functionality required by the user. The expected use-case of symbol is solely to be operated on to create the final arithmetic operation, which is reflected implemented functions and traits.
+```rust
+pub trait Symbol<F: FieldImpl>:
```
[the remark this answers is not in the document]
Koren-Brand: It's a quick way to list all of Symbol's functionality
The user could require a symbol creation in the lambda function for some reason (from_constant for instance)

mickeyasa (approved):

docs/docs/icicle/rust-bindings/program.md
```diff
@@ -0,0 +1,150 @@
+# Rust FFI Bindings for Program
+
+>**_NOTE:_**
```
yshekel, line 3: I think this is not the format for docsaurus. Should be like https://github.com/ingonyama-zk/icicle/blob/6251c3a1555e6e6ea57cac9a8c95d64f7fbf3590/docs/docs/icicle/rust-bindings/polynomials.md?plain=1#L3

docs/docs/icicle/rust-bindings/program.md
```diff
@@ -0,0 +1,150 @@
...
+
+# Usage
+This section will outline how to use Program and Symbol, mirroring the examples from the [cpp overview](../primitives/program.md). The program use-case splits to three steps:
+1. Defining a function/lambda that describes the program to be ran (or choosing one of the predefined list?).
+2. Creating a new program given the above function.
+3. Executing the program using the Vector Operations API.
+## Defining a Function for Program
+A function operating on a vector of symbols, with outputs being written to said input vector. The input symbols in the vector represent inputs and outputs of field elements, and will be replaced by vectors of field elements when executed.
+>**_NOTE:_**
+> The defined function defines arithmetic operations to be done in series, without control-flow i.e. loops, conditions etc.
```
yshekel, line 83: This not accurate and maybe confusing. You can use control flow but the lambda is traced and the path taken is the computation that is later computed.

I suggest "The defined function performs arithmetic operations in a traced execution, meaning control flow (e.g., loops, conditions) is not parsed. Instead, the computation follows the exact execution path taken during tracing, which determines the final computation that will be performed."

docs/docs/icicle/rust-bindings/program.md
yshekel: you did not fix

icicle/include/icicle/program/program.h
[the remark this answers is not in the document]
yshekel: We agreed to remove this

icicle/include/icicle/program/program.h
```diff
@@ -206,4 +210,10 @@ namespace icicle {
...
+  template <typename S>
+  Program<S>* create_empty_program()
```
yshekel, line 215: We agreed to remove this

icicle/include/icicle/program/returning_value_program.h
yshekel: remove

yshekel (changes requested):
You did not resolve a few things. Please do and I will approve

icicle/include/icicle/program/program.h
Koren-Brand: No, we talked about not having it used on the rust side. It is still used it's just not exposed to the FFI.

docs/docs/icicle/rust-bindings/program.md
Koren-Brand: changed to 

:::note
The defined function defines arithmetic operations to be done in series, and could be represented as set of equations (for each output). Practically, control flow (e.g., loops, conditions) is not parsed, instead the computation follows the exact execution path taken during tracing, which determines the final computation that will be performed. 
:::

yshekel (approved):

PR #772, opened 2025-02-12
tests for single poly input

idanfr-ingo:
Add tests for sumcheck with single input program

cuda-backend-branch: idan/sumchek_single_poly


mickeyasa (approved):
great

PR #773, opened 2025-02-12
Koren/predefined program wrapper

Koren-Brand:
Rust wrapper for predefined program added
Fixed extension field missing from backend implementations of program executor

CUDA backend branch (Until it will be merged to main)
cuda-backend-branch: koren/register_program_execution_for_extension_field


icicle/cmake/target_editor.cmake
```diff
@@ -7,6 +7,8 @@ function(handle_field TARGET)
       src/fields/ffi_extern.cpp
       src/vec_ops.cpp
       src/matrix_ops.cpp
+      src/program/program_c_api.cpp
+      src/symbol/symbol_api.cpp
```
yshekel, line 11: can remove symbol for now

icicle/src/program/program_c_api.cpp
```diff
@@ -0,0 +1,122 @@
...
+using namespace icicle;
+
+typedef Symbol<scalar_t>* SymbolHandle;
+typedef Program<scalar_t>* ProgramHandle;
+typedef ReturningValueProgram<scalar_t>* ReturningValueProgramHandle;
+
+extern "C" {
+// Program functions
+// Constructor
+ProgramHandle CONCAT_EXPAND(FIELD, create_empty_program)() { return create_empty_program<scalar_t>(); }
```
yshekel, line 17: why do you need it if you have create_predefined_program below?

icicle/src/program/program_c_api.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+// Destructor
+eIcicleError delete_program(ProgramHandle program)
+{
+  if (!program) { return eIcicleError::INVALID_POINTER; }
+  delete program;
+  return eIcicleError::SUCCESS;
+}
+
+void CONCAT_EXPAND(FIELD, generate_program)(ProgramHandle program, SymbolHandle* parameters_ptr, int nof_parameters)
```
yshekel, line 32: remove from this PR

icicle/src/program/program_c_api.cpp
```diff
@@ -0,0 +1,122 @@
...
+{
+  return create_empty_returning_value_program<scalar_t>();
+}
+
+ReturningValueProgramHandle CONCAT_EXPAND(FIELD, create_predefined_returning_value_program)(PreDefinedPrograms pre_def)
+{
+  return new ReturningValueProgram<scalar_t>(pre_def);
+}
+
+int CONCAT_EXPAND(FIELD, get_program_polynomial_degree)(ReturningValueProgramHandle program)
```
yshekel, line 56: I think we said remove this right?

icicle/src/symbol/symbol_api.cpp
```diff
@@ -0,0 +1,111 @@
+#include "icicle/program/symbol.h"
```
yshekel, line 1: we agreed to remove symbol from PR

wrappers/rust/icicle-core/src/merkle/mod.rs
```diff
@@ -211,6 +211,7 @@ impl MerkleTree {
             );
 
             if handle.is_null() {
+                println!("Null handle!");
```
yshekel, line 214: remove

wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -261,6 +273,31 @@ fn check_vec_ops_args_slice<F>(
...
+    cfg: &VecOpsConfig
+) -> VecOpsConfig
+where
+    F: FieldImpl,
+    <F as FieldImpl>::Config: VecOps<F>,
+    Data: HostOrDeviceSlice<F> + ?Sized,
+{
+    // All parameters' config should match so each one is compared to the first one
+    let nof_iterations = data[0].len();
+    let is_on_device = data[0].is_on_device();
```
yshekel, line 287: do you assume everything is on device based on the first? may be a bug

wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -261,6 +273,31 @@ fn check_vec_ops_args_slice<F>(
...
+    data: &Vec<&Data>,
+    cfg: &VecOpsConfig
+) -> VecOpsConfig
+where
+    F: FieldImpl,
+    <F as FieldImpl>::Config: VecOps<F>,
+    Data: HostOrDeviceSlice<F> + ?Sized,
+{
+    // All parameters' config should match so each one is compared to the first one
+    let nof_iterations = data[0].len();
```
yshekel, line 286: what iterations?

wrappers/rust/icicle-core/src/vec_ops/mod.rs
yshekel: what about batch? missing error

wrappers/rust/icicle-core/src/program/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+pub trait ProgramBaseTrait<F>: Sized
+where
+  F: FieldImpl,
+{
+  fn new_predefined(pre_def: PreDefinedProgram) -> Result<Self, eIcicleError>;
+  fn handle(&self) -> Handle;
+}
+
+// COMMENT is there even a difference between returning and non returning value when it is predefined?
+pub trait ProgramTrait<F>:
```
yshekel, line 23: and those two traits are empty so why do you need them?
I think you simply need a

struct Program {
handle
}

And then implement the constructors:

impl Program {
 fn from_predefined(...) -> Self             
 fn from_custom_program(...) -> Self     // FUTURE
}

Also it's not template in the type of the field.

icicle/src/vec_ops.cpp
```diff
@@ -531,18 +531,56 @@ namespace icicle {
 
   ICICLE_DISPATCHER_INST(ExecuteProgramDispatcher, execute_program, programExecutionImpl)
 
+  // TODO update documentation
```
yshekel, line 534: remove comment

icicle/src/program/program_c_api.cpp
Koren-Brand: You're right I forgot to reduce the cpp api

icicle/src/program/program_c_api.cpp
Koren-Brand: Nope it is required for non predefined programs - I'll remove it

icicle/src/program/program_c_api.cpp
Koren-Brand: Will do (same for the extension field below)

wrappers/rust/icicle-core/src/vec_ops/mod.rs
Koren-Brand: I'm checking that all the slices are in the same place (Host / Device) you can see it in the for loop below
Same for the size
What about batch? its existence is to maintain the same config as the other vecops but given that program is an "element-wise" operation batch has no meaning besides being a multiplier to size provided in cpp. However, given that both batch and size are deduced implicitly and not by user input, I decided to keep batch as 1 and size as the length of the inputs.

PR #780, opened 2025-02-17
Remove MerkleTreeBasic test

Koren-Brand:
Removed MerkleTreeBasic test and its hash function HashSumBackend, as the latter does not serve as a good hash function which caused false failures in edge cases where randomly modifying the leaves can still possibly lead to the same output.

Changes:
Removed MerkleTreeBasic test from test_hash_api
Removed associated hash HashSumBackend


mickeyasa (approved):
V

LeonHibnik (approved):

PR #787, opened 2025-02-23
Implement Integer Rings in RNS Representation

yshekel:
This PR introduces a template class for integer rings in the Residue Number System (RNS) representation. Instead of storing an integer modulo a composite modulus q = p1 * p2 * ... * pn, this implementation represents the integer as a tuple (x mod p1, x mod p2, ..., x mod pn), where each component resides in a different field.

- [x] Implement ZqRns type - RNS representation of ring elements
- [x] Instantiate vecops, ntt dispatchers for rns type
- [x] register vecops CPU (and CUDA and backend PR)
- [x] register ntt CPU (and CUDA and backend PR)
- [x] rns <--> direct conversions, including vector API and test

Future PRs:

- Large ring (geryhound)
- Rust bindings
- Additional APIs

cuda-backend-branch: yshekel/ring_rns


icicle/backend/cpu/src/field/cpu_vec_ops.cpp
```diff
@@ -1005,4 +999,21 @@ REGISTER_VECTOR_PRODUCT_EXT_FIELD_BACKEND("CPU", cpu_vector_product<extension_t>
...
+
+#ifdef RING
+REGISTER_MATRIX_TRANSPOSE_RING_RNS_BACKEND("CPU", cpu_matrix_transpose<scalar_rns_t>);
+REGISTER_BIT_REVERSE_RING_RNS_BACKEND("CPU", cpu_bit_reverse<scalar_rns_t>);
+REGISTER_SLICE_RING_RNS_BACKEND("CPU", cpu_slice<scalar_rns_t>);
+REGISTER_VECTOR_ADD_RING_RNS_BACKEND("CPU", cpu_vector_add<scalar_rns_t>);
+REGISTER_VECTOR_ACCUMULATE_RING_RNS_BACKEND("CPU", cpu_vector_accumulate<scalar_rns_t>);
+REGISTER_VECTOR_SUB_RING_RNS_BACKEND("CPU", cpu_vector_sub<scalar_rns_t>);
+REGISTER_VECTOR_MUL_RING_RNS_BACKEND("CPU", (cpu_vector_mul<scalar_rns_t, scalar_rns_t>));
+REGISTER_VECTOR_DIV_RING_RNS_BACKEND("CPU", cpu_vector_div<scalar_rns_t>);
```
mickeyasa, line 1012: does it have the same interface?
return 0 if there is no inverse?

yshekel: yes

yshekel: Fixed it, now non-invertible elements return 0 so it's consistent with the direct representation.

icicle/include/icicle/rings/integer_ring_rns.h
```diff
@@ -0,0 +1,315 @@
...
+  template <size_t I>
+  using field_type = std::tuple_element_t<I, Fields>;
+
+  storage<RNS_CONFIG::limbs_count> limbs_storage;
+
+  // Stream Output
+  friend std::ostream& operator<<(std::ostream& os, const IntegerRingRns& xs)
+  {
+    std::stringstream hex_string;
+    hex_string << std::hex << std::setfill('0');
```
mickeyasa, line 34: I think it will be nicer to divide to the sub fields?
{0x1234, 0x1234}, {0x1234, 0x1234}, {0x1234, 0x1234}
i.e. you can use the << operator of field

icicle/include/icicle/rings/integer_ring_rns.h
```diff
@@ -0,0 +1,315 @@
...
+    for (int i = 0; i < limbs_count; i++) {
+      hex_string << std::setw(8) << xs.limbs_storage.limbs[limbs_count - i - 1];
+    }
+
+    os << "0x" << hex_string.str();
+    return os;
+  }
+
+  // Access Field Elements
+  template <size_t I>
```
mickeyasa, line 45: template?
constexpr will do the trick as well.

mickeyasa (approved):
some comments inside

icicle/include/icicle/rings/integer_ring_rns.h
yshekel: I don't understand what you mean here. It's used as get_field<i> to get the i'th field with correct type. Each is another type so this must be template.
Regarding constexpr, I can add it but it still should be template.

icicle/include/icicle/rings/integer_ring_rns.h
yshekel: Good point. I will do that

yshekel: done

icicle/include/icicle/rings/integer_ring_rns.h
```diff
@@ -0,0 +1,314 @@
...
+  {
+    return has_inverse_impl(x, std::make_index_sequence<nof_fields>{});
+  }
+
+  static HOST_DEVICE_INLINE IntegerRingRns pow(const IntegerRingRns& x, const IntegerRingRns& y)
+  {
+    return apply_binary_op(x, y, [](auto x, auto y) { return x.pow(x, y); }, std::make_index_sequence<nof_fields>{});
+  }
+
+  static HOST_DEVICE_INLINE IntegerRingRns to_montgomery(const IntegerRingRns& x)
```
emirsoyturk, line 187: just a small naming suggestion: we can use `to_mont` and `from_mont` like the way with ScalarField

icicle/include/icicle/rings/integer_ring_rns.h
```diff
@@ -0,0 +1,314 @@
...
+  }
+
+  template <typename T>
+  static constexpr bool has_member_omegas_count()
+  {
+    return sizeof(T::omegas_count) > 0;
+  }
+
+  // Arithmetic Operators
+  friend HOST_DEVICE_INLINE IntegerRingRns operator+(const IntegerRingRns& a, const IntegerRingRns& b)
```
emirsoyturk, line 130: a general comment: do we need to add compound assignments. we also don't have it for Fields

emirsoyturk (approved):

icicle/include/icicle/rings/integer_ring_rns.h
yshekel: We could, I am not sure it's more efficient but we could when we need it

PR #794, opened 2025-03-02
Miki/msm-optimiations

mickeyasa:
avoid scalar copy
improve c and nof-cores formula by using a decision tree
added decision tree to decide about the optimal C
run time reduced by 30-40% for bn354, bls 377


icicle/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -42,22 +42,28 @@ class Msm
   }
 
   // run MSM based on pippenger algorithm
-  void run_msm(const scalar_t* scalars, const A* bases, P* results)
+  void run_msm(const scalar_t* scalars, const A* bases, P* results, bool last_task_on_batch)
```
yshekel, line 45: I think it;s last_task_in_batch? instead of on_batch

icicle/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -42,22 +42,28 @@ class Msm
...
+
     phase2_collapse_segments();
 
-    // TBD: start the next task in batch in parallel to phase 3
-    phase3_final_accumulator(results);
+    if (last_task_on_batch) {
+      phase3_final_accumulator(results);
+      m_phase3_thread = nullptr;
+    } else {
+      m_phase3_thread = new std::thread(&Msm::phase3_final_accumulator, this, results);
```
yshekel, line 57: I don't see that you release it so you leak memory. Why not a unique_ptr? you can still have it nullptr

icicle/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -217,18 +226,35 @@ class Msm
     }
   }
 
+  //   // phase 2: accumulate m_segment_size buckets into a line_sum and triangle_sum
```
yshekel, line 229: do you need this code?

yshekel (changes requested):
Can you please elaborate which case you checked (curve, size) and the before/after performance?

icicle/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -217,18 +250,35 @@ class Msm
...
+  //     run_workers_and_wait();
+  //   }
+
   // phase 2: accumulate m_segment_size buckets into a line_sum and triangle_sum
   void phase2_collapse_segments()
   {
-    uint64_t bucket_start = 0;
-    for (int segment_idx = 0; segment_idx < m_segments.size();
-         segment_idx++) { // TBD: divide the work among m_nof_workers only.
+    //    for (int segment_idx = 0; segment_idx < m_segments.size();
```
idanfr-ingo, line 274: delete

mickeyasa: done

idanfr-ingo (approved):

LeonHibnik (approved):

jeremyfelder (commented):
Can we remove the spaces from the `decision tree params` directory name?

icicle/include/icicle/utils/platform.h
```diff
@@ -0,0 +1,69 @@
...
+#endif
+
+inline bool is_apple_silicon()
+{
+  struct utsname unameData;
+  if (uname(&unameData) == 0) {
+    std::string machine(unameData.machine);
+    return machine == "arm64" || machine == "aarch64";
+  }
+  return false;
```
jeremyfelder, line 20: What about running on android devices that are arm based?

mickeyasa: will be examined in the future - currently we are using the AMD as default.

icicle/include/icicle/utils/platform.h
```diff
@@ -0,0 +1,69 @@
...
+  #include <cpuid.h>
+  #include <stdint.h>
+#endif
+
+inline bool is_apple_silicon()
+{
+  struct utsname unameData;
+  if (uname(&unameData) == 0) {
+    std::string machine(unameData.machine);
+    return machine == "arm64" || machine == "aarch64";
```
yshekel, line 18: it seems that is_arm_platform() is what this function does

yshekel:
consider also pushing the python code that generates the params (and maybe a short readme), in case we need to extend it in the future

yshekel (approved):

LeonHibnik (approved):

PR #795, opened 2025-03-02
feat: FRI merkle tree API + CPU Backend

ShanieWinitz:
## Describe the changes

This PR implements FRI frontend, CPU, GPU backend

cuda-backend-branch: feat/fri


icicle/backend/metal
yshekel, line 1: remove submodules. Metal and also taskflow

icicle/include/icicle/fri/fri.h
```diff
@@ -0,0 +1,221 @@
...
+   */
+  template <typename S, typename F>
+  class Fri
+  {
+  public:
+    /**
+     * @brief Constructor for the Fri class.
+     * @param backend A shared pointer to the backend (FriBackend<S, F>) responsible for FRI operations.
+     */
+    explicit Fri(std::shared_ptr<FriBackend<S, F>> backend) : m_backend(std::move(backend)) {}
```
yshekel, line 72: why move a shared ptr? don't do that

icicle/include/icicle/fri/fri.h
```diff
@@ -0,0 +1,221 @@
...
+     * @param fri_transcript_config Configuration for encoding/hashing (Fiat-Shamir).
+     * @param fri_proof The proof object to verify.
+     * @param verification_pass (OUT) Set to true if verification succeeds, false otherwise.
+     * @return An eIcicleError indicating success or failure.
+     */
+    eIcicleError verify(
+      const FriConfig& fri_config,
+      const FriTranscriptConfig<F>&& fri_transcript_config,
+      FriProof<F>& fri_proof,
+      bool& verification_pass /* OUT */) const
```
yshekel, line 103: 'pass' is a little strange. Let's make it consistent with other APIs. for merkle it's simply 'valid'

icicle/include/icicle/fri/fri.h
```diff
@@ -0,0 +1,221 @@
...
+    {
+      verification_pass = false;
+      ICICLE_ASSERT(fri_config.nof_queries > 0) << "No queries specified in FriConfig.";
```
yshekel, line 106: replace asserts with errors. No need to crash for no reason

icicle/include/icicle/fri/fri.h
```diff
@@ -0,0 +1,221 @@
...
+      const uint32_t log_input_size = nof_fri_rounds + static_cast<uint32_t>(std::log2(final_poly_size));
+      const size_t input_size = 1 << (log_input_size);
+      std::vector<F> alpha_values(nof_fri_rounds);
+
+      // set up the transcript
+      FriTranscript<F> transcript(
+        std::move(const_cast<FriTranscriptConfig<F>&>(fri_transcript_config)), log_input_size);
+      for (size_t round_idx = 0; round_idx < nof_fri_rounds; ++round_idx) {
+        auto [root_ptr, root_size] = fri_proof.get_root(round_idx);
+        ICICLE_ASSERT(root_ptr != nullptr && root_size > 0) << "Failed to retrieve Merkle root for round " << round_idx;
```
yshekel, line 118: same here and generally in software.

icicle/include/icicle/fri/fri.h
ShanieWinitz: I did like in hash.h, sumcheck.h, merkle_tree.h ... How should it be done?

icicle/backend/cpu/include/cpu_fri_backend.h
```diff
@@ -0,0 +1,181 @@
...
+  template <typename S, typename F>
+  class CpuFriBackend : public FriBackend<S, F>
+  {
+  public:
+    /**
+     * @brief Constructor that accepts an existing array of Merkle trees.
+     *
+     * @param folding_factor   The factor by which the codeword is folded each round.
+     * @param stopping_degree  Stopping degree threshold for the final polynomial.
+     * @param merkle_trees     A vector of MerkleTrees.
```
yshekel, line 24: vector<MerkleTree> is clearly a vector of merkle tree but I think you need to mention that the trees are one per fri step.

icicle/backend/cpu/include/cpu_fri_backend.h
```diff
@@ -0,0 +1,181 @@
...
+     */
+    CpuFriBackend(const size_t folding_factor, const size_t stopping_degree, std::vector<MerkleTree> merkle_trees)
```
yshekel, line 26: usually we don't pass vectors by value. Is there are reason to do so? you can pass a const reference or move it if you want

icicle/backend/cpu/include/cpu_fri_backend.h
```diff
@@ -0,0 +1,181 @@
...
+        : FriBackend<S, F>(folding_factor, stopping_degree, merkle_trees), m_nof_fri_rounds(merkle_trees.size()),
+          m_log_input_size(merkle_trees.size() + std::log2(static_cast<double>(stopping_degree + 1))),
+          m_input_size(pow(2, m_log_input_size)), m_fri_rounds(merkle_trees, m_log_input_size)
+    {
+    }
+
+    eIcicleError get_fri_proof(
```
yshekel, line 33: I think other APIs are 'get_proof' so please be consistent

icicle/backend/cpu/include/cpu_fri_backend.h
```diff
@@ -0,0 +1,181 @@
...
+      const FriConfig& fri_config,
+      const FriTranscriptConfig<F>&& fri_transcript_config,
+      const F* input_data,
+      FriProof<F>& fri_proof /*out*/) override
+    {
+      ICICLE_ASSERT(fri_config.nof_queries > 0) << "Number of queries must be > 0";
```
yshekel, line 39: Get rid of all the asserts and replace with errors. It's not more efficient since assert is also checking the condition. Also note this assert is also compiled in release mode.

Then, no need to crash the user program (for example Eigen have a cloud service and you will make them crash for no reason).

This is true for this entire PR.

icicle/backend/cpu/include/cpu_fri_backend.h
```diff
@@ -0,0 +1,181 @@
...
+
+      FriTranscript<F> transcript(std::move(fri_transcript_config), m_log_input_size);
+
+      // Initialize the proof
+      fri_proof.init(fri_config.nof_queries, m_nof_fri_rounds, this->m_stopping_degree + 1);
+
+      // commit fold phase
+      ICICLE_CHECK(commit_fold_phase(input_data, transcript, fri_config, fri_proof));
```
yshekel, line 47: same here, return the error. You should only use asserts as a last resort, but usually it's not desired

icicle/backend/cpu/include/cpu_fri_backend.h
```diff
@@ -0,0 +1,181 @@
...
+
+      // proof of work
+      if (fri_config.pow_bits != 0) { ICICLE_CHECK(proof_of_work(transcript, fri_config.pow_bits, fri_proof)); }
+
+      // query phase
+      ICICLE_CHECK(query_phase(transcript, fri_config, fri_proof));
+
+      return eIcicleError::SUCCESS;
```
yshekel, line 55: You see, so you either crash or return success. In that case you don't even need a return value since it crashes anyway if not successful.

icicle/backend/cpu/include/cpu_fri_backend.h
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+  private:
+    const size_t m_nof_fri_rounds; // Number of FRI rounds
```
yshekel, line 59: No need to comment everything if the comment doesn't give any information. If names are not clear, replace them. Here for example it's quite clear that 'm_nof_fri_rounds' is the number of fri rounds

icicle/backend/cpu/include/cpu_fri_backend.h
```diff
@@ -0,0 +1,181 @@
...
+     * @return eIcicleError Error code indicating success or failure.
+     */
+    eIcicleError commit_fold_phase(
+      const F* input_data, FriTranscript<F>& transcript, const FriConfig& fri_config, FriProof<F>& fri_proof)
+    {
+      ICICLE_ASSERT(this->m_folding_factor == 2)
+        << "Currently only folding factor of 2 is supported"; // TODO SHANIE - remove when supporting other folding
+                                                              // factors
+
+      const S* twiddles = ntt_cpu::CpuNttDomain<S>::s_ntt_domain.get_twiddles();
```
yshekel, line 79: (1) You should check that the domain exists and large enough or return an error.
(2) This coupling of fri and ntt is not ideal. I wonder how to avoid it.

icicle/backend/cpu/include/cpu_fri_backend.h
```diff
@@ -0,0 +1,181 @@
...
+      uint64_t domain_max_size = ntt_cpu::CpuNttDomain<S>::s_ntt_domain.get_max_size();
+
+      // Get persistent storage for round from FriRounds. m_fri_rounds already allocated a vector for each round with
```
yshekel, line 82: what do you mean by persistent storage?

icicle/backend/cpu/include/cpu_fri_rounds.h
```diff
@@ -0,0 +1,74 @@
...
+  {
+  public:
+    /**
+     * @brief Constructor that accepts an already-existing array of Merkle trees.
+     *        Ownership is transferred from the caller.
+     *
+     * @param merkle_trees A vector of MerkleTrees.
+     * @param log_input_size The log of the input size.
+     */
+    FriRounds(std::vector<MerkleTree> merkle_trees, const size_t log_input_size) : m_merkle_trees(merkle_trees)
```
yshekel, line 26: same comment here about the clone of vector

icicle/backend/cpu/include/cpu_fri_rounds.h
```diff
@@ -0,0 +1,74 @@
+#pragma once
```
yshekel, line 1: I think it's cleaner to move those header to src/fri as it's not really a public header. Not that important though but I think cleaner (I know there are a few headers there, I would move them too.

icicle/cmake/target_editor.cmake
```diff
@@ -107,3 +107,14 @@ function(handle_sumcheck TARGET FEATURE_LIST)
     set(SUMCHECK OFF CACHE BOOL "SUMCHECK not available for this field" FORCE)
   endif()
 endfunction()
+
+function(handle_fri TARGET FEATURE_LIST)
+  if(FRI AND "FRI" IN_LIST FEATURE_LIST)
+    target_compile_definitions(${TARGET} PUBLIC FRI=${FRI})
+    target_sources(${TARGET} PRIVATE src/fri/fri.cpp src/fri/fri_c_api.cpp)
+    target_link_libraries(${TARGET} PRIVATE icicle_hash)
```
yshekel, line 115: Why do you need to link to icicle_hash but sumcheck doesn't? They both rely on hashes.

icicle/include/icicle/backend/fri_backend.h
```diff
@@ -0,0 +1,135 @@
...
+    /**
+     * @brief Generate the FRI proof from given inputs.
+     *
+     * @param fri_config            Configuration for FRI operations (e.g., proof-of-work bits, queries).
+     * @param fri_transcript_config Configuration for encoding/hashing FRI messages (Fiat-Shamir).
+     * @param input_data          Evaluations of the polynomial (or other relevant data).
+     * @param fri_proof           (OUT) A FriProof object to store the proof's Merkle layers, final poly, etc.
+     * @return eIcicleError       Error code indicating success or failure.
+     */
+    virtual eIcicleError get_fri_proof(
```
yshekel, line 49: rename get_fri_proof -> get_proof

icicle/include/icicle/fri/fri.h
yshekel: I see now. Actually it's not an issue but it's strange. You can ignore it

icicle/include/icicle/fri/fri.h
```diff
@@ -0,0 +1,221 @@
...
+   * @param merkle_tree_compress_hash The hash function used for compressing Merkle tree nodes.
+   * @param output_store_min_layer (Optional) The layer at which to store partial results. Default = 0.
+   * @return A `Fri<F>` object built around the chosen backend.
+   */
+  template <typename S, typename F>
+  Fri<S, F> create_fri(
+    const size_t input_size,
+    const size_t folding_factor,
+    const size_t stopping_degree,
+    const Hash& merkle_tree_leaves_hash,
```
yshekel, line 42: Note that Hash should be copied, not referenced. So replace 'const Hash&' to 'Hash'. We have a few special cases like this. Underneath is simply a shared_ptr and usually it should be copied to share ownership.
I assume you later copy it into a merkle tree so it's fine but still I think this should be 'Hash' here

icicle/include/icicle/fri/fri.h
```diff
@@ -0,0 +1,221 @@
...
+          }
+          current_log_size--;
+        }
+      }
+      verification_pass = true;
+      return eIcicleError::SUCCESS;
+    }
+
+  private:
+    std::shared_ptr<FriBackend<S, F>> m_backend; // Shared pointer to the backend for FRI operations.
```
yshekel, line 218: again, it's minor but no need to comment obvious comments

icicle/include/icicle/fri/fri_config.h
```diff
@@ -0,0 +1,37 @@
...
+   * execution modes (synchronous/asynchronous), and device/host data placement.
+   * It also supports backend-specific extensions for customization.
+   */
+  struct FriConfig {
+    icicleStreamHandle stream = nullptr; // Stream for asynchronous execution. Default is nullptr.
+    size_t pow_bits = 0;                 // Number of leading zeros required for proof-of-work. Default is 0.
+    size_t nof_queries = 1;              // Number of queries, computed for each folded layer of FRI. Default is 1.
+    bool are_inputs_on_device =
+      false; // True if inputs reside on the device (e.g., GPU), false if on the host (CPU). Default is false.
+    bool are_outputs_on_device = false; // True if outputs reside on the device, false if on the host. Default is false.
```
yshekel, line 22: what is output here? The proof is always on host right?

icicle/include/icicle/fri/fri_proof.h
```diff
@@ -0,0 +1,101 @@
...
+        throw std::out_of_range("Invalid round index");
+      }
+      return m_query_proofs[query_idx][round_idx];
+    }
+
+    /**
+     * @brief Returns a pair containing the pointer to the root data and its size.
+     * @return A pair of (root data pointer, root size).
+     */
+    std::pair<const std::byte*, std::size_t> get_root(const size_t round_idx) const
```
yshekel, line 66: what root?

icicle/include/icicle/fri/fri_proof.h
```diff
@@ -0,0 +1,101 @@
...
+    size_t get_nof_fri_rounds() const { return m_query_proofs[0].size(); }
+
+    /**
+     * @brief Get the final poly size.
+     *
+     * @return final_poly_size.
+     */
+    size_t get_final_poly_size() const { return m_final_poly_size; }
+
+    void set_pow_nonce(uint64_t pow_nonce) { m_pow_nonce = pow_nonce; }
```
yshekel, line 85: can you move this to init? and make m_pow_nonce const?

icicle/include/icicle/fri/fri_proof.h
```diff
@@ -0,0 +1,101 @@
...
+    uint64_t get_pow_nonce() const { return m_pow_nonce; }
+
+    // get pointer to the final polynomial
+    F* get_final_poly() const { return m_final_poly.get(); }
+
+  private:
+    std::vector<std::vector<MerkleProof>>
+      m_query_proofs; // Matrix of Merkle proofs [query][round] - contains path, root, leaf. for each query, we have 2
+                      // proofs (for the leaf in [2*query] and its symmetric in [2*query+1])
+    std::unique_ptr<F[]> m_final_poly; // Final polynomial (constant in canonical FRI)
```
yshekel, line 96: This can be a vector too and then you don't need an additional 'm_final_poly_size

icicle/include/icicle/merkle/merkle_tree.h
```diff
@@ -75,7 +75,7 @@ namespace icicle {
     inline eIcicleError
     build(const T* leaves, uint64_t nof_leaves /* number of leaf elements */, const MerkleTreeConfig& config)
     {
-      return build(reinterpret_cast<const std::byte*>(leaves), nof_leaves * sizeof(T), config);
+      return build(reinterpret_cast<const std::byte*>(leaves), nof_leaves * m_backend->get_leaf_element_size(), config);
```
yshekel, line 78: I think this is super confusing for users. I would revert this change. Otherwise you expect the user to make 'nof_leaves' not the number of T elements. If you need, add another build method but leave this one as is.

icicle/include/icicle/utils/rand_gen.h
```diff
@@ -14,4 +15,31 @@ static uint32_t rand_uint_32b(uint32_t min = 0, uint32_t max = UINT32_MAX)
...
+  return dist(rand_generator);
+}
+
+/**
+ * @brief Generate random unsigned integer in range (inclusive)
+ * @param min Lower limit.
+ * @param max Upper limit.
+ * @return Random (uniform distribution) unsigned integer s.t. min <= integer <= max.
+ */
+static std::vector<size_t> rand_size_t_vector(size_t nof_queries, size_t min = 0, size_t max = SIZE_MAX)
```
yshekel, line 38: I think move this to where you need it. It's a strange api, especially with the input 'size_t nof_queries'

icicle/src/fri/fri_c_api.cpp
```diff
@@ -0,0 +1,308 @@
...
+  std::byte* public_state;
+  size_t public_state_len;
+  const scalar_t* seed_rng;
+};
+
+/**
+ * @brief Structure representing creation parameters for the "hash-based" constructor
+ *        `create_fri<scalar_t>(folding_factor, stopping_degree, Hash&, output_store_min_layer)`.
+ */
+struct FriCreateHashFFI {
```
yshekel, line 39: HashFFI?

icicle/src/fri/fri_c_api.cpp
```diff
@@ -0,0 +1,308 @@
+#include "icicle/fields/field_config.h"
```
yshekel, line 1: This code is not used right? as there is no FFI. I think it should be in a future PR

yshekel (changes requested):
I mostly reviewed APIs, not the internals

icicle/tests/test_mod_arithmetic_api.h
```diff
@@ -17,6 +17,11 @@
 #include "../backend/cpu/include/cpu_program_executor.h"
 #include "icicle/sumcheck/sumcheck.h"
 
+#include "icicle/fri/fri.h"
```
yshekel, line 20: why is this here? should be in test_field_api right?

icicle/include/icicle/fri/fri_proof.h
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * @brief Get a reference to a specific Merkle proof.
+     *
+     * @param query_idx Index of the query.
+     * @param round_idx Index of the round (FRI round).
+     * @return Reference to the Merkle proof at the specified position.
+     */
+    MerkleProof& get_query_proof(const size_t query_idx, const size_t round_idx)
```
yshekel, line 53: It's not the full query but a single stage right? maybe name is confusing

icicle/include/icicle/fri/fri_transcript.h
```diff
@@ -0,0 +1,251 @@
...
+      m_first_round = true;
+    }
+
+    /**
+     * @brief Add a Merkle commit to the transcript and generate a new alpha challenge.
+     *
+     * @param merkle_commit The raw bytes of the Merkle commit.
+     * @return A field element alpha derived via Fiat-Shamir.
+     */
+    F get_alpha(const std::vector<std::byte>& merkle_commit)
```
yshekel, line 34: This should be two methods I think. Why get_alpha() also appends data?

icicle/include/icicle/fri/fri_transcript.h
```diff
@@ -0,0 +1,251 @@
...
+
+      const Hash& hasher = m_transcript_config.get_hasher();
+      std::vector<std::byte> hash_result(hasher.output_size());
+      hasher.hash(hash_input.data(), hash_input.size(), m_hash_config, hash_result.data());
+      uint64_t seed = bytes_to_uint_64(hash_result);
+      return seed;
+    }
+
+  private:
+    const FriTranscriptConfig<F>&& m_transcript_config; // Transcript configuration (labels, seeds, etc.)
```
yshekel, line 97: why is this an rvalue reference member? should be 'const FriTranscriptConfig<F>

icicle/include/icicle/fri/fri_transcript.h
```diff
@@ -0,0 +1,251 @@
...
+    const HashConfig m_hash_config;                     // hash config - default
```
yshekel, line 98: This is not input so why is it a member? You can generate it when you need it, when hashing. Please make the code as simple and short as possible. And again, bad comments

icicle/include/icicle/fri/fri_transcript.h
```diff
@@ -0,0 +1,251 @@
...
+    bool m_first_round;                                 // Indicates if this is the first round
```
yshekel, line 99: If you can infer this, better. Otherwise you need to update the state and risk being in a bad state

icicle/include/icicle/fri/fri.h
```diff
@@ -0,0 +1,221 @@
...
+
+    /**
+     * @brief Verify a FRI proof.
+     * @param fri_config Configuration for FRI operations.
+     * @param fri_transcript_config Configuration for encoding/hashing (Fiat-Shamir).
+     * @param fri_proof The proof object to verify.
+     * @param verification_pass (OUT) Set to true if verification succeeds, false otherwise.
+     * @return An eIcicleError indicating success or failure.
+     */
+    eIcicleError verify(
```
yshekel, line 99: try to refactor this method with smaller calls. It's quite long and hard to read

icicle/tests/test_field_api.cpp
```diff
@@ -576,6 +576,124 @@ TEST_F(FieldTestBase, SumcheckSingleInputProgram)
...
+  fri_config.pow_bits = pow_bits;
+  FriProof<TypeParam> fri_proof;
+
+  ICICLE_CHECK(prover_fri.get_fri_proof(fri_config, std::move(transcript_config), scalars.get(), fri_proof));
+
+  // ===== Verifier side ======
+  Fri verifier_fri = create_fri<scalar_t, TypeParam>(
+    input_size, folding_factor, stopping_degree, hash, compress, output_store_min_layer);
+  bool verification_pass = false;
+  ICICLE_CHECK(verifier_fri.verify(fri_config, std::move(transcript_config), fri_proof, verification_pass));
```
yshekel, line 624: You move the transcript_config again here. This is a bug, it's undefined behavior.
Also, why use default config in the test? better test it

icicle/tests/test_field_api.cpp
```diff
@@ -576,6 +576,124 @@ TEST_F(FieldTestBase, SumcheckSingleInputProgram)
...
+  Fri verifier_fri = create_fri<scalar_t, TypeParam>(folding_factor, stopping_degree, merkle_trees);
+  bool verification_pass = false;
+  ICICLE_CHECK(verifier_fri.verify(fri_config, std::move(transcript_config), fri_proof, verification_pass));
+
+  ASSERT_EQ(true, verification_pass);
+
+  // Release domain
+  ICICLE_CHECK(ntt_release_domain<scalar_t>());
+}
+#endif // FRI
```
yshekel, line 695: Missing tests for errors cases, to test it returns errors as expected.
Also failed verification due to different labels or things like that should be tested. It's not enough to test a simple case

icicle/backend/cpu/include/cpu_fri_backend.h
ShanieWinitz: merkle_trees is now moved into CpuFriBackend when calling create_fri. It is then passed by reference to fri_rounds.

icicle/backend/cpu/include/cpu_fri_backend.h
ShanieWinitz: changed the comment to:
      // Retrieve pre-allocated memory for the round from m_fri_rounds. 
      // The instance of FriRounds has already allocated a vector for each round with 
      // a capacity of 2^(m_log_input_size - round_idx).

icicle/backend/cpu/include/cpu_fri_rounds.h
ShanieWinitz: Here I now changed it to pass a reference

icicle/cmake/features.cmake
```diff
@@ -2,19 +2,19 @@
 # Define available field libraries with an index and their supported features
 # Format: index:field:features
 set(ICICLE_FIELDS
-  1001:babybear:NTT,EXT_FIELD,POSEIDON,POSEIDON2,SUMCHECK
-  1002:stark252:NTT,POSEIDON,POSEIDON2,SUMCHECK
+  1001:babybear:NTT,EXT_FIELD,POSEIDON,POSEIDON2,SUMCHECK,FRI
+  1002:stark252:NTT,POSEIDON,POSEIDON2,SUMCHECK,FRI
   1003:m31:EXT_FIELD,POSEIDON,POSEIDON2,SUMCHECK
-  1004:koalabear:NTT,EXT_FIELD,POSEIDON,POSEIDON2,SUMCHECK
+  1004:koalabear:NTT,EXT_FIELD,POSEIDON,POSEIDON2,SUMCHECK,FRI
```
mickeyasa, line 8: do er have fri for koalbear but not for m31?

icicle/cmake/features.cmake
```diff
@@ -2,19 +2,19 @@
...
 set(ICICLE_CURVES
-  1:bn254:NTT,MSM,G2,ECNTT,POSEIDON,POSEIDON2,SUMCHECK
-  2:bls12_381:NTT,MSM,G2,ECNTT,POSEIDON,POSEIDON2,SUMCHECK
-  3:bls12_377:NTT,MSM,G2,ECNTT,POSEIDON,POSEIDON2,SUMCHECK
-  4:bw6_761:NTT,MSM,G2,ECNTT,POSEIDON,POSEIDON2,SUMCHECK
+  1:bn254:NTT,MSM,G2,ECNTT,POSEIDON,POSEIDON2,SUMCHECK,FRI
+  2:bls12_381:NTT,MSM,G2,ECNTT,POSEIDON,POSEIDON2,SUMCHECK,FRI
+  3:bls12_377:NTT,MSM,G2,ECNTT,POSEIDON,POSEIDON2,SUMCHECK,FRI
+  4:bw6_761:NTT,MSM,G2,ECNTT,POSEIDON,POSEIDON2,SUMCHECK,FRI
   5:grumpkin:MSM,POSEIDON,POSEIDON2,SUMCHECK
```
mickeyasa, line 18: same q for grumpkin

icicle/cmake/target_editor.cmake
ShanieWinitz: FRI uses Merkle trees, which require icicle_hash, while sumcheck does not seem to directly depend on Merkle trees. Removing icicle_hash from fri caused a linker error due to an undefined reference to create_merkle_tree.

icicle/include/icicle/fri/fri_config.h
ShanieWinitz: Removed

icicle/include/icicle/fri/fri_config.h
```diff
@@ -0,0 +1,37 @@
...
+    bool is_async = false; // True to run operations asynchronously, false to run synchronously. Default is false.
+    ConfigExtension* ext = nullptr; // Pointer to backend-specific configuration extensions. Default is nullptr.
```
mickeyasa, line 24: do you plan to support batch?

icicle/include/icicle/fri/fri_proof.h
ShanieWinitz: changed to "get_merkle_tree_root(round_idx)"

icicle/include/icicle/fri/fri_proof.h
```diff
@@ -0,0 +1,101 @@
...
+    // Constructor
+    FriProof() : m_pow_nonce(0) {}
+
+    /**
+     * @brief Initialize the Merkle proofs and final polynomial storage.
+     *
+     * @param nof_queries Number of queries in the proof.
+     * @param nof_fri_rounds Number of FRI rounds (rounds).
+     */
+    void init(const size_t nof_queries, const size_t nof_fri_rounds, const size_t final_poly_size)
```
mickeyasa, line 32: who suppose to use this function?
the user or the backend?
why not move it to the constructor?

icicle/include/icicle/fri/fri_proof.h
```diff
@@ -0,0 +1,101 @@
...
+
+    // get pointer to the final polynomial
+    F* get_final_poly() const { return m_final_poly.get(); }
+
+  private:
+    std::vector<std::vector<MerkleProof>>
+      m_query_proofs; // Matrix of Merkle proofs [query][round] - contains path, root, leaf. for each query, we have 2
+                      // proofs (for the leaf in [2*query] and its symmetric in [2*query+1])
+    std::unique_ptr<F[]> m_final_poly; // Final polynomial (constant in canonical FRI)
+    size_t m_final_poly_size;          // Size of the final polynomial
```
mickeyasa, line 97: do we need that at the proof?

icicle/include/icicle/fri/fri_transcript.h
```diff
@@ -0,0 +1,251 @@
...
+
+      // Hash the input and return alpha
+      const Hash& hasher = m_transcript_config.get_hasher();
+      std::vector<std::byte> hash_result(hasher.output_size());
+      hasher.hash(hash_input.data(), hash_input.size(), m_hash_config, hash_result.data());
+      m_prev_alpha = F::from(hash_result.data(), hasher.output_size());
+      return m_prev_alpha;
+    }
+
+    size_t hash_and_get_nof_leading_zero_bits(uint64_t nonce)
```
mickeyasa, line 58: who use this function?
does it have to be public?

mickeyasa: same question to all functions below

icicle/include/icicle/fri/fri_transcript_config.h
```diff
@@ -0,0 +1,107 @@
...
+   *
+   * @tparam F Type of the field element (e.g., prime field or extension field elements).
+   */
+  template <typename F>
+  class FriTranscriptConfig
+  {
+  public:
+    // Default Constructor
+    FriTranscriptConfig()
+        : m_hasher(create_keccak_256_hash()), m_domain_separator_label(cstr_to_bytes("ds")),
```
mickeyasa, line 22: why are these the default values? why not empty?

icicle/src/fri/fri.cpp
```diff
@@ -0,0 +1,183 @@
...
+namespace icicle {
+
+  using FriFactoryScalar = FriFactoryImpl<scalar_t, scalar_t>;
+  ICICLE_DISPATCHER_INST(FriDispatcher, fri_factory, FriFactoryScalar);
+  /**
+   * @brief Create a FRI instance.
+   * @return A `Fri<scalar_t>` object built around the chosen backend.
+   */
+  template <typename S, typename F>
+  Fri<S, F> create_fri_with_merkle_trees(
```
mickeyasa, line 21: do we want to expose this api? 
We don't support that flexibility

icicle/include/icicle/fri/fri_proof.h
ShanieWinitz: No, the nonce is computed during the proof of work stage and saved in the proof.

icicle/include/icicle/merkle/merkle_tree.h
ShanieWinitz: I expect the user to set nof_leaves as the number of leaves, and if each leaf is of size sizeof(T), then the total size calculation remains the same.

Wouldn't it be clearer to explicitly write:
leaves_size = nof_leaves * m_backend->get_leaf_element_size()
rather than:
leaves_size = nof_leaves * sizeof(T)?

icicle/tests/test_mod_arithmetic_api.h
ShanieWinitz: Ok, moved also #include "icicle/sumcheck/sumcheck.h" to test_field_api

icicle/include/icicle/fri/fri_transcript.h
ShanieWinitz: The alpha generation is done according to the merkle_commit, it should be added to the hash input. I inserted this line to build_hash_input_round_0 and build_hash_input_round_i, maybe it will be more clear that way.

icicle/include/icicle/fri/fri_transcript.h
ShanieWinitz: Did it like in sumcheck

icicle/cmake/features.cmake
ShanieWinitz: We have Fri where we  have NTT

icicle/cmake/features.cmake
ShanieWinitz: Same

icicle/include/icicle/fri/fri_config.h
ShanieWinitz: It was listed as an open question in the design document, Let's decide?

icicle/include/icicle/fri/fri_proof.h
ShanieWinitz: We create an instance of FriProof, which is then passed by reference to prover_fri.get_proof() and later to verifier_fri.verify(). fri_proof.init() is done based on the parameters received when creating the prover instance and parameters recived when calling get_proof. So init cannot be part of the FriProof constructor.

ShanieWinitz: Added this explanation to the function description.

icicle/include/icicle/fri/fri_proof.h
ShanieWinitz: Yes, but I changed m_final_poly to a vector so m_final_poly_size is removed anyway.

icicle/include/icicle/fri/fri_transcript.h
ShanieWinitz: Those functions are used by the prover.  
Additionally, I understood from Jeremy that an implementation for PoW was recently added to Icicle, so he (or Timur) will replace it in their PR.

icicle/include/icicle/fri/fri_transcript_config.h
ShanieWinitz: I'll consoult with Karthik.

icicle/src/fri/fri.cpp
ShanieWinitz: We don't support different Merkle tree arities, but a user can already choose to use different hash functions per layer. If we decide to expose this, I need to add error logging for cases where the tree arity is not 2. (Those logs are currently missing for that API. I'll add them).

icicle/cmake/features.cmake
mickeyasa: We can do NTT on M31, no?

icicle/cmake/features.cmake
mickeyasa: we can do up to 2^28

icicle/src/fri/fri.cpp
mickeyasa: Ok.
Note that if you see that there are too many checks and limitations on the user, you can add this API only after you support that flexibility.

mickeyasa (approved):

icicle/cmake/features.cmake
yshekel: No NTT in M31, it has some DCCT instead but it's not implemented in ICICLE V3, only for the stwo V2 branch.

icicle/backend/cpu/include/cpu_fri_backend.h
yshekel: But it is copied here, not moved. I am not sure I understand what you mean

yshekel: In Sumcheck we understood that if we want to avoid copying we can pass it to the prover/verifier rather than the constructor. Maybe it's not a big deal here. It's copying logn pointers so maybe it's not too bad but it is copying.

icicle/backend/cpu/include/cpu_fri_backend.h
```diff
@@ -0,0 +1,199 @@
...
+      const FriConfig& fri_config,
+      const FriTranscriptConfig<F>& fri_transcript_config,
+      const F* input_data,
+      FriProof<F>& fri_proof /*out*/) override
+    {
+      if (__builtin_expect(fri_config.nof_queries <= 0, 0)) {
+        ICICLE_LOG_ERROR << "Number of queries must be > 0";
+        return eIcicleError::INVALID_ARGUMENT;
+      }
+      if (__builtin_expect(this->m_folding_factor != 2, 0)) {
```
yshekel, line 44: I think you need to check this in the constructor, it's not even an argument of this method so it's strange to fail here for that.

icicle/backend/cpu/include/cpu_fri_backend.h
```diff
@@ -0,0 +1,199 @@
...
+     * @brief Constructor that accepts an existing array of Merkle trees.
+     *
+     * @param folding_factor   The factor by which the codeword is folded each round.
+     * @param stopping_degree  Stopping degree threshold for the final polynomial.
+     * @param merkle_trees     A vector of MerkleTrees, tree per FRI round.
+     */
+    CpuFriBackend(const size_t folding_factor, const size_t stopping_degree, std::vector<MerkleTree> merkle_trees)
+        : FriBackend<S, F>(folding_factor, stopping_degree, std::move(merkle_trees)),
+          m_nof_fri_rounds(this->m_merkle_trees.size()),
+          m_log_input_size(this->m_merkle_trees.size() + std::log2(static_cast<double>(stopping_degree + 1))),
```
yshekel, line 29: Why is this the log_size? number of trees is probably the number of rounds, so does it assume folding factor is 2?

icicle/backend/cpu/include/cpu_fri_backend.h
```diff
@@ -0,0 +1,199 @@
...
+          m_input_size(pow(2, m_log_input_size)), m_fri_rounds(this->m_merkle_trees, m_log_input_size)
+    {
+    }
+
+    eIcicleError get_proof(
+      const FriConfig& fri_config,
+      const FriTranscriptConfig<F>& fri_transcript_config,
+      const F* input_data,
+      FriProof<F>& fri_proof /*out*/) override
```
yshekel, line 38: I think it is probably better to get the size of input_data here as another parameter. You assume it is valid but maybe it's not. If not, you need to identify and log an error.

icicle/backend/cpu/include/cpu_fri_backend.h
```diff
@@ -0,0 +1,199 @@
...
+    /**
+     * @brief Perform the commit-fold phase of the FRI protocol.
+     *
+     * @param input_data  The initial polynomial evaluations.
+     * @param fri_proof   The proof object to update.
+     * @param transcript  The transcript to generate challenges.
+     * @return eIcicleError Error code indicating success or failure.
+     */
+    eIcicleError commit_fold_phase(
+      const F* input_data, FriTranscript<F>& transcript, const FriConfig& fri_config, FriProof<F>& fri_proof)
```
yshekel, line 87: I think you should accept the size and not assume it it valid when you take a pointer like that

icicle/backend/cpu/include/cpu_fri_backend.h
```diff
@@ -0,0 +1,199 @@
...
+
+        current_size >>= 1;
+        current_log_size--;
+      }
+      return eIcicleError::SUCCESS;
+    }
+
+    eIcicleError proof_of_work(FriTranscript<F>& transcript, const size_t pow_bits, FriProof<F>& fri_proof)
+    {
+      for (uint64_t nonce = 0; nonce < UINT64_MAX; nonce++) {
```
yshekel, line 152: nonce is only 64b? Why don't you use the pow API here?

icicle/backend/cpu/include/cpu_fri_rounds.h
```diff
@@ -0,0 +1,68 @@
...
+      return m_merkle_trees[round_idx].get_merkle_root();
+    }
+
+  private:
+    // Persistent polynomial evaluations for each round (heap allocated).
+    // For round i, the expected length is 2^(m_initial_log_size - i).
+    std::vector<std::unique_ptr<F[]>> m_rounds_evals;
+
+    // Holds MerkleTree for each round. m_merkle_trees[i] is the tree for round i.
+    std::vector<MerkleTree>& m_merkle_trees;
```
yshekel, line 65: I think it's a bad idea to store a reference to another object like that since you don't control its lifetime and it may be released too soon (before the FriRounds object). In such cases you need to either pass by value, move or use a smart pointer. Even if this is not exposed to users, it  still creates coupling and assumptions for code using this class and changes in other places may break this code.

icicle/include/icicle/backend/fri_backend.h
```diff
@@ -0,0 +1,135 @@
...
+     * @param fri_config            Configuration for FRI operations (e.g., proof-of-work bits, queries).
+     * @param fri_transcript_config Configuration for encoding/hashing FRI messages (Fiat-Shamir).
+     * @param input_data          Evaluations of the polynomial (or other relevant data).
+     * @param fri_proof           (OUT) A FriProof object to store the proof's Merkle layers, final poly, etc.
+     * @return eIcicleError       Error code indicating success or failure.
+     */
+    virtual eIcicleError get_proof(
+      const FriConfig& fri_config,
+      const FriTranscriptConfig<F>& fri_transcript_config,
+      const F* input_data,
```
yshekel, line 52: I think you need to add input_size here

icicle/include/icicle/backend/fri_backend.h
```diff
@@ -0,0 +1,135 @@
...
+  class FriBackend
+  {
+  public:
+    /**
+     * @brief Constructor that accepts an existing array of Merkle trees.
+     *
+     * @param folding_factor   The factor by which the codeword is folded each round.
+     * @param stopping_degree  Stopping degree threshold for the final polynomial.
+     */
+    FriBackend(const size_t folding_factor, const size_t stopping_degree, std::vector<MerkleTree> merkle_trees)
```
yshekel, line 33: You probably need to check the values are valid here. Since it's a constructor it may be problematic to fail here. Why are those not part of the config?

icicle/include/icicle/fri/fri_transcript.h
```diff
@@ -0,0 +1,256 @@
...
+
+  template <typename F>
+  class FriTranscript
+  {
+  public:
+    FriTranscript(const FriTranscriptConfig<F>& transcript_config, const uint32_t log_input_size)
+        : m_transcript_config(transcript_config), m_prev_alpha(F::zero()), m_pow_nonce(0)
+    {
+      m_entry_0.clear();
+      m_entry_0.reserve(1024); // pre-allocate some space
```
yshekel, line 22: don't hardcode values like that. Define it somewhere above

icicle/include/icicle/fri/fri_proof.h
```diff
@@ -0,0 +1,109 @@
...
+     * the prover and generating the proof.
+     *
+     * @param nof_queries Number of queries in the proof.
+     * @param nof_fri_rounds Number of FRI rounds (rounds).
+     * @param final_poly_size Size of the final polynomial.
+     * @return An eIcicleError indicating success or failure.
+     */
+    eIcicleError init(const size_t nof_queries, const size_t nof_fri_rounds, const size_t final_poly_size)
+    {
+      if (nof_queries <= 0 || nof_fri_rounds <= 0) {
```
yshekel, line 40: no upper bound limit?

icicle/include/icicle/fri/fri_proof.h
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /**
+     * @brief Get a reference to a specific Merkle proof for a given query index in a specific FRI round.
+     *
+     * @param query_idx Index of the query.
+     * @param round_idx Index of the round (FRI round).
+     * @return Reference to the Merkle proof at the specified position.
+     */
+    MerkleProof& get_query_proof(const size_t query_idx, const size_t round_idx)
```
yshekel, line 62: Is it not const by design?

icicle/tests/test_field_api.cpp
```diff
@@ -595,6 +600,179 @@ TEST_F(FieldTestBase, SumcheckSingleInputProgram)
...
+      const uint64_t output_store_min_layer = 0;
+
+      // Generate input polynomial evaluations
+      auto scalars = std::make_unique<TypeParam[]>(input_size);
+      TypeParam::rand_host_many(scalars.get(), input_size);
+
+      auto run = [log_input_size, input_size, folding_factor, stopping_degree, output_store_min_layer, nof_queries,
+                  pow_bits, &scalars](const std::string& dev_type, bool measure) {
+        Device dev = {dev_type, 0};
+        icicle_set_device(dev);
```
yshekel, line 634: need ICICLE_CHECK around icicle APIs in tests so it fails if this fails

icicle/tests/test_field_api.cpp
```diff
@@ -595,6 +600,179 @@ TEST_F(FieldTestBase, SumcheckSingleInputProgram)
...
+
+        // Initialize ntt domain
+        NTTInitDomainConfig init_domain_config = default_ntt_init_domain_config();
+        ICICLE_CHECK(ntt_init_domain(scalar_t::omega(log_input_size), init_domain_config));
+
+        // ===== Prover side ======
+        uint64_t merkle_tree_arity = 2; // TODO SHANIE (future) - add support for other arities
```
yshekel, line 641: For now this is not supported right? Do you check it?

icicle/tests/test_field_api.cpp
```diff
@@ -595,6 +600,179 @@ TEST_F(FieldTestBase, SumcheckSingleInputProgram)
...
+
+        // Define hashers for merkle tree
+        Hash hash = Keccak256::create(sizeof(TypeParam));                          // hash element -> 32B
+        Hash compress = Keccak256::create(merkle_tree_arity * hash.output_size()); // hash every 64B to 32B
+
+        Fri prover_fri = create_fri<scalar_t, TypeParam>(
```
yshekel, line 647: I see you only test the case of hash+compress. Do you currently expose the other API too? It's fine if not, but if you do then you must test it as well

icicle/tests/test_field_api.cpp
```diff
@@ -595,6 +600,179 @@ TEST_F(FieldTestBase, SumcheckSingleInputProgram)
...
+
+      ASSERT_EQ(true, valid);
+    }
+    ASSERT_EQ(error, eIcicleError::INVALID_ARGUMENT);
+
+    // Release domain
+    ICICLE_CHECK(ntt_release_domain<scalar_t>());
+  };
+
+  run(IcicleTestBase::reference_device(), 0 /*nof_queries*/, 2 /*folding_factor*/, log_input_size /*log_domain_size*/);
```
yshekel, line 765: (1) Please comment why each case should fail
(2) I think there are more invalid cases like tree arity > 2

icicle/tests/test_field_api.cpp
```diff
@@ -595,6 +600,179 @@ TEST_F(FieldTestBase, SumcheckSingleInputProgram)
...
+      pow_bits = 16;
+      nof_queries = 100;
+    } else {
+      log_stopping_size = 8;
+      pow_bits = 0;
+      nof_queries = 50;
+    }
+    for (size_t log_input_size = 16; log_input_size <= 24; log_input_size += 4) {
+      const size_t input_size = 1 << log_input_size;
+      const size_t folding_factor = 2; // TODO SHANIE (future) - add support for other folding factors
```
yshekel, line 622: I think you assume this when you compute the input size in the cpu backend. I commented there that it is a mistake. Also if you don't support then missing test to assert it fails.
Anyway I think you need to pass the logsize of size of data to the fri prover and not implicitly assume things like that and compute it. If you are wrong you may have segmentation faults or issues like that

icicle/include/icicle/fri/fri.h
```diff
@@ -0,0 +1,345 @@
...
+     * @param fri_config Configuration for FRI operations (e.g., proof-of-work, queries).
+     * @param fri_transcript_config Configuration for encoding/hashing (Fiat-Shamir).
+     * @param input_data Evaluations or other relevant data for constructing the proof.
+     * @param fri_proof Reference to a FriProof object (output).
+     * @return An eIcicleError indicating success or failure.
+     */
+    eIcicleError get_proof(
+      const FriConfig& fri_config,
+      const FriTranscriptConfig<F>& fri_transcript_config,
+      const F* input_data,
```
yshekel, line 74: missing size here

icicle/include/icicle/fri/fri.h
```diff
@@ -0,0 +1,345 @@
...
+
+    /**
+     * @brief Generate a FRI proof from the given polynomial evaluations (or input data).
+     * @param fri_config Configuration for FRI operations (e.g., proof-of-work, queries).
+     * @param fri_transcript_config Configuration for encoding/hashing (Fiat-Shamir).
+     * @param input_data Evaluations or other relevant data for constructing the proof.
+     * @param fri_proof Reference to a FriProof object (output).
+     * @return An eIcicleError indicating success or failure.
+     */
+    eIcicleError get_proof(
```
yshekel, line 71: So we expose only this constructor for now?

icicle/include/icicle/fri/fri.h
yshekel: I actually now see you pass it to create_fri() but do you use it? I may be confused here. I commented about it multiple times

icicle/src/fri/fri.cpp
```diff
@@ -0,0 +1,140 @@
...
+    const size_t folding_factor, const size_t stopping_degree, std::vector<MerkleTree> merkle_trees)
+  {
+    std::shared_ptr<FriBackend<S, F>> backend;
+    ICICLE_CHECK(FriDispatcher::execute(folding_factor, stopping_degree, std::move(merkle_trees), backend));
```
yshekel, line 17: I don't this this move does anything. It probably copies

yshekel (changes requested):
Better but still I commented, I think there are a few issues still.

icicle/include/icicle/fri/fri_proof.h
yshekel: Each query has a proof per round for two values? Does it assume bit reverse vectors?

icicle/backend/cpu/include/cpu_fri_backend.h
ShanieWinitz: As we discussed, we want the merkle_trees in the constructor, so I guess we will keep the coping of the pointers. @mickeyasa

icicle/include/icicle/fri/fri_transcript.h
```diff
@@ -0,0 +1,249 @@
...
+     * @return Random (uniform distribution) unsigned integer s.t. min <= integer <= max.
+     */
+    std::vector<size_t> rand_queries_indicies(size_t nof_queries, size_t min, size_t max, eIcicleError& err)
+    {
+      // Prepare a buffer for hashing
+      std::vector<std::byte> hash_input;
+      hash_input.reserve(1024); // pre-allocate some space
+
+      // Build the hash input
+      build_hash_input_query_phase(hash_input);
```
yshekel, line 105: name not clear. I had to go back and forth to understand what it does. You only use it once too so I think you'd better inline the logic here to make code easier to read

icicle/include/icicle/fri/fri_transcript.h
```diff
@@ -0,0 +1,249 @@
...
+
+    /**
+     * @brief Generates random query indices for the query phase.
+     *        The seed is derived from the current transcript state.
+     * @param nof_queries Number of query indices to generate.
+     * @param min Lower limit.
+     * @param max Upper limit.
+     * @return Random (uniform distribution) unsigned integer s.t. min <= integer <= max.
+     */
+    std::vector<size_t> rand_queries_indicies(size_t nof_queries, size_t min, size_t max, eIcicleError& err)
```
yshekel, line 98: It seems that the way we generate queries is unique to ICICLE. In that case only ICICLE verifier is consistent with ICICLE FRI prover. Is this by design? are we aware of this? am I wrong?

@mickeyasa @ShanieWinitz @krakhit

icicle/backend/cpu/include/cpu_fri_backend.h
ShanieWinitz: for now we only support folding_factor 2. I left a TODO comment and as we said we return invalid argument if user is trying to do something else

icicle/backend/cpu/include/cpu_fri_backend.h
ShanieWinitz: With the new API its not relevant right?

icicle/backend/cpu/include/cpu_fri_backend.h
ShanieWinitz: With the new API its not relevant right?

icicle/backend/cpu/include/cpu_fri_backend.h
ShanieWinitz: Merged this change

icicle/include/icicle/fri/fri_transcript.h
krakhit: does it not use a seeded query generator
https://github.com/ingonyama-zk/research_POCs/blob/b1ccaf48d29e57dde5c615fa5451497d7ff343c0/fri_poc/src/prover.rs#L110

then u can produce the exact query index in the verifier side 
https://github.com/ingonyama-zk/research_POCs/blob/b1ccaf48d29e57dde5c615fa5451497d7ff343c0/fri_poc/src/verifier.rs#L128

and the rest Is like sumcheck verifier, just use same config, fiat Shamir config etc, and get same random challenge. 

Although, in above the Merkle proof actually contains the index, I just did this to sanity check the query index produced by verifier FS+seeded sampler is same as in that of the index extracted from the proofs. 

I used a rng based on chacha for this https://github.com/ingonyama-zk/research_POCs/blob/b1ccaf48d29e57dde5c615fa5451497d7ff343c0/fri_poc/src/utils.rs#L144

btw: I need to actually make sure seedbytes comes from extension fields, and max>0
I haven't done that yet.

krakhit: But you are right that it depends on the sampler used. 

Do we have a list of seeded samplers we can support ?

icicle/include/icicle/fri/fri_proof.h
ShanieWinitz: added upper bound limit and moved that check to the get_proof/ verify in fri.cpp

icicle/include/icicle/fri/fri_proof.h
ShanieWinitz: not assuming bit reverse vectors.
added in the brief "Each query includes a proof for two values per round."

ShanieWinitz: It is not const, added to the description:
 * This function returns a reference to a pre-allocated Merkle proof in the `m_query_proofs` array.
 * The proof is initially empty and will be populated by another function responsible for generating 
 * the actual proof data.

and changed the name to: "get_query_proof_slot"

icicle/tests/test_field_api.cpp
ShanieWinitz: yes

icicle/tests/test_field_api.cpp
ShanieWinitz: I removed them for this PR

icicle/include/icicle/fri/fri.h
ShanieWinitz: API changed

icicle/include/icicle/fri/fri.h
ShanieWinitz: yes

icicle/include/icicle/fri/fri_transcript.h
ShanieWinitz: As we decided, I left it this way for now.

icicle/include/icicle/fri/fri.h
```diff
@@ -0,0 +1,414 @@
...
+  eIcicleError verify_fri_mt(
+    const FriConfig& fri_config,
+    const FriTranscriptConfig<F>& fri_transcript_config,
+    FriProof<F>& fri_proof,
+    Hash merkle_tree_leaves_hash,
+    Hash merkle_tree_compress_hash,
+    const uint64_t output_store_min_layer,
+    bool& valid /* OUT */);
+
+  struct FRI {
```
yshekel, line 81: why struct? you mean namespace?

yshekel: I think it should be 'namespace fri_merkle_tree' and then it becomes for users fri_merkle_tree::prove() and fri_merkle_tree::verify(). 

In any case note that we have name conventions, for example structs Should be Fri, not FRI etc.

icicle/include/icicle/fri/fri.h
```diff
@@ -0,0 +1,414 @@
...
+
+  /**
+   * @brief Class for performing FRI operations.
+   *
+   * This class provides a high-level interface for constructing and managing a FRI proof.
+   *
+   * @tparam F The field type used in the FRI protocol.
+   */
+  template <typename S, typename F>
+  class Fri
```
yshekel, line 122: Do you want users to use this class? if not, move to fri.cpp

icicle/include/icicle/fri/fri.h
yshekel: I would still take size but if you move it to fri.cpp then that's not an issue

docs/versioned_docs/version-3.5.0/icicle/primitives/fri.md
LeonHibnik, line 1: Move out of versioned docs @ShanieWinitz, this only relates to v3.5 and we don't have fri in this version

docs/versioned_docs/version-3.5.0/icicle/primitives/fri.md
```diff
@@ -0,0 +1,250 @@
...
+
+#### FriConfig
+The `FriConfig` struct is used to specify parameters for the FRI protocol. It contains the following fields:
+- **`stream: icicleStreamHandle`**: The CUDA stream for asynchronous execution. If `nullptr`, the default stream is used.
+- **`folding_factor: size_t`**: The factor by which the codeword is folded in each round.
+- **`stopping_degree: size_t`**: The minimal polynomial degree at which folding stops.
+- **`pow_bits: size_t`**: Number of leading zeros required for proof-of-work. If set, the optional proof-of-work phase is executed.
+- **`nof_queries: size_t`**: Number of queries computed for each folded layer of FRI.
+- **`are_inputs_on_device: bool`**: If true, the input polynomials are stored on the device (e.g., GPU); otherwise, they remain on the host (e.g., CPU).
+- **`is_async: bool`**: If true, it runs the hash asynchronously.
```
LeonHibnik, line 56: What are the defaults?

LeonHibnik (changes requested):
Very nice! Left couple of comments

docs/versioned_docs/version-3.5.0/icicle/primitives/fri.md
ShanieWinitz: The default values are written bellow

LeonHibnik (approved):

docs/docs/icicle/primitives/fri.md
```diff
@@ -0,0 +1,250 @@
...
+  size_t folding_factor = 2;
+  size_t stopping_degree = 0;
+  size_t pow_bits = 16;
+  size_t nof_queries = 100;
+  bool are_inputs_on_device = false;
+  bool is_async = false;
+  ConfigExtension* ext = nullptr;
+};
+```
+> **_NOTE:_** Currently, only a folding factor of 2 is supported.
```
jeremyfelder, line 73: Please use docusaurus admonitions for all NOTE, WARNING, INFO, etc:
https://docusaurus.io/docs/markdown-features/admonitions

```suggestion
:::note

Currently, only a folding factor of 2 is supported.

:::
```

Otherwise, its very difficult to see these in the rendered docs:

image

yshekel (approved):

PR #799, opened 2025-03-05
fix sumcheck transcript-config copy rather than move

yshekel:
This PR addresses a critical issue in the SumcheckTranscript API where move semantics are improperly implemented due to the use of const rvalue references. This misuse leads to unintended copying of objects instead of moving them, which contradicts the original design intentions and can lead to performance degradation, especially since some of the data structures involved (like labels and public data) are potentially large.

Issues Addressed:

1. Const Rvalue Reference Misuse: The API currently accepts and stores configurations as const rvalue references (`const SumcheckTranscriptConfig<S>&&`). This approach mistakenly assumes that objects can be moved efficiently, but due to their constness, they are actually copied. This undermines the efficiency of move semantics.
2. Faulty Move Operations in Tests: The associated tests incorrectly move an object twice. This behavior, under normal circumstances, should lead to undefined behavior but appears to work correctly due to the accidental copying. This masks potential bugs and can lead to false positives during testing.

cuda-backend-branch: yshekel/fix_sumcheck_transcript

### DIFF
--- a/icicle/include/icicle/sumcheck/sumcheck_transcript_config.h
+++ b/icicle/include/icicle/sumcheck/sumcheck_transcript_config.h
@@ -98,12 +98,12 @@ namespace icicle {
     const F& get_seed_rng() const { return m_seed_rng; }
 
   private:
-    Hash m_hasher;                                         ///< Hash function used for randomness generation.
-    const std::vector<std::byte> m_domain_separator_label; ///< Label for the domain separator in the transcript.
-    const std::vector<std::byte> m_round_poly_label;       ///< Label for round polynomials in the transcript.
-    const std::vector<std::byte> m_round_challenge_label;  ///< Label for round challenges in the transcript.
-    const bool m_little_endian = true;                     ///< Encoding endianness (default: little-endian).
-    F m_seed_rng;                                          ///< Seed for initializing the RNG.
+    Hash m_hasher;                                   ///< Hash function used for randomness generation.
+    std::vector<std::byte> m_domain_separator_label; ///< Label for the domain separator in the transcript.
+    std::vector<std::byte> m_round_poly_label;       ///< Label for round polynomials in the transcript.
+    std::vector<std::byte> m_round_challenge_label;  ///< Label for round challenges in the transcript.
+    const bool m_little_endian = true;               ///< Encoding endianness (default: little-endian).
+    F m_seed_rng;                                    ///< Seed for initializing the RNG.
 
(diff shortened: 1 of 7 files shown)


idanfr-ingo (approved):

icicle/include/icicle/sumcheck/sumcheck_transcript_config.h
```diff
@@ -82,7 +82,7 @@ namespace icicle {
 
     // Move Constructor
     SumcheckTranscriptConfig(SumcheckTranscriptConfig&& other) noexcept
-        : m_hasher(other.m_hasher), m_domain_separator_label(std::move(other.m_domain_separator_label)),
+        : m_hasher(std::move(other.m_hasher)), m_domain_separator_label(std::move(other.m_domain_separator_label)),
```
mickeyasa, line 85: do we have move for hasher?

icicle/include/icicle/sumcheck/sumcheck_transcript_config.h
```diff
@@ -98,12 +98,12 @@ namespace icicle {
...
 
   private:
-    Hash m_hasher;                                         ///< Hash function used for randomness generation.
-    const std::vector<std::byte> m_domain_separator_label; ///< Label for the domain separator in the transcript.
-    const std::vector<std::byte> m_round_poly_label;       ///< Label for round polynomials in the transcript.
-    const std::vector<std::byte> m_round_challenge_label;  ///< Label for round challenges in the transcript.
-    const bool m_little_endian = true;                     ///< Encoding endianness (default: little-endian).
-    F m_seed_rng;                                          ///< Seed for initializing the RNG.
+    Hash m_hasher;                                   ///< Hash function used for randomness generation.
+    std::vector<std::byte> m_domain_separator_label; ///< Label for the domain separator in the transcript.
```
mickeyasa, line 102: why not const?
doesn't it work with the move?

mickeyasa (approved):

icicle/include/icicle/sumcheck/sumcheck_transcript_config.h
yshekel: I think it's more correct for a move constructor. In this case copy is cheap but still.

icicle/include/icicle/sumcheck/sumcheck_transcript_config.h
yshekel: Yes it makes the move copy instead. Now I test it in one test

PR #800, opened 2025-03-06
Idan/add sumcheck tests

idanfr-ingo:
This PR add some tests for sumcheck, and change an undefined behaviour at transcript creation in the tests.
The additional tests should check few more cases:
- when not polys is the max allowed value (on device)
- when degree of combine polynomial is the max allowed value (on device)
- non-default transcript config
- validate that the separators are created properly

 cuda-backend-branch: idan/hotfix_sumcheck


icicle/tests/test_field_api.cpp
```diff
@@ -444,24 +681,27 @@ TEST_F(FieldTestBase, SumcheckCudaShouldFailCases)
 
     ASSERT_EQ(error, eIcicleError::INVALID_ARGUMENT);
   };
+  for (const auto& device : s_registered_devices) {
+    if (device == "CPU") continue;
```
yshekel, line 685: So every backend has different limitations? Seems like there is some minimum and then some backends can support more? Is it documented somewhere? It means CPU code can work but then moving to GPU wouldn't. Maybe it's not a good idea even it it works. @mickeyasa @idanfr-ingo

icicle/tests/test_field_api.cpp
```diff
@@ -361,7 +362,243 @@ TEST_F(FieldTestBase, SumcheckUserDefinedCombine)
...
+  for (const auto& device : s_registered_devices) {
+    run(device, mle_polynomials, mle_poly_size, claimed_sum, "Sumcheck");
+  }
+
+  for (auto& mle_poly_ptr : mle_polynomials) {
+    delete[] mle_poly_ptr;
+  }
+}
+
+MlePoly max_allowed_nof_polys_comine(const std::vector<MlePoly>& inputs)
```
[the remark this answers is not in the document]
idanfr-ingo: (1) They are defined at the backends
(2) Most error massages are clear and state the problem explicitly. This specific error is related to internals of program, and not really likely to happen to users.

icicle/tests/test_field_api.cpp
idanfr-ingo: Currently there are no limitations on the CPU backend. Only the other devices have limitations, which are identicall right now.

You suggest to add the same limitations to the CPU backend to make all backends equall?

mickeyasa: As we discussed f2f, you can move the limitation checks to the frontend

mickeyasa (changes requested):

icicle/tests/test_field_api.cpp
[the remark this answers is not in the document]
yshekel: I mean where in the documentation

mickeyasa (approved):
V

PR #813, opened 2025-03-18
FRI rust wrappers

nonam3e:
This PR adds FRI rust wrappers into icicle

cuda-backend-branch: main


wrappers/rust/icicle-curves/icicle-bw6-761/src/lib.rs
jeremyfelder, line 1: Why aren't we adding support for FRI here? is it supported via bls12377?

icicle/include/icicle/fri/fri_proof.h
```diff
@@ -91,6 +90,19 @@ namespace icicle {
...
     }
 
+    /**
+     * @brief Get a pointer to the Merkle proofs at a specific query index.
+     *
+     * @param query_idx The index of the query to retrieve proofs for.
+     * @return Pointer to the first MerkleProof at the given query index.
+     */
+    MerkleProof* get_proofs_at_query(const size_t query_idx) { return m_query_proofs[query_idx].data(); }
+    /**
```
jeremyfelder, line 100: Add new line between function and next function's dev doc

icicle/src/fri/fri_c_api.cpp
```diff
@@ -0,0 +1,295 @@
...
+  if (!proof_ptr) {
+    ICICLE_LOG_ERROR << "CANNOT USE null proof_ptr";
+    return eIcicleError::INVALID_POINTER;
+  }
+  nof_queries = proof_ptr->get_nof_fri_queries();
+  nof_rounds = proof_ptr->get_nof_fri_rounds();
+  return eIcicleError::SUCCESS;
+}
+
+eIcicleError CONCAT_EXPAND(ICICLE_FFI_PREFIX, fri_proof_get_round_proof_at)(
```
jeremyfelder, line 99: The function name here is confusing. Sounds like a single round proof should be returned based on a round index but we are getting a pointer to all of round proofs at a specific query index
```suggestion
eIcicleError CONCAT_EXPAND(ICICLE_FFI_PREFIX, fri_proof_get_round_proofs_at)(
```

wrappers/rust/icicle-core/src/fri/fri_proof.rs
```diff
@@ -0,0 +1,143 @@
...
+                    }
+                    Ok(proofs)
+                }
+            }
+
+            fn get_final_poly(&self) -> Result<Vec<$field>, eIcicleError> {
+                let mut nof: usize = 0;
+                unsafe {
+                    fri_proof_get_final_poly_size(self.handle, &mut nof).wrap()?;
+                    println!("nof {}", nof);
```
jeremyfelder, line 112: Remove
```suggestion
```

icicle/include/icicle/fri/fri_proof.h
```diff
@@ -91,6 +90,19 @@ namespace icicle {
                      // 0 of the current round
     }
 
+    /**
+     * @brief Get a pointer to the Merkle proofs at a specific query index.
+     *
+     * @param query_idx The index of the query to retrieve proofs for.
+     * @return Pointer to the first MerkleProof at the given query index.
+     */
+    MerkleProof* get_proofs_at_query(const size_t query_idx) { return m_query_proofs[query_idx].data(); }
```
yshekel, line 99: I think it's better to return the 'const vector<MerkleProof>&' rather than a pointer like that. This is riskier.

ShanieWinitz, line 99: Can we use "get_query_proof_slot" instead of adding this method?

icicle/include/icicle/utils/rand_gen.h
```diff
@@ -27,4 +27,19 @@ static size_t rand_size_t(size_t min = 0, size_t max = SIZE_MAX)
...
   return dist(rand_generator);
+}
+
+/**
+ * @brief Generate random unsigned integer in range (inclusive)
+ * @param min Lower limit.
+ * @param max Upper limit.
+ * @return Random (uniform distribution) unsigned integer s.t. min <= integer <= max.
+ */
+static std::vector<size_t> rand_size_t_vector(size_t nof_queries, size_t min = 0, size_t max = SIZE_MAX)
```
yshekel, line 38: (1) I would make it template rather than hardcode size_t
(2) nof_queries should be more general. probably 'size'

wrappers/rust/icicle-core/src/fri/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+use fri_transcript_config::FriTranscriptConfig;
+use icicle_runtime::{config::ConfigExtension, eIcicleError, memory::HostOrDeviceSlice, IcicleStreamHandle};
+
+pub type FriProof<F> = <<F as FieldImpl>::Config as FriImpl<F>>::FriProof;
+
+/// Computes the FRI proof using the given configuration and input data.
+/// # Returns
+/// - `Ok(())` if the FRI proof was successfully computed.
+/// - `Err(eIcicleError)` if an error occurred during proof generation.
+pub fn fri<F: FieldImpl>(
```
jeremyfelder, line 16:
```suggestion
pub fn prove<F: FieldImpl>(
```

icicle/src/fri/fri_c_api.cpp
```diff
@@ -0,0 +1,295 @@
...
+  if (!proof_ptr) return eIcicleError::INVALID_POINTER;
+  delete proof_ptr;
+  return eIcicleError::SUCCESS;
+}
+
+eIcicleError CONCAT_EXPAND(ICICLE_FFI_PREFIX, fri_proof_get_proof_sizes)(
+  FriProofHandle proof_ptr, size_t& nof_queries, size_t& nof_rounds)
+{
+  if (!proof_ptr) {
+    ICICLE_LOG_ERROR << "CANNOT USE null proof_ptr";
```
yshekel, line 91: why this upper case like this?

icicle/src/fri/fri_c_api.cpp
```diff
@@ -0,0 +1,295 @@
...
+    return eIcicleError::INVALID_POINTER;
+  }
+  if (query_idx >= proof_ptr->get_nof_fri_queries()) {
+    ICICLE_LOG_ERROR << "query_idx IS OUT OF RANGE";
+    return eIcicleError::INVALID_ARGUMENT;
+  }
+  *proofs = proof_ptr->get_proofs_at_query(query_idx);
```
yshekel, line 110: I now see why you did it but still you can do the '.data()' here to avoid the merkle proof having a problematic api

icicle/src/fri/fri_c_api.cpp
```diff
@@ -0,0 +1,295 @@
...
+{
+  if (!proof_ptr) {
+    ICICLE_LOG_ERROR << "CANNOT USE null proof_ptr";
+    return eIcicleError::INVALID_POINTER;
+  }
+  result = proof_ptr->get_pow_nonce();
+  return eIcicleError::SUCCESS;
+}
+
+eIcicleError CONCAT_EXPAND(ICICLE_FFI_PREFIX, get_fri_proof_mt)(
```
yshekel, line 144: rename to fri_merkle_tree_prove to align with the latest api. Same for verify

icicle/src/fri/fri_c_api.cpp
```diff
@@ -0,0 +1,295 @@
...
+  const FriConfig& fri_config,
```
yshekel, line 145: Don't use reference in C api. Use only pointers. It's better and we moved to use only that in ICICLE.

icicle/src/fri/fri_c_api.cpp
```diff
@@ -0,0 +1,295 @@
...
+    ICICLE_LOG_ERROR << "Invalid FFI Fri transcript configuration.";
+    return eIcicleError::INVALID_POINTER;
+  }
+  auto fri_transcript_config = convert_ffi_transcript_config(ffi_transcript_config);
+  return prove_fri_merkle_tree<scalar_t>(
+    fri_config, fri_transcript_config, input_data, input_size, merkle_tree_leaves_hash, merkle_tree_compress_hash,
+    output_store_min_layer, fri_proof);
+}
+
+eIcicleError CONCAT_EXPAND(ICICLE_FFI_PREFIX, verify_fri_mt)(
```
yshekel, line 164: (1) rename to fri_merkle_tree_verify
(2) same comment about the references here. Also should it be HashHandle? I don't think the ffi has such an object to give

wrappers/rust/icicle-core/src/fri/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+                            input_data.len() as u64,
+                            merkle_tree_leaves_hash.handle,
+                            merkle_tree_compress_hash.handle,
+                            output_store_min_layer,
+                            fri_proof.handle,
+                        )
+                        .wrap()
+                    }
+                }
+                fn verify_fri_mt(
```
jeremyfelder, line 187:
```suggestion

                fn verify_fri_mt(
```

wrappers/rust/icicle-core/src/fri/mod.rs
ShanieWinitz, line 16: In C++, we decided to name it prove_fri_merkle_tree to ensure clarity and to allow for a future version of FRI that uses MMCS instead of Merkle trees as the commitment scheme. We might consider doing the same in the Rust wrapper.

wrappers/rust/icicle-core/src/fri/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+                    test_utilities::test_load_and_init_devices();
+                    // init domain for both devices
+                    test_utilities::test_set_ref_device();
+                    init_domain::<$ntt_field>(MAX_SIZE, FAST_TWIDDLES_MODE);
+
+                    test_utilities::test_set_main_device();
+                    init_domain::<$ntt_field>(MAX_SIZE, FAST_TWIDDLES_MODE);
+                });
+                test_utilities::test_set_main_device();
+            }
```
jeremyfelder, line 248: Does this need a `--skip phase` in the CI due to initializing domains?

wrappers/rust/icicle-core/src/fri/fri_proof.rs
```diff
@@ -0,0 +1,143 @@
...
+    F: FieldImpl,
+{
+    /// Constructs a new instance of the FRI proof.
+    fn new() -> Self;
+
+    /// Returns the query proofs.
+    ///
+    /// Note: `ManuallyDrop` is used because the values are borrowed from the `FriProof`
+    /// and will be dropped along with it.
+    fn get_query_proofs(&self) -> Result<Vec<Vec<ManuallyDrop<MerkleProof>>>, eIcicleError>;
```
yshekel, line 21: No need for get_nof_queries()? final_poly_size()? things like that? I mean for users

wrappers/rust/icicle-core/src/fri/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+        fri_proof,
+    )
+}
+
+/// Verifies a FRI proof using the given configuration and Merkle tree hashes.
+/// # Returns
+/// - `Ok(true)` if the proof is valid.
+/// - `Ok(false)` if the proof is invalid.
+/// - `Err(eIcicleError)` if verification failed due to an error.
+pub fn verify_fri<F: FieldImpl>(
```
ShanieWinitz, line 44: Same here, "verify_fri_merkle_tree" in C++.

wrappers/rust/icicle-core/src/fri/fri_proof.rs
```diff
@@ -0,0 +1,143 @@
...
+            fn fri_proof_get_pow_nonce(handle: FriProofHandle, result: *mut u64) -> eIcicleError;
+
+            #[link_name = concat!($field_prefix, "_fri_proof_get_final_poly_size")]
+            fn fri_proof_get_final_poly_size(handle: FriProofHandle, result: *mut usize) -> eIcicleError;
+
+            #[link_name = concat!($field_prefix, "_fri_proof_get_final_poly")]
+            fn fri_proof_get_final_poly(handle: FriProofHandle, final_poly: *mut *const $field) -> eIcicleError;
+
+            #[link_name = concat!($field_prefix, "_fri_proof_get_proof_sizes")]
+            fn fri_proof_get_proof_sizes(
```
yshekel, line 63: I think proof_sizes is a bad name. Please rename or maybe split to two functions (one for queries, one for rounds)

wrappers/rust/icicle-core/src/fri/fri_proof.rs
```diff
@@ -0,0 +1,143 @@
...
+                handle: FriProofHandle,
+                nof_queries: *mut usize,
+                nof_rounds: *mut usize,
+            ) -> eIcicleError;
+
+            #[link_name = concat!($field_prefix, "_fri_proof_get_round_proof_at")]
+            fn fri_proof_get_round_proof_at(
```
yshekel, line 70: maybe rename to fri_proof_get_proofs_for_query()?

wrappers/rust/icicle-core/src/fri/fri_proof.rs
```diff
@@ -0,0 +1,143 @@
...
+            fn get_query_proofs(&self) -> Result<Vec<Vec<ManuallyDrop<MerkleProof>>>, eIcicleError> {
+                let mut nof_queries: usize = 0;
+                let mut nof_rounds: usize = 0;
+                unsafe {
+                    fri_proof_get_proof_sizes(self.handle, &mut nof_queries, &mut nof_rounds).wrap()?;
+                    let mut proofs: Vec<Vec<ManuallyDrop<MerkleProof>>> = Vec::with_capacity(nof_queries as usize);
+                    for i in 0..nof_queries {
+                        let mut proof: MerkleProofHandle = std::ptr::null();
+                        fri_proof_get_round_proof_at(self.handle, i, &mut proof).wrap()?;
+                        let round_slice = slice::from_raw_parts(proof, nof_rounds as usize);
```
yshekel, line 96: please try to make your code easier to read using clear names. For example here 'round_slice' is all the proofs for a single query right? so if this is correct I find 'round_slice' very confusing.

wrappers/rust/icicle-core/src/fri/fri_proof.rs
```diff
@@ -0,0 +1,143 @@
...
+                        proofs.push(
+                            round_slice
+                                .iter()
+                                .map(|x| ManuallyDrop::new(MerkleProof::from_handle(x)))
+                                .collect(),
```
yshekel, line 101: Here you basically share the same underlying pointer and this ManuallyDrop avoids the drop() being called when the object is dropped?

wrappers/rust/icicle-core/src/fri/fri_proof.rs
yshekel, line 112: remove

wrappers/rust/icicle-core/src/fri/fri_proof.rs
jeremyfelder, line 1: We will need a way for a verifier to create a FriProof from data that they are deserializing. Right now I don't see a way for a verifier to set a nonce or initialize the FriProof object in order to update the final_poly and merkle_proofs

jeremyfelder (changes requested):

wrappers/rust/icicle-core/src/fri/fri_proof.rs
```diff
@@ -0,0 +1,143 @@
...
+                    Ok(proofs)
+                }
+            }
+
+            fn get_final_poly(&self) -> Result<Vec<$field>, eIcicleError> {
+                let mut nof: usize = 0;
+                unsafe {
+                    fri_proof_get_final_poly_size(self.handle, &mut nof).wrap()?;
+                    println!("nof {}", nof);
+                    let mut ptr: *const $field = std::ptr::null();
```
yshekel, line 113: here too 'ptr' is not clear so maybe 'finaly_poly_ptr'

icicle/src/fri/fri_c_api.cpp
ShanieWinitz, line 99: This function returns the Merkle proofs for all rounds of a given query, right? the name might be a bit confusing

wrappers/rust/icicle-core/src/fri/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+/// - `Err(eIcicleError)` if an error occurred during proof generation.
+pub fn fri<F: FieldImpl>(
+    config: &FriConfig,
+    fri_transcript_config: &FriTranscriptConfig<F>,
+    input_data: &(impl HostOrDeviceSlice<F> + ?Sized),
+    merkle_tree_leaves_hash: &Hasher,
+    merkle_tree_compress_hash: &Hasher,
+    output_store_min_layer: u64,
+    fri_proof: &mut FriProof<F>,
+) -> Result<(), eIcicleError>
```
yshekel, line 24: I think better return Result<FriProof, eIcicleError> to be consistent with other ICICLE APIs. in C++ we don't do that because it's less convenient but in rust I think it's better

wrappers/rust/icicle-core/src/fri/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+            is_async: false,
+            ext: ConfigExtension::new(),
+        }
+    }
+}
+
+pub trait FriImpl<F: FieldImpl> {
+    type FieldConfig: FieldConfig + GenerateRandom<F> + FieldArithmetic<F>;
+    type FriProof: FriProofTrait<F>;
+    fn get_fri_proof_mt(
```
yshekel, line 97: rename here too to FriMerkleTreeImpl and then prove/verify. I think it's cleaner. But in any case please be consistent (meaning verify_fri_mt should go with prove_fro_mt).

wrappers/rust/icicle-core/src/fri/tests.rs
```diff
@@ -0,0 +1,53 @@
...
+use crate::fri::fri_proof::FriProofTrait;
+use crate::{
+    fri::{fri, fri_transcript_config::FriTranscriptConfig, verify_fri, FriConfig, FriProof},
+    hash::Hasher,
+    traits::{FieldImpl, GenerateRandom},
+};
+use icicle_runtime::eIcicleError;
+use icicle_runtime::{memory::HostSlice, test_utilities};
+
+pub fn check_fri<F: FieldImpl>(hash_new: &dyn Fn(u64) -> Result<Hasher, eIcicleError>)
```
yshekel, line 11: I think passing the hasher here is simpler but your call

wrappers/rust/icicle-core/src/fri/tests.rs
ShanieWinitz, line 1: In C++, I added tests for non-default configurations, and tests for cases where FRI should return an error for invalid arguments. Should we do the same in Rust?

jeremyfelder: I would prefer to keep Rust tests to a minimum where we can ensure the FFI is working correctly. If we have specific checks in Rust or error handling across the FFI that needs checking that should be tested as well but I don't think we should replicate all tests in each language

wrappers/rust/icicle-core/src/fri/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+    input_data: &(impl HostOrDeviceSlice<F> + ?Sized),
+    merkle_tree_leaves_hash: &Hasher,
+    merkle_tree_compress_hash: &Hasher,
+    output_store_min_layer: u64,
+    fri_proof: &mut FriProof<F>,
+) -> Result<(), eIcicleError>
+where
+    <F as FieldImpl>::Config: FriImpl<F>,
+{
+    <F::Config as FriImpl<F>>::get_fri_proof_mt(
```
yshekel, line 28: I think that you don't handle here DeviceVec and also don't check params

wrappers/rust/icicle-core/src/fri/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+                    test_utilities::test_set_ref_device();
+                    init_domain::<$ntt_field>(MAX_SIZE, FAST_TWIDDLES_MODE);
+
+                    test_utilities::test_set_main_device();
+                    init_domain::<$ntt_field>(MAX_SIZE, FAST_TWIDDLES_MODE);
+                });
+                test_utilities::test_set_main_device();
+            }
+
+            #[test]
```
yshekel, line 250: please add test with DeviceVec. I don't think you implemented this

wrappers/rust/icicle-core/src/merkle/mod.rs
```diff
@@ -131,6 +132,10 @@ impl MerkleProof {
             }
         }
     }
+
+    pub fn from_handle(handle: MerkleProofHandle) -> Self {
```
yshekel, line 136: I wonder if this is not risky in terms of sharing the handle. I mean it can be dropped and you won't know. Maybe we need to return a ManuallyDrop<Self> if possible here. What do you think? @jeremyfelder

yshekel (changes requested):
Can you please make another PR for the formatting only? It's harder to review like that.

wrappers/rust/icicle-core/src/merkle/mod.rs
nonam3e: I'm not sure that the result will be `ManuallyDrop` everytime, but yeah, some problems may occur here, maybe add unsafe here?

wrappers/rust/icicle-core/src/fri/fri_proof.rs
nonam3e: yes, I didn't found the other way to do that without copying the whole Matrix of Fri proofs

wrappers/rust/icicle-curves/icicle-bw6-761/src/lib.rs
nonam3e: yes, the same way it's done for `ntt`

icicle/src/fri/fri_c_api.cpp
nonam3e: changed to `fri_proof_get_round_proofs_for_query`

wrappers/rust/icicle-core/src/fri/fri_proof.rs
nonam3e: to_vec() copies values here, am I right?

wrappers/rust/icicle-core/src/fri/mod.rs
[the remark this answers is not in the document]
nonam3e: renamed to `prove()` and `verify()`

wrappers/rust/icicle-core/src/fri/mod.rs
nonam3e: Ok, just thought there could be a problem if the config is async

icicle/include/icicle/fri/fri_proof.h
nonam3e: Here I return the array of round proofs, but `get_query_proof_slot` returns the element

wrappers/rust/icicle-core/src/fri/fri_proof.rs
nonam3e: User can get it from the fri config if it's needed

wrappers/rust/icicle-core/src/fri/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+    let mut local_cfg = config.clone();
+    local_cfg.are_inputs_on_device = true;
+    let mut fri_proof = FriProof::<F>::new();
+    <F::Config as FriMerkleTreeImpl<F>>::fri_merkle_tree_prove(
+        config,
+        fri_transcript_config,
+        input_data,
+        merkle_tree_leaves_hash,
+        merkle_tree_compress_hash,
+        output_store_min_layer,
```
yshekel, line 39: I think this is very unclear for users. What does 'output_store_min_layer' mean?
Maybe rename to 'merkle_tree_min_layer_to_store'. I think it should be in the config struct honestly (in C++ too). Most users don't want to care about it, and so they get the default value from the config.

wrappers/rust/icicle-core/src/fri/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+/// - `Ok(true)` if the proof is valid.
+/// - `Ok(false)` if the proof is invalid.
+/// - `Err(eIcicleError)` if verification failed due to an error.
+pub fn fri_merkle_tree_verify<F: FieldImpl>(
+    config: &FriConfig,
+    fri_transcript_config: &FriTranscriptConfig<F>,
+    fri_proof: &FriProof<F>,
+    merkle_tree_leaves_hash: &Hasher,
+    merkle_tree_compress_hash: &Hasher,
+    output_store_min_layer: u64,
```
yshekel, line 56: you don't need this 'output_store_min_layer' here since you don't build a merkle tree in the verifier. Remove it. If FFI needs it then it probably is wrong too or just pass 0.

wrappers/rust/icicle-core/src/fri/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+    type FriProof: FriProofTrait<F>;
+    fn fri_merkle_tree_prove(
+        config: &FriConfig,
+        fri_transcript_config: &FriTranscriptConfig<F>,
+        input_data: &(impl HostOrDeviceSlice<F> + ?Sized),
+        merkle_tree_leaves_hash: &Hasher,
+        merkle_tree_compress_hash: &Hasher,
+        output_store_min_layer: u64,
+        fri_proof: &mut Self::FriProof,
+    ) -> Result<(), eIcicleError>;
```
yshekel, line 111: Why not return Result<FriProof, eIcicleError>? I think it's cleaner and also consistent with the verifier which is Result<bool, eIcicleError>

wrappers/rust/icicle-core/src/fri/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+            pow_bits: 16,
+            nof_queries: 100,
+            are_inputs_on_device: false,
+            is_async: false,
+            ext: ConfigExtension::new(),
+        }
+    }
+}
+
+pub trait FriMerkleTreeImpl<F: FieldImpl> {
```
yshekel, line 100: I suggest removing ths 'Impl' postfix. I don't think that we name traits like that in icicle

wrappers/rust/icicle-core/src/fri/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+    };
+}
+
+#[macro_export]
+macro_rules! impl_fri_tests {
+    (
+      $field_prefix_ident:ident,
+      $ntt_field:ident,
+      $field:ident,
+      $hasher_new:expr
```
yshekel, line 228: Can you remove this from here and have it in the tests below instead? I mean so that when instantiating it the tests, we don't need to have such a function in each crate.

wrappers/rust/icicle-core/src/fri/mod.rs
nonam3e: fair, I changed it only `pub fn fri_merkle_tree_prove<F: FieldImpl>` before, now changed it in the trait as well

wrappers/rust/icicle-core/src/fri/tests.rs
```diff
@@ -0,0 +1,116 @@
...
+            &transcript_config,
+            &fri_proof,
+            &merkle_tree_leaves_hash,
+            &merkle_tree_compress_hash,
+            output_store_min_layer,
+        )
+        .unwrap();
+        assert!(valid);
+    };
+    test_utilities::test_set_main_device();
```
yshekel, line 49: I guess that is the reason you are unable to pass a hasher as input right?

wrappers/rust/icicle-curves/icicle-bls12-381/src/fri/mod.rs
```diff
@@ -0,0 +1,15 @@
...
+impl_fri!("bls12_381", bls12_381_fri, ScalarField, ScalarCfg);
+
+#[cfg(test)]
+mod tests {
+    use icicle_core::impl_fri_tests;
+    use icicle_hash::keccak::Keccak256;
+
+    use crate::curve::ScalarField;
+
+    impl_fri_tests!(bls12_381_fri_test, ScalarField, ScalarField, Keccak256::new);
```
yshekel, line 14: I commented on this but I think this is a mistake. Why would this scope choose the hash function for the tests? it should be the tests repsonsibility.

wrappers/rust/icicle-core/src/merkle/mod.rs
```diff
@@ -131,6 +135,16 @@ impl MerkleProof {
             }
         }
     }
+
+    pub unsafe fn from_handle(handle: MerkleProofHandle) -> Self {
```
yshekel, line 139: I think this is a risky method and at least deserves a good comment. Can we avoid it?

wrappers/rust/icicle-curves/icicle-bls12-381/src/fri/mod.rs
nonam3e: it's a bit tricky, tests are located in `icicle-core` and specific hashes are located in `icicle-hashes`, so I don't know if there's a way to pass a hasher in the other way. Here I can pass a `Hasher` instance ofc, but I'm not sure we can make it a test responsibility

wrappers/rust/icicle-core/src/fri/fri_proof.rs
```diff
@@ -0,0 +1,207 @@
+use std::{ffi::c_void, mem::ManuallyDrop};
+
+use icicle_runtime::eIcicleError;
+
+use crate::{merkle::MerkleProof, traits::{FieldImpl, Handle}};
+
+pub struct PointerArray {
```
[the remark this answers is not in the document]
nonam3e: oh forgot to delete it, sry

wrappers/rust/icicle-core/src/merkle/mod.rs
nonam3e: I don't see any other way to get a MerkleTree instance without using icicle_merkle_proof_create, but we can't use that for FRI. We could copy the entire matrix of proofs to avoid using from_handle, if it's worth the trade-off

wrappers/rust/icicle-core/src/fri/tests.rs
```diff
@@ -0,0 +1,116 @@
...
+            .get_query_proofs()
+            .unwrap();
+        let final_poly = fri_proof
+            .get_final_poly()
+            .unwrap();
+        let pow_nonce = fri_proof
+            .get_pow_nonce()
+            .unwrap();
+
+        let fri_proof_copy = FriProof::<F>::create_with_arguments(query_proofs, final_poly, pow_nonce);
```
yshekel, line 99: Is this  testing serialization/deserialization?

wrappers/rust/icicle-core/src/fri/fri_proof.rs
```diff
@@ -0,0 +1,207 @@
...
+    fn get_query_proofs(&self) -> Result<Vec<Vec<ManuallyDrop<MerkleProof>>>, eIcicleError>;
+
+    /// Returns the final polynomial values.
+    fn get_final_poly(&self) -> Result<Vec<F>, eIcicleError>;
+
+    /// Returns the proof-of-work nonce.
+    fn get_pow_nonce(&self) -> Result<u64, eIcicleError>;
+
+    /// Creates a new instance of `FriProof` with the given proof data and wraps it in `ManuallyDrop`.
+    fn create_with_arguments(query_proofs: Vec<Vec<ManuallyDrop<MerkleProof>>>, final_poly: Vec<F>, pow_nonce: u64) -> ManuallyDrop<Self>;
```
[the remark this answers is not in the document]
nonam3e: I thought it could be used to create an FriProof from a proof obtained outside of Icicle, for verification after, do you think it's not needed?

wrappers/rust/icicle-core/src/fri/tests.rs
nonam3e: yes

icicle/src/fri/fri_c_api.cpp
```diff
@@ -0,0 +1,475 @@
...
+
+extern "C" {
+
+// FriProof API
+
+typedef FriProof<scalar_t>* FriProofHandle;
+
+FriProofHandle CONCAT_EXPAND(ICICLE_FFI_PREFIX, icicle_initialize_fri_proof)() { return new FriProof<scalar_t>(); }
+
+FriProofHandle CONCAT_EXPAND(ICICLE_FFI_PREFIX, icicle_create_with_arguments_fri_proof)(MerkleProof*** query_proofs_ffi, size_t nof_queries, size_t nof_rounds, scalar_t* final_poly_ffi, size_t final_poly_size, uint64_t pow_nonce) {
```
yshekel, line 81: Can you explain the scenario where we use this? The issue I see is that you have rust vectors for for proofs, the final poly etc. and here you copy them three times !! (one here from query_proofs_ffi to query_proofs, second time you copy into FriProof constructor and third time inside the FriProof constructor). Then later if you call proof.get_queries() you copy again from C++ vector to rust vector!!

I really really think we should have a design doc here. First it's not really clear to me why you need this method but then look how inefficient it is.
Maybe in that case it would be better to have to rust friProof not based on a C++ handle object since anyway you copy C++ vectors to rust vectors at least once. Then doing so makes this 'icicle_create_with_arguments_fri_proof' method uneccessary and actually zero copy to create such a merkle proof in rust.

wrappers/rust/icicle-core/src/fri/fri_proof.rs
```diff
@@ -0,0 +1,207 @@
...
+                }
+            }
+
+            fn get_final_poly(&self) -> Result<Vec<$field>, eIcicleError> {
+                let mut nof: usize = 0;
+                unsafe {
+                    fri_proof_get_final_poly_size(self.handle, &mut nof).wrap()?;
+                    let mut final_poly_ptr: *const $field = std::ptr::null();
+                    fri_proof_get_final_poly(self.handle, &mut final_poly_ptr).wrap()?;
+                    Ok(slice::from_raw_parts(final_poly_ptr, nof as usize).to_vec())
```
yshekel, line 178: Note that here you copy from C++ to rust. I commented earlier about the rust to C++ copy (The create_with_params()) so if you anyway copy from C++ to Rust, why have the friProof wrap the handle anyway? Just copy once when proving and that's it. Then you don't need much FFI around the proof and the create_with_params() can be zero copy if moving instead of copying

yshekel (changes requested):
I believe the absence of a design document led us to generate code that isn’t as effective as it could be. If we had started with a design and involved the team in a review early on, we might have ended up with a better solution.

wrappers/rust/icicle-core/src/fri/fri_proof.rs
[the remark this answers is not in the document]
yshekel: I actually think it can be very useful to serialize/deserialize proofs. I just think this way is not efficient. Also you assume they can generate MerkleProof which is not that trivial either.

wrappers/rust/icicle-core/src/fri/mod.rs
yshekel: Honestly I am not sure either. I just see we are not consistent. In that case you can choose yes

wrappers/rust/icicle-curves/icicle-bls12-381/src/fri/mod.rs
yshekel: see example of sumcheck here https://github.com/ingonyama-zk/icicle/blob/baa0c88b1980f20991895f8b45bd67cdca58a775/wrappers/rust/icicle-core/src/sumcheck/mod.rs#L455

wrappers/rust/icicle-core/src/fri/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+    ) => {
+        mod $field_prefix_ident {
```
jeremyfelder, line 230: Why is there another module here within the tests mod?

wrappers/rust/icicle-core/src/fri/fri_proof.rs
jeremyfelder: > why have the friProof wrap the handle anyway?

While I don't see a case where the `FriProof` would be used by the prover aside from serializing it, the current design allows a prover/user to perform the copies when they want to and not have it be forced at the time of creating the proof

wrappers/rust/icicle-core/src/fri/mod.rs
nonam3e: We need to add tests for both the base field and the extension field. Without a mod here, they will overlap

wrappers/rust/icicle-core/src/merkle/mod.rs
yshekel: (1) Since MerkleProof is simply a struct with a handle, maybe we could simply have Fri logic instantiate the struct directly and remove this from here?
(2) Regardless of (1), we must make sure to not drop twice but I guess if you only do it in the ffi core logic of fri, then it should be fine.

wrappers/rust/icicle-core/src/fri/fri_proof.rs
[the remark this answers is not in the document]
yshekel: UPDATE: you explained that it's only the MerkleProof pointers being copied and the final_poly is assumed to be very small so maybe copy here is not an issue actually.

Regardless, if users cannot serialize proofs to file and read it back on another machine, then I am not sure how they can even create this 'Vec<Vec<ManuallyDrop<MerkleProof>>>'

wrappers/rust/icicle-core/src/merkle/mod.rs
jeremyfelder: 1) This would require making the handle attribute public here which would be worse in my opinion.
2) The ManualDrop in FRI covers this no?

wrappers/rust/icicle-core/src/fri/fri_proof.rs
```diff
@@ -0,0 +1,278 @@
...
+
+pub trait FriProofTrait<F: FieldImpl>: Sized + Handle + Serialize + DeserializeOwned
+where
+    F: FieldImpl,
+{
+    /// Constructs a new instance of the `FriProof`.
+    fn new() -> Self;
+
+    /// Returns the query proofs.
+    fn get_query_proofs<T: Clone>(&self) -> Result<Vec<Vec<MerkleProofData<T>>>, eIcicleError>;
```
yshekel, line 17: Can it be that T!=F?

wrappers/rust/icicle-core/src/fri/fri_proof.rs
```diff
@@ -0,0 +1,278 @@
...
+
+        pub struct FriProof {
+            handle: FriProofHandle,
+        }
+
+        impl FriProofTrait<$field> for FriProof {
+            fn new() -> Self {
+                let handle: FriProofHandle = unsafe { icicle_initialize_fri_proof() };
+                if handle.is_null() {
+                    panic!("Couldn't create FriProof");
```
yshekel, line 107: can you return a Result here if it can fail?

wrappers/rust/icicle-core/src/fri/fri_proof.rs
yshekel, line 17: please add comment and describe the structure. I guess it's [query_idx][round_idx] but it's not clear

wrappers/rust/icicle-core/src/fri/fri_proof.rs
```diff
@@ -0,0 +1,278 @@
+use icicle_runtime::eIcicleError;
+
+use crate::{
+    merkle::MerkleProofData,
+    traits::{FieldImpl, Handle},
+};
+use serde::{de::DeserializeOwned, Serialize};
+
+pub trait FriProofTrait<F: FieldImpl>: Sized + Handle + Serialize + DeserializeOwned
```
yshekel, line 9: Clarifying question: In FRI, users can invoke the prover, obtain the proof, serialize it, and later deserialize it for verification — all without any copying between Rust and C++. Is that correct?
Alternatively, users can use the FriProofTrait to extract proof elements into Rust vectors from C++, or construct a proof in Rust using native Rust types. In this case, the data is copied into C++ vectors, allowing the proof to be verified using the Icicle verifier. Is that understanding accurate?

wrappers/rust/icicle-core/src/fri/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+    fri_transcript_config: &FriTranscriptConfig<F>,
+    input_data: &(impl HostOrDeviceSlice<F> + ?Sized),
+    merkle_tree_leaves_hash: &Hasher,
+    merkle_tree_compress_hash: &Hasher,
+    merkle_tree_min_layer_to_store: u64,
+) -> Result<FriProof<F>, eIcicleError>
+where
+    <F as FieldImpl>::Config: FriMerkleTree<F>,
+{
+    if input_data.is_on_device() && !input_data.is_on_active_device() {
```
yshekel, line 27: I think you'd better move the check logic to the trait so that if using the trait directly it works correctly as well. We have this bug in icicle in other places but let's avoid it for new code.

wrappers/rust/icicle-core/src/fri/tests.rs
```diff
@@ -0,0 +1,183 @@
...
+
+        let valid = fri_merkle_tree_verify(
+            &fri_config,
+            &transcript_config,
+            &fri_proof_copy,
+            &merkle_tree_leaves_hash,
+            &merkle_tree_compress_hash,
+        )
+        .unwrap();
+        stream
```
yshekel, line 110: minor but you need to release the stream here or it leaks

wrappers/rust/icicle-core/src/fri/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+                    &merkle_tree_leaves_hash,
+                    &merkle_tree_compress_hash,
+                    &transcript_hash,
+                    |fri_proof| serde_json::to_string(fri_proof).unwrap(),
+                    |s| serde_json::from_str(&s).unwrap(),
+                );
+            }
+        }
+    };
+}
```
yshekel, line 296: I think no tests is testing the trait FriProofTrait

wrappers/rust/icicle-core/src/fri/tests.rs
nonam3e: probably that's why tests failed (I hope)

wrappers/rust/icicle-core/src/fri/fri_proof.rs
nonam3e: everything is right, serialization allows user not to thing about contained structure to prove -> serialize -> deserialize -> verify, and exposing allows user to get the proofs data and use it outside icicle

wrappers/rust/icicle-core/src/fri/tests.rs
yshekel: I don't think it should make the tests fail but you should see a warning I think when dropping the rust object

nonam3e:
> Very nice PR, looks good. Left a few comments. What about Goldilocks field?

thanks for the mansion, will add it too, the pr was just opened before we supported this field

LeonHibnik (approved):

ChickenLover (approved):

LeonHibnik (approved):

jeremyfelder (approved):

yshekel (approved):

PR #834, opened 2025-04-02
norm

emirsoyturk:
## Describe the changes

This PR...

## Linked Issues

Resolves #

### DIFF
--- a/icicle/backend/cpu/src/ring/cpu_norm.cpp
+++ b/icicle/backend/cpu/src/ring/cpu_norm.cpp
@@ -1,0 +1,38 @@
+#include "icicle/norm.h"
+#include "icicle/backend/vec_ops_backend.h"
+#include "taskflow/taskflow.hpp"
+#include <cmath>
+
+static_assert(field_t::TLC == 2, "Norm checking assumes q ~64b");
+
+// extract the number of threads to run from config
+int get_nof_workers(const VecOpsConfig& config); // defined in cpu_vec_ops.cpp
+
+// CPU implementation for icicle::norm::check_norm_bound()
+static eIcicleError cpu_check_norm_bound(
+  const Device& device,
+  const field_t* input,
+  size_t size,
+  eNormType norm,
+  uint64_t norm_bound,
+  const VecOpsConfig& config,
+  bool* output)
+{
+  return eIcicleError::SUCCESS;
+}
+
+// CPU implementation for icicle::norm::check_norm_relative()
+static eIcicleError cpu_check_norm_relative(
+  const Device& device,
+  const field_t* input_a,
+  const field_t* input_b,
+  size_t size,
+  eNormType norm,
+  uint64_t scale,
+  const VecOpsConfig& config,
+  bool* output)
+{
+  return eIcicleError::SUCCESS;
+}
+
+REGISTER_NORM_CHECK_BACKEND("CPU", cpu_check_norm_bound, cpu_check_norm_relative);
--- a/icicle/backend/cpu/src/ring/cpu_norm.cpp
+++ b/icicle/backend/cpu/src/ring/cpu_norm.cpp
@@ -18,6 +18,66 @@ static eIcicleError cpu_check_norm_bound(
   const VecOpsConfig& config,
   bool* output)
 {
+  static_assert(field_t::TLC == 2, "Norm checking assumes q ~64b");
+  constexpr auto q_storage = field_t::get_modulus();
+  const int64_t q = *(const int64_t*)&q_storage;
+
+  if (q < 0) {
+    ICICLE_LOG_ERROR << "Field modulus q must be less than 64 bits; received q = " << q;
+    return eIcicleError::INVALID_ARGUMENT;
+  }
+
+  if (!input || !output || size == 0) {
+    ICICLE_LOG_ERROR << "Invalid argument: null pointer or zero size.";
+    return eIcicleError::INVALID_ARGUMENT;
+  }
+
+  if (config.columns_batch) {
+    ICICLE_LOG_ERROR << "Unsupported config: norm checking does not support column batch.";
+    return eIcicleError::INVALID_ARGUMENT;
+  }
+
+  const int64_t* input_i64 = reinterpret_cast<const int64_t*>(input);
+  bool is_bounded = true;
+
+  if (norm == eNormType::L2) {
+    // For L2 norm, we compute sum of squares and compare with bound^2
+    // We use field_t to handle the intermediate calculations
+    field_t bound_squared = field_t::from(norm_bound) * field_t::from(norm_bound);
+    field_t norm_squared = field_t::from(0);
+
+    for (size_t i = 0; i < size; ++i) {
+      int64_t val = input_i64[i];
+      // Convert to centered representation [-q/2, q/2)
+      if (val > q/2) {
+        val -= q;
+      } else if (val < -q/2) {
+        val += q;
+      }
+      norm_squared = norm_squared + field_t::from(val) * field_t::from(val);
+      if (!field_t::lt(norm_squared, bound_squared)) {
+        is_bounded = false;
+        break;
+      }
+    }
+  } else { // LInfinity norm
+    // For LInfinity norm, we check if any element's absolute value exceeds the bound
+    for (size_t i = 0; i < size; ++i) {
+      int64_t val = input_i64[i];
+      // Convert to centered representation [-q/2, q/2)
+      if (val > q/2) { // TODO: emirsoyturk make this a function
+        val -= q;
+      } else if (val < -q/2) {
+        val += q;
+      }
+      if (std::abs(val) >= static_cast<int64_t>(norm_bound)) {
+        is_bounded = false;
+        break;
+      }
+    }
+  }
+
+  *output = is_bounded;
   return eIcicleError::SUCCESS;
 }
 
@@ -32,6 +92,69 @@ static eIcicleError cpu_check_norm_relative(
   const VecOpsConfig& config,
   bool* output)
 {
+  static_assert(field_t::TLC == 2, "Relative norm checking assumes q ~64b");
+  constexpr auto q_storage = field_t::get_modulus();
+  const int64_t q = *(const int64_t*)&q_storage;
+
+  if (q < 0) {
+    ICICLE_LOG_ERROR << "Field modulus q must be less than 64 bits; received q = " << q;
+    return eIcicleError::INVALID_ARGUMENT;
+  }
+
+  if (!input_a || !input_b || !output || size == 0) {
+    ICICLE_LOG_ERROR << "Invalid argument: null pointer or zero size.";
+    return eIcicleError::INVALID_ARGUMENT;
+  }
+
+  if (config.columns_batch) {
+    ICICLE_LOG_ERROR << "Unsupported config: relative norm checking does not support column batch.";
+    return eIcicleError::INVALID_ARGUMENT;
+  }
+
+  const int64_t* input_a_i64 = reinterpret_cast<const int64_t*>(input_a);
+  const int64_t* input_b_i64 = reinterpret_cast<const int64_t*>(input_b);
+  bool is_bounded = true;
+
+  if (norm == eNormType::L2) {
+    // For L2 norm, we compute sum of squares and compare with (scale * bound)^2
+    field_t scale_bound_squared = field_t::from(scale) * field_t::from(scale);
+    field_t norm_a_squared = field_t::from(0);
+    field_t norm_b_squared = field_t::from(0);
+
+    for (size_t i = 0; i < size; ++i) {
+      int64_t val_a = input_a_i64[i];
+      int64_t val_b = input_b_i64[i];
+      // Convert to centered representation [-q/2, q/2)
+      if (val_a > q/2) val_a -= q;
+      else if (val_a < -q/2) val_a += q;
+      if (val_b > q/2) val_b -= q;
+      else if (val_b < -q/2) val_b += q;
+
+      norm_a_squared = norm_a_squared + field_t::from(val_a) * field_t::from(val_a);
+      norm_b_squared = norm_b_squared + field_t::from(val_b) * field_t::from(val_b);
+    }
+
+    // Check if norm_a^2 < (scale * norm_b)^2
+    is_bounded = field_t::lt(norm_a_squared, scale_bound_squared * norm_b_squared);
+  } else { // LInfinity norm
+    // For LInfinity norm, we check if any element's absolute value exceeds scale * bound
+    for (size_t i = 0; i < size; ++i) {
+      int64_t val_a = input_a_i64[i];
+      int64_t val_b = input_b_i64[i];
+      // Convert to centered representation [-q/2, q/2)
+      if (val_a > q/2) val_a -= q;
+      else if (val_a < -q/2) val_a += q;
+      if (val_b > q/2) val_b -= q;
+      else if (val_b < -q/2) val_b += q;
+
+      if (std::abs(val_a) >= static_cast<int64_t>(scale) * std::abs(val_b)) {
+        is_bounded = false;
+        break;
+      }
+    }
+  }
+
+  *output = is_bounded;
   return eIcicleError::SUCCESS;
 }
 
--- a/icicle/include/icicle/norm.h
+++ b/icicle/include/icicle/norm.h
@@ -1,0 +1,67 @@
+#pragma once
+
+#include "icicle/vec_ops.h" // For VecOpsConfig
+
+namespace icicle {
+
+  /**
+   * @brief Enum to represent supported norm types for vectors.
+   */
+  enum class eNormType {
+    L2 = 0,         ///< Euclidean norm: sqrt(sum of squares)
+    LInfinity       ///< Max norm: maximum absolute element value
+  };
+
+  namespace norm {
+    /**
+     * @brief Checks whether the norm of a vector is within a specified bound.
+     *
+     * @tparam T             Element type of the input vector (e.g., int64_t, uint64_t)
+     * @param input          Pointer to the input vector
+     * @param size           Number of elements in the vector
+     * @param norm           Type of norm to compute (L2 or LInfinity)
+     * @param norm_bound     Upper bound on the norm value
+     * @param cfg            Configuration for vector operations (e.g., threading, SIMD, batching)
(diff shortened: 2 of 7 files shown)


icicle/tests/test_ring_api.cpp
```diff
@@ -230,4 +230,91 @@ TEST_F(RingTestBase, BalancedDecompositionErrorCases)
...
+  // Set specific values that we know should be within bounds
+  input[0] = field_t::from(1);
+  input[1] = field_t::from(2);
+  input[2] = field_t::from(3);
+  input[3] = field_t::from(4);
+  
+  bool is_bounded;
+
+  // Test L2 norm with a generous bound
+  const uint64_t l2_bound = q / 2; // A more generous bound for testing
```
yshekel, line 255: I think this is too generous :) 
Note that for L-infinity q/2 is the max value possible. I think you'd better randomize polynomials with small norm and check tighter bounds for both norms

icicle/tests/test_ring_api.cpp
```diff
@@ -230,4 +230,91 @@ TEST_F(RingTestBase, BalancedDecompositionErrorCases)
...
+
+  const size_t size = 1 << 10;
+  auto input_a = std::vector<field_t>(size);
+  auto input_b = std::vector<field_t>(size);
+  field_t::rand_host_many(input_a.data(), size);
+  field_t::rand_host_many(input_b.data(), size);
+  bool is_bounded;
+
+  // Test relative norm comparison
+  const uint64_t scale = 2; // Compare if norm(a) < 2 * norm(b)
```
yshekel, line 295: I think you need to test more cases and corner cases:
(1) Different scales to test robustness. for example [2,3,17,1000,q-1]
(2) Also test the fail case. For example generate a vector with norm N and test that for N it passes but for N-1 it fails etc. For both APIs
(3) test batch processing too.

icicle/tests/test_ring_api.cpp
```diff
@@ -230,4 +230,91 @@ TEST_F(RingTestBase, BalancedDecompositionErrorCases)
...
+    ICICLE_CHECK(icicle_set_device(device));
+
+    field_t *d_input_a, *d_input_b;
+    ICICLE_CHECK(icicle_malloc((void**)&d_input_a, size * sizeof(field_t)));
+    ICICLE_CHECK(icicle_malloc((void**)&d_input_b, size * sizeof(field_t)));
+    ICICLE_CHECK(icicle_copy(d_input_a, input_a.data(), size * sizeof(field_t)));
+    ICICLE_CHECK(icicle_copy(d_input_b, input_b.data(), size * sizeof(field_t)));
+
+    auto cfg = VecOpsConfig{};
+    cfg.is_a_on_device = true;
```
yshekel, line 306: here you have two inputs but you ignored is_b_on_device. Also the output is not on device. It maybe works because it's CPU but for CUDA it would not

icicle/backend/cpu/src/ring/cpu_norm.cpp
```diff
@@ -0,0 +1,163 @@
...
+    ICICLE_LOG_ERROR << "Unsupported config: norm checking does not support column batch.";
+    return eIcicleError::INVALID_ARGUMENT;
+  }
+
+  const int64_t* input_i64 = reinterpret_cast<const int64_t*>(input);
+  bool is_bounded = true;
+
+  if (norm == eNormType::L2) {
+    // For L2 norm, we compute sum of squares and compare with bound^2
+    uint128_t bound_squared = square(norm_bound, q);
```
yshekel, line 60: I am not sure if this is correct. I mean to square the bound like that means you first compute abs_centered(). I think it should be a simple norm_bound^2.

@Tomer-Solberg

icicle/backend/cpu/src/ring/cpu_norm.cpp
```diff
@@ -0,0 +1,163 @@
...
+    uint128_t norm_squared = 0;
+
+    for (size_t i = 0; i < size; ++i) {
+      int64_t val = input_i64[i];
+      
+      val = abs_centered(val, q);
+      norm_squared += square(val, q);
+      if (norm_squared >= bound_squared) {
```
yshekel, line 68: Note that you also have the case where norm_squared overflows and then it can be smaller than bound_squared before and after the addition. You ignored this case

icicle/backend/cpu/src/ring/cpu_norm.cpp
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// CPU implementation for icicle::norm::check_norm_bound()
+static eIcicleError cpu_check_norm_bound(
+  const Device& device,
+  const field_t* input,
+  size_t size,
+  eNormType norm,
+  uint64_t norm_bound,
+  const VecOpsConfig& config,
```
yshekel, line 33: You need to implement batch too. You can skip column batch for the row batch we need

icicle/backend/cpu/src/ring/cpu_norm.cpp
```diff
@@ -0,0 +1,163 @@
...
+      }
+    }
+  }
+
+  *output = is_bounded;
+  return eIcicleError::SUCCESS;
+}
+
+// CPU implementation for icicle::norm::check_norm_relative()
+static eIcicleError cpu_check_norm_relative(
```
yshekel, line 91: here too need to support batch. Note how balanced_decomposition uses task_flow to parallel. This means that early return needs to be handled carefully too

icicle/backend/cpu/src/ring/cpu_norm.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+      val_a = abs_centered(val_a, q);
+      val_b = abs_centered(val_b, q);
+
+      norm_a_squared += square(val_a, q);
+      norm_b_squared += square(val_b, q);
+    }
+
+    // Check if norm_a^2 < (scale * norm_b)^2
+    is_bounded = norm_a_squared < scale_bound_squared * norm_b_squared;
```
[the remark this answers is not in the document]
Tomer-Solberg: Agree

icicle/include/icicle/norm.h
```diff
@@ -0,0 +1,62 @@
...
+     *                       eIcicleError::INVALID_ARGUMENT or other error codes on failure
+     */
+    template <typename T>
+    eIcicleError check_norm_bound(
+      const T* input, size_t size, eNormType norm, uint64_t norm_bound, const VecOpsConfig& cfg, bool* output);
+
+    /**
+     * @brief Checks whether norm(a) < scalar:u64 * norm(b)
+     *
+     * This is useful in lattice-based schemes and other relative-norm comparisons
```
yshekel, line 36: maybe write here the assumptions he had about elements being at most sqrt(q) and size at most 2^16. If not an error should return.

icicle/tests/test_ring_api.cpp
```diff
@@ -230,4 +235,139 @@ TEST_F(RingTestBase, BalancedDecompositionErrorCases)
...
+  input[3] = field_t::from(1);
+  // LInfinity norm = 4
+  ASSERT_EQ(
+    norm::check_norm_bound(input.data(), size, eNormType::LInfinity, 3, config, &result), eIcicleError::SUCCESS);
+  ASSERT_FALSE(result);
+
+  // Test case 5: Invalid input - element exceeds q
+  constexpr auto q_storage = field_t::get_modulus();
+  const int64_t q = *(const int64_t*)&q_storage;
+  input[0] = field_t::from(q);
```
yshekel, line 291: note that you rely here on from() not reducing mod q. It's kind of a bug/feature in this from() method.
Anyway we can assume values or not >=q since it mod q arithmetics.

icicle/tests/test_ring_api.cpp
```diff
@@ -230,4 +235,139 @@ TEST_F(RingTestBase, BalancedDecompositionErrorCases)
...
+  ASSERT_EQ(
+    norm::check_norm_bound(input.data(), size, eNormType::L2, 10, config, &result), eIcicleError::INVALID_ARGUMENT);
+}
```
yshekel, line 294: one more interesting case is large values since norm looks at the values between (-q/2,q/2] so values close to q should be low norm too. For example norm(q-1)=norm(1)=1.
You can maybe use balanced decomposition on a random vector and check it is low norm as it should be.

icicle/tests/test_ring_api.cpp
```diff
@@ -230,4 +235,139 @@ TEST_F(RingTestBase, BalancedDecompositionErrorCases)
...
+
+TEST_F(RingTestBase, NormRelative)
```
yshekel, line 296: same comment applies here regarding values close to q.

icicle/backend/cpu/src/ring/cpu_norm.cpp
```diff
@@ -0,0 +1,211 @@
...
+  if (!input || !output || size == 0) {
+    ICICLE_LOG_ERROR << "Invalid argument: null pointer or zero size";
+    return eIcicleError::INVALID_ARGUMENT;
+  }
+
+  static_assert(field_t::TLC == 2, "Norm checking assumes q ~64b");
+  constexpr auto q_storage = field_t::get_modulus();
+  const int64_t q = *(const int64_t*)&q_storage;
+
+  if (!validate_inputs(input, size * config.batch_size, q)) { return eIcicleError::INVALID_ARGUMENT; }
```
yshekel, line 49: Note that can be avoided if you check it when you compute. Currently you use 1 thread to iterate over all elements and then parallel the norm bound check.

icicle/backend/cpu/src/ring/cpu_norm.cpp
```diff
@@ -0,0 +1,211 @@
...
+  const uint64_t total_elements = size * config.batch_size;
+  const int nof_workers = get_nof_workers(config);
+  const uint64_t worker_task_size = (total_elements + nof_workers - 1) / nof_workers;
+
+  std::atomic<bool> early_exit{false};
+
+  if (norm == eNormType::L2) {
+    std::vector<__int128> partial_sums(nof_workers, 0);
+    const __int128 bound_squared = static_cast<__int128>(norm_bound) * norm_bound;
+    __int128 total_sum = 0;
```
yshekel, line 65: (1) note that this is not an atomic variable but you use many threads to write/read it in an unsafe way
(2) I think you handle batch incorrectly because you should compute a bool per batch and not compute a norm of the the entire vector as if there is no batch at all.
(3) you don't support column batch so return an error in that case

icicle/backend/cpu/src/ring/cpu_norm.cpp
```diff
@@ -0,0 +1,211 @@
...
+        if (early_exit.load(std::memory_order_relaxed)) return;
+
+        const uint64_t start_idx = worker_id * worker_task_size;
+        const uint64_t end_idx = std::min(start_idx + worker_task_size, total_elements);
+
+        for (uint64_t i = start_idx; i < end_idx; ++i) {
+          int64_t abs_val = abs_centered(input_i64[i], q);
+          __int128 val_squared = static_cast<__int128>(abs_val) * abs_val;
+          total_sum += val_squared;
+          if (total_sum > bound_squared) {
```
yshekel, line 78: note that 'total_sum += val_squared' can overflow and result in a small value. In this case the norm is obviously not bound by a 64b bound but your code may return true in that case

icicle/backend/cpu/src/ring/cpu_norm.cpp
```diff
@@ -0,0 +1,211 @@
...
+    executor.run(taskflow).wait();
+    taskflow.clear();
+  } else { // L∞ norm
+    for (uint64_t start_idx = 0; start_idx < total_elements; start_idx += worker_task_size) {
+      taskflow.emplace([=, &early_exit]() {
+        if (early_exit.load(std::memory_order_relaxed)) return;
+
+        const uint64_t end_idx = std::min(start_idx + worker_task_size, total_elements);
+
+        for (uint64_t i = start_idx; i < end_idx; ++i) {
```
yshekel, line 96: same comments here about batch

icicle/backend/cpu/src/ring/cpu_norm.cpp
```diff
@@ -0,0 +1,211 @@
...
+    ICICLE_LOG_ERROR << "Invalid argument: null pointer or zero size";
+    return eIcicleError::INVALID_ARGUMENT;
+  }
+
+  static_assert(field_t::TLC == 2, "Norm checking assumes q ~64b");
+  constexpr auto q_storage = field_t::get_modulus();
+  const int64_t q = *(const int64_t*)&q_storage;
+
+  if (
+    !validate_inputs(input_a, size * config.batch_size, q) || !validate_inputs(input_b, size * config.batch_size, q)) {
```
yshekel, line 134: here we agreed that values do not exceed sqrt(q) and at most 2^16 elements but you check something else. And also can move the check of sqrt(q) inside

icicle/backend/cpu/src/ring/cpu_norm.cpp
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  return eIcicleError::SUCCESS;
+}
+
+static eIcicleError cpu_check_norm_relative(
+  const Device& device,
+  const field_t* input_a,
+  const field_t* input_b,
+  size_t size,
```
yshekel, line 118: maybe input_a and input_b may have different size. @Tomer-Solberg is it interesting? 
Currently we assume at most 2^16 elements, each at most sqrt(q) in value so we only need 80b but we can support different sizes if required.

icicle/backend/cpu/src/ring/cpu_norm.cpp
```diff
@@ -0,0 +1,211 @@
...
+        if (early_exit.load(std::memory_order_relaxed)) return;
+
+        const uint64_t end_idx = std::min(start_idx + worker_task_size, total_elements);
+
+        for (uint64_t i = start_idx; i < end_idx; ++i) {
+          int64_t abs_a = abs_centered(input_a_i64[i], q);
+          int64_t abs_b = abs_centered(input_b_i64[i], q);
+
+          // |a| >= scale * |b|
+          if (abs_a >= scale * abs_b) {
```
yshekel, line 194: same comments about early exit and batching.

yshekel (changes requested):
Left comments, I think the current implementation is wrong but the tests still pass so maybe I would start from making tests that fail (for the reasons I mentioned in the review) and then fix the code to make them pass

icicle/backend/cpu/src/ring/cpu_norm.cpp
```diff
@@ -1,5 +1,6 @@
 #include "icicle/norm.h"
 #include "icicle/backend/vec_ops_backend.h"
+#include "icicle/rings/integer_rings/labrador.h"  // For field_t definition
```
yshekel, line 3: you need to use
```
#include "icicle/fields/field_config.h"
using namespace field_config;
```

so that field_t is defined for other libraries too, not only labrador, but anyway I think you can remove this include sincec `vec_ops_backend.h` is already including it.

icicle/backend/cpu/src/ring/cpu_norm.cpp
```diff
@@ -37,75 +27,88 @@ static eIcicleError cpu_check_norm_bound(
...
-
-        for (uint64_t i = start_idx; i < end_idx; ++i) {
-          int64_t abs_val = abs_centered(input_i64[i], q);
-          if (abs_val >= norm_bound) {
-            *output = false;
-            early_exit.store(true, std::memory_order_relaxed);
-            return;
-          }
+
+    *output = sum_squares.load(std::memory_order_relaxed) <= bound_squared;
```
yshekel, line 79: This is correct but note that in this bound case you may be able to have early exit (if you go above the bound since the bound is constant, unlike the other relative case where you really cannot early exit).

icicle/backend/cpu/src/ring/cpu_norm.cpp
```diff
@@ -37,75 +27,88 @@ static eIcicleError cpu_check_norm_bound(
...
+
+    for (uint64_t start_idx = 0; start_idx < size; start_idx += worker_task_size) {
+      taskflow.emplace([=, &sum_squares]() {
+        const uint64_t end_idx = std::min(start_idx + worker_task_size, size);
+        uint128_t local_sum = 0;
+
+        for (uint64_t idx = start_idx; idx < end_idx; ++idx) {
+          // TODO: emirsoyturk validate input value is in the range [0, sqrt(q))
+          int64_t val = abs_centered(input_i64[idx], q);
+          local_sum += static_cast<uint128_t>(val) * static_cast<uint128_t>(val);
```
yshekel, line 69: note that local_sum and sum_squares may overflow but you don't account for it.
If you assume (and check!) here too that values are at most sqrt(q) and that we have size<=2^16 then I think you can safely ignore the overflow since you won't need more than 80b but you have 128b. But if you don't assume and check it then you could have overflow.

icicle/tests/test_ring_api.cpp
```diff
@@ -239,135 +247,74 @@ TEST_F(RingTestBase, BalancedDecompositionErrorCases)
...
-  const int64_t q = *(const int64_t*)&q_storage;
-  input[0] = field_t::from(q);
-  ASSERT_EQ(
-    norm::check_norm_bound(input.data(), size, eNormType::L2, 10, config, &result), eIcicleError::INVALID_ARGUMENT);
+  const int64_t q = *(int64_t*)&q_storage; // Note this is valid since TLC == 2
+  ICICLE_ASSERT(q > 0) << "Expecting at least one slack bit to use int64 arithmetic";
+
+  const size_t size = 1 << 10;
+  auto input = std::vector<field_t>(size);
+  field_t::rand_host_many(input.data(), size);
```
yshekel, line 257: note that you may get values larger than sqrt(q) here.

icicle/tests/test_ring_api.cpp
```diff
@@ -239,135 +247,74 @@ TEST_F(RingTestBase, BalancedDecompositionErrorCases)
...
+  bool output;
+
+  // Test L2 norm
+  {
+    uint64_t actual_norm_squared = 0;
+    for (size_t i = 0; i < size; ++i) {
+      int64_t val = abs_centered(*(int64_t*)&input[i], q);
+      actual_norm_squared += static_cast<uint64_t>(val) * static_cast<uint64_t>(val);
```
yshekel, line 265: note you need to use 128b here and also for actual_norm_squared

PR #835, opened 2025-04-02
Adding Goldilocks Field

HadarIngonyama:
This PR adds support for the Goldilocks field, but it requires using an extended representation for Field elements that is not yet supported.

Multiplication, addition and subtraction are optimized for CPU.
NTT is not optimized yet.


icicle/cmake/features.cmake
```diff
@@ -6,6 +6,7 @@ set(ICICLE_FIELDS
   1002:stark252:NTT,POSEIDON,POSEIDON2,SUMCHECK,FRI
   1003:m31:EXT_FIELD,POSEIDON,POSEIDON2,SUMCHECK
   1004:koalabear:NTT,EXT_FIELD,POSEIDON,POSEIDON2,SUMCHECK,FRI
+  1005:goldilocks:NTT,SUMCHECK
```
nonam3e, line 9: do we need to support poseidon and sumcheck for goldilocks?

icicle/include/icicle/fields/stark_fields/goldilocks.h
```diff
@@ -0,0 +1,325 @@
...
+
+  template <class CONFIG>
+  class GoldilocksField : public Field<CONFIG>
+  {
+  public:
+    static constexpr unsigned TLC = CONFIG::limbs_count;
+    typedef storage<TLC> ff_storage;
+
+    HOST_DEVICE_INLINE GoldilocksField(const GoldilocksField& other) : Field<CONFIG>(other) {}
+    HOST_DEVICE_INLINE GoldilocksField(const uint32_t& x = 0) : Field<CONFIG>({x}) {}
```
nonam3e, line 26: I think it's better to use `Field<CONFIG>::from` here. As I see, `limbs_count = 2`

icicle/cmake/features.cmake
omershlo: yes

yshekel: also ext_field and FRI no?

icicle/include/icicle/fields/params_gen.h
```diff
@@ -63,6 +63,7 @@ namespace params_gen {
     storage<2 * NLIMBS> pow_of_2 = host_math::template left_shift<2 * NLIMBS, NBITS>(one);
     host_math::template multiply_raw<NLIMBS, 2 * NLIMBS, true>(modulus, pow_of_2, x3);
     host_math::template add_sub_limbs<3 * NLIMBS, true, false, true>(x3, x2, x2);
+    if (pow_of_2.limbs[2 * NLIMBS - 1] == 0) return 1; // todo - verify this
```
yshekel, line 66: Please either finalize this or at least put your name like 'TODO Hadar'. Otherwise it will be here forever and we don't even know who is supposed to handle the TODO

icicle/include/icicle/fields/stark_fields/goldilocks.h
```diff
@@ -0,0 +1,331 @@
...
+1. It has no slack bits (the modulus uses the entire 64 bits of storage) meaning we need to make sure there is no
+overflow in addition, and that the carry-less optimizations used in other fields do not apply here.
+2. It has a special reduction algorithm due to the modulus special form.
+3. In order to optimize addition and multiplication - the elements are reduced to the range 0<=x<2^64 instead of 0=<x<p.
+This requires changing the operator== implementation such that it will return true for values with a difference of p.
+4. The operations that are implemented differently from the base Field class are: addition, reduction, inverse and
+comparison. The specific details are documented in the relevant functions. */
+namespace goldilocks {
+
+  template <class CONFIG>
```
yshekel, line 18: I find it strange that Goldilocks needs a CONFIG here. It's not like you can instantiate it with different CONFIGS righs?

I would derive 
class Goldilocks : public Field<GOLDILOCKS_CONFIG> {...}

icicle/include/icicle/fields/stark_fields/goldilocks.h
```diff
@@ -0,0 +1,331 @@
...
+  class GoldilocksField : public Field<CONFIG>
```
yshekel, line 19: please rename to Goldilocks to align with codebase. Other fields are not BabybearField but Babybear

icicle/include/icicle/fields/stark_fields/goldilocks.h
```diff
@@ -0,0 +1,331 @@
...
+    the arguments are larger than p. Sometimes we can guarantee that one of the arguments is smaller than p and then we
+    use NO_OVERFLOW=true. That is the case in the call that is inside the reduction function. When we can't guarantee
+    this then we hint the compiler that this is a rare case.*/
+    template <bool NO_OVERFLOW = false>
+    static HOST_DEVICE_INLINE GoldilocksField goldi_add(const GoldilocksField& xs, const GoldilocksField& ys)
+    {
+      GoldilocksField rs = {};
+      const ff_storage modulus = Field<CONFIG>::get_modulus();
+      if constexpr (NO_OVERFLOW == false) { // Handle the rare case where we would need to subtract 2p
+        if (__builtin_expect(xs.limbs_storage.limbs64[0] >= modulus.limbs64[0], 0))
```
yshekel, line 83: use curly brackets, this is super bug prone. Also below, actually for every if-else block

icicle/include/icicle/fields/stark_fields/goldilocks.h
```diff
@@ -0,0 +1,331 @@
...
+    addition with NO_OVERFLOW=true*/
+    static constexpr HOST_DEVICE_INLINE GoldilocksField reduce(const typename Field<CONFIG>::Wide xs)
+    {
+      constexpr uint32_t gold_fact = uint32_t(-1); //(2^32 - 1)
+      GoldilocksField rs = {};
+      const GoldilocksField& x_lo = *reinterpret_cast<const GoldilocksField*>(xs.limbs_storage.limbs);
+      GoldilocksField x_hi_hi = Field<CONFIG>::from(xs.limbs_storage.limbs[3]);
+      auto carry = Field<CONFIG>::template sub_limbs<TLC, true>(
+        x_lo.limbs_storage, x_hi_hi.limbs_storage, rs.limbs_storage); // xs[63:0] - xs[127:96]
+      if (__builtin_expect(carry, 0)) {
```
yshekel, line 127: does this work for CUDA too? is it really improving something?

icicle/include/icicle/fields/stark_fields/goldilocks.h
```diff
@@ -0,0 +1,331 @@
...
+      return (u == one) ? b : c;
+    }
+
+    friend HOST_DEVICE_INLINE GoldilocksField operator*(const GoldilocksField& xs, const GoldilocksField& ys)
+    {
+      typename Field<CONFIG>::Wide xy = Field<CONFIG>::mul_wide(xs, ys);
+      return reduce(xy);
+    }
+
+    /*Since we allow the elements to be between 0 and 2^64, if they are larger than p we need to subtract p before the
```
yshekel, line 189: why do you need them be larger than p?

icicle/include/icicle/fields/stark_fields/goldilocks.h
```diff
@@ -0,0 +1,331 @@
...
+        exp >>= 1;
+      }
+      return res;
+    }
+  };
+
+  struct fp_config {
+    static constexpr storage<2> modulus = {0x00000001, 0xffffffff}; // 2^64 - 2^32 + 1
+    static constexpr unsigned reduced_digits_count = 9;
+    static constexpr storage_array<reduced_digits_count, 2> reduced_digits = {
```
yshekel, line 267: Please add scripts you used to generate those into scripts/python dir (Assuming python) with instructions how to use.

icicle/tests/test_field_api.cpp
```diff
@@ -19,6 +19,54 @@ TYPED_TEST(FieldTest, FieldSanityTest)
...
   auto b = TypeParam::rand_host();
+  auto c = TypeParam::rand_host();
+  auto d = TypeParam::rand_host();
+  auto e = TypeParam::rand_host();
+  auto f = TypeParam::rand_host();
+  auto g = TypeParam::rand_host();
+  auto h = TypeParam::rand_host();
+  auto i = TypeParam::rand_host();
+  auto j = TypeParam::rand_host();
+  START_TIMER(mult)
```
yshekel, line 30: remove this. You are not testing anything here anyway

wrappers/rust/icicle-fields/icicle-goldilocks/build.rs
```diff
@@ -0,0 +1,71 @@
...
+    let icicle_install_dir = if let Ok(dir) = env::var("ICICLE_INSTALL_DIR") {
+        PathBuf::from(dir)
+    } else {
+        // Define the default install directory to be under the build directory
+        PathBuf::from(format!("{}/icicle/", deps_dir.display()))
+    };
+    config
+        .define("FIELD", "goldilocks")
+        .define("HASH", "ON")
+        .define("EXT_FIELD", "OFF")  // Disable extension field support
```
yshekel, line 29: no ext?

wrappers/rust/icicle-fields/icicle-goldilocks/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+pub mod field;
+pub mod ntt;
+pub mod polynomials;
+// TODO: uncomment once implemented for goldilocks
```
yshekel, line 4: Maybe remove the poseidon module until then from this crate? Seems strange to have code that doesn't compile.

icicle/include/icicle/fields/stark_fields/goldilocks.h
yshekel: It means some numbers have two representations. Can you avoid it?

@mickeyasa @HadarIngonyama

icicle/include/icicle/fields/stark_fields/goldilocks.h
HadarIngonyama: don't understand, the from constructor is below

icicle/include/icicle/fields/stark_fields/goldilocks.h
HadarIngonyama: In that case mersenne should be changed too

icicle/include/icicle/fields/stark_fields/goldilocks.h
HadarIngonyama: other fields don't have a class of their own.. I don't see the misalignment here

icicle/include/icicle/fields/stark_fields/goldilocks.h
HadarIngonyama: ok

icicle/include/icicle/fields/stark_fields/goldilocks.h
HadarIngonyama: it improves significantly for cpu. I will implement a separate gpu implementation soon

icicle/include/icicle/fields/stark_fields/goldilocks.h
HadarIngonyama: Only at the cost of performance because then and extra -p would we needed.

yshekel: BTW I think this must be explained in the design doc before coding. It's causing trouble for FFI and possibly other things too

icicle/include/icicle/fields/stark_fields/goldilocks.h
HadarIngonyama: ok

icicle/tests/test_field_api.cpp
HadarIngonyama: I'm testing the performance. Where would be the appropriate place to put it? It's about time we have a bench for basic operations..

icicle/tests/test_field_api.cpp
```diff
@@ -1074,29 +1122,29 @@ TYPED_TEST(FieldTest, FriShouldFailCases)
...
-//       ASSERT_EQ(scalar_t::from(product), scalar_t::one());
-//       std::byte* a_bytes = reinterpret_cast<std::byte*>(a.limbs);
-//       std::byte* b_bytes = reinterpret_cast<std::byte*>(b.limbs);
-//       std::byte* sum_bytes = reinterpret_cast<std::byte*>(sum.limbs);
-//       std::byte* product_bytes = reinterpret_cast<std::byte*>(product.limbs);
-//       ASSERT_EQ(scalar_t::from(a), scalar_t::from(a_bytes, 18 * 4));
-//       ASSERT_EQ(scalar_t::from(a_bytes, 18 * 4) + scalar_t::from(b_bytes, 18 * 4), scalar_t::from(sum_bytes, 18 *
-//       4)); ASSERT_EQ(scalar_t::from(product_bytes, 4 * 4), scalar_t::one());
-//     } else {
+//     // if constexpr (scalar_t::TLC == 1) {
```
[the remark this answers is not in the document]
HadarIngonyama: will fix in future pr, doesn't matter right now

icicle/include/icicle/fields/stark_fields/goldilocks.h
HadarIngonyama: what trouble is it causing? I can add it to the doc

HadarIngonyama:
This requires using an extended representation for Field elements that is not yet supported.

PR #836, opened 2025-04-02
copy | rust

emirsoyturk:
## Describe the changes

This PR...

## Linked Issues

Resolves #

### DIFF
--- a/wrappers/rust/icicle-runtime/src/memory.rs
+++ b/wrappers/rust/icicle-runtime/src/memory.rs
@@ -254,6 +254,9 @@ impl<T> DeviceSlice<T> {
         if !self.is_on_active_device() {
             panic!("not allocated on an inactive device");
         }
+        if !val.is_on_active_device() {
+            panic!("source is not allocated on an active device");
+        }
 
         let size = size_of::<T>() * self.len();
         unsafe { runtime::icicle_copy(self.as_mut_ptr() as *mut c_void, val.as_ptr() as *const c_void, size).wrap() }
@@ -270,6 +273,9 @@ impl<T> DeviceSlice<T> {
         }
         if !self.is_on_active_device() {
             panic!("not allocated on an inactive device");
+        }
+        if !val.is_on_active_device() {
+            panic!("source is not allocated on an active device");
         }
 
         let size = size_of::<T>() * self.len();
--- a/wrappers/rust/icicle-runtime/src/memory.rs
+++ b/wrappers/rust/icicle-runtime/src/memory.rs
@@ -20,42 +20,7 @@ pub trait HostOrDeviceSlice<T> {
     unsafe fn as_mut_ptr(&mut self) -> *mut T;
     fn len(&self) -> usize;
     fn is_empty(&self) -> bool;
-    fn copy(&mut self, src: &(impl HostOrDeviceSlice<T> + ?Sized)) -> Result<(), eIcicleError>;
-    fn copy_async(
-        &mut self,
-        src: &(impl HostOrDeviceSlice<T> + ?Sized),
-        stream: &IcicleStream,
-    ) -> Result<(), eIcicleError>;
-}
-
-impl<T> HostOrDeviceSlice<T> for HostSlice<T> {
-    fn is_on_device(&self) -> bool {
-        false
-    }
-
-    fn is_on_active_device(&self) -> bool {
-        false
-    }
-
-    unsafe fn as_ptr(&self) -> *const T {
-        self.0
-            .as_ptr()
-    }
-
-    unsafe fn as_mut_ptr(&mut self) -> *mut T {
-        self.0
-            .as_mut_ptr()
-    }
-
-    fn len(&self) -> usize {
-        self.0
-            .len()
-    }
-
-    fn is_empty(&self) -> bool {
-        self.len() == 0
-    }
-
+    
     fn copy(&mut self, src: &(impl HostOrDeviceSlice<T> + ?Sized)) -> Result<(), eIcicleError> {
         assert!(
             self.len() >= src.len(),
@@ -86,6 +51,35 @@ impl<T> HostOrDeviceSlice<T> for HostSlice<T> {
             )
             .wrap()
         }
+    }
+}
+
+impl<T> HostOrDeviceSlice<T> for HostSlice<T> {
+    fn is_on_device(&self) -> bool {
+        false
+    }
+
+    fn is_on_active_device(&self) -> bool {
+        false
+    }
+
+    unsafe fn as_ptr(&self) -> *const T {
+        self.0
+            .as_ptr()
+    }
+
+    unsafe fn as_mut_ptr(&mut self) -> *mut T {
+        self.0
+            .as_mut_ptr()
+    }
+
+    fn len(&self) -> usize {
+        self.0
+            .len()
+    }
+
+    fn is_empty(&self) -> bool {
+        self.len() == 0
     }
 }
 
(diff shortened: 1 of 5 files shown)


yshekel (commented):
What if we do it for HostOrDeviceSlice? In that case we can slice two slices with same size and copy using icicle_copy() which identifies device/host and handles correctly. So this should work for any subslice of host/device memory and also with any offset into the memory.
The way you implemented allows only for a full DeviceVec to be copied to a full DeviceVec.
Also if we can subslice a HostOrDeviceSlice that would be amazing. I am not sure if this is possible

wrappers/rust/icicle-runtime/src/memory.rs
```diff
@@ -241,6 +241,48 @@ impl<T> DeviceSlice<T> {
...
+    pub fn copy_from_device(&mut self, val: &DeviceSlice<T>) -> Result<(), eIcicleError> {
+        assert!(
+            self.len() == val.len(),
+            "In copy from device, destination and source slices have different lengths"
+        );
+
+        if self.is_empty() {
+            return Ok(());
+        }
+        if !self.is_on_active_device() {
```
yshekel, line 254: also need to check this for val right?

wrappers/rust/icicle-runtime/src/memory.rs
```diff
@@ -241,6 +241,48 @@ impl<T> DeviceSlice<T> {
             .wrap()
         }
     }
+
+    pub fn copy_from_device(&mut self, val: &DeviceSlice<T>) -> Result<(), eIcicleError> {
+        assert!(
```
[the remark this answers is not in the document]
LeonHibnik: We don't, it's better to check and return an error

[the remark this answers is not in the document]
yshekel: I mean the restriction that both Vectors are same size.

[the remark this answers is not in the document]
emirsoyturk: Leon also mentioned this but i think since host->device and device->host also doesn't support working with different lengths either we should disable it for all of them or keep it like that and add examples to show how to copy between different sizes with taking slice

wrappers/rust/icicle-runtime/src/memory.rs
emirsoyturk: right

wrappers/rust/icicle-runtime/src/memory.rs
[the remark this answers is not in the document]
yshekel: I think we should add a copy function with the following signature:
```
fn copy<T>(
    src: &(impl HostOrDeviceSlice<T> + ?Sized),
    dst: &mut (impl HostOrDeviceSlice<T> + ?Sized),
);
```

This function should support copying between any HostSlice and DeviceSlice, regardless of their origin or storage type. The slices can be of any size, as long as they match. Since we’re working with slices (not full Vecs or DeviceVecs), it’s fine for the slices to represent only a portion of the underlying buffer. This naturally supports copying with offsets into Vec or DeviceVec, allowing for flexible and efficient memory operations.

This addition was requested also for the collaborative SNARK grant implementation.

Let’s make sure to keep the existing API unchanged to avoid breaking V3. We’re only adding this new helper function.

wrappers/rust/icicle-runtime/src/memory.rs
```diff
@@ -76,9 +97,29 @@ impl<T> HostOrDeviceSlice<T> for DeviceSlice<T> {
...
     }
+
+    fn copy(&mut self, src: &(impl HostOrDeviceSlice<T> + ?Sized)) -> Result<(), eIcicleError> {
+        let on_device = src.is_on_device();
+
+        if on_device {
+            //TODO: emirsoyturk add checks
+            let size = size_of::<T>() * src.len();
+            unsafe {
+                runtime::icicle_copy(self.as_mut_ptr() as *mut c_void, src.as_ptr() as *const c_void, size).wrap()
```
yshekel, line 112: again, you can always use icicle_copy() directly

wrappers/rust/icicle-runtime/src/memory.rs
```diff
@@ -113,6 +154,25 @@ impl<T> HostOrDeviceSlice<T> for DeviceVec<T> {
...
+            }
+        } else {
+            //TODO: emirsoyturk add checks
+            let size = size_of::<T>() * src.len();
+            unsafe {
+                runtime::icicle_copy_to_device(self.as_mut_ptr() as *mut c_void, src.as_ptr() as *const c_void, size)
```
yshekel, line 171: same

wrappers/rust/icicle-runtime/src/tests.rs
```diff
@@ -123,6 +123,74 @@ mod tests {
...
+        assert_ne!(input, output);
+
+        // Copy from host to first device
+        let mut d_mem1 = DeviceVec::device_malloc(input.len()).unwrap();
+        d_mem1
+            .copy_from_host(HostSlice::from_slice(&input))
+            .unwrap();
+
+        // Copy from first device to second device
+        let mut d_mem2 = DeviceVec::device_malloc(input.len()).unwrap();
```
yshekel, line 142: let's make d_mem2 larger and copy into an offset into it. FOr example instead of 4 elements make it 10 and write to some offset. This is the general case

emirsoyturk: can you check it again?

emirsoyturk: https://github.com/ingonyama-zk/icicle/pull/836/files#diff-555317aec66e1066f3e99872579f0f14677afd1b036588453925865e54afe802R61-R62


they can take a slice still. but i think this is better dev UX

wrappers/rust/icicle-runtime/src/tests.rs
```diff
@@ -123,6 +123,74 @@ mod tests {
         assert_eq!(input, output);
     }
 
+    #[test]
+    fn test_device_to_device_copy() {
```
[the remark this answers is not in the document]
LeonHibnik: I agree, this crate is only tested manually rn, @emirsoyturk

wrappers/rust/icicle-runtime/src/memory.rs
```diff
@@ -45,9 +51,38 @@ impl<T> HostOrDeviceSlice<T> for HostSlice<T> {
...
     }
+
     fn is_empty(&self) -> bool {
         self.len() == 0
     }
+
+    fn copy(&mut self, src: &(impl HostOrDeviceSlice<T> + ?Sized)) -> Result<(), eIcicleError> {
+        //TODO: emirsoyturk add checks
+        let min_len = core::cmp::min(self.len(), src.len());
+        let size = size_of::<T>() * min_len;
```
LeonHibnik, line 62: Do we want to implicitly support truncating a slice?

emirsoyturk: i think there is no harm but i couldn't be sure also. wdyt?

yshekel: I think the slices should be of same size, otherwise error. But then also we maybe need to be able to sub slice a slice. I mean something like slice[x..y]

LeonHibnik: I think that we can copy to an extended slice but truncating should be explicit like @yshekel mentioned

emirsoyturk: i thought we were going to remove same size assertion

LeonHibnik: I mean here check that you are copying into a same or larger slice

LeonHibnik (approved):

wrappers/rust/icicle-runtime/src/memory.rs
```diff
@@ -45,9 +51,42 @@ impl<T> HostOrDeviceSlice<T> for HostSlice<T> {
         self.0
             .len()
     }
+
     fn is_empty(&self) -> bool {
         self.len() == 0
     }
+
+    fn copy(&mut self, src: &(impl HostOrDeviceSlice<T> + ?Sized)) -> Result<(), eIcicleError> {
```
yshekel, line 59: can you implement this only once in the trait itself? Seems the exact same implementation 3 times

yshekel (approved):
Approved but see one more comment

yshekel (approved):

PR #837, opened 2025-04-06
Adding Goldilocks Field

HadarIngonyama:
This PR adds support for the Goldilocks field.  It doesn't require the use of the extended representation.

Multiplication, addition and subtraction are optimized for CPU.
NTT is not optimized yet.


yshekel (approved):

wrappers/rust/icicle-fields/icicle-goldilocks/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+pub mod field;
+pub mod ntt;
+pub mod polynomials;
+// TODO: uncomment once implemented for goldilocks
+// pub mod poseidon;
+// pub mod poseidon2;
```
jeremyfelder, line 6: Danny added poseidon2 constants in #840, who is adding goldilocks poseidon2 for the backends?

jeremyfelder (commented):
Can we close #835?

jeremyfelder (approved):

PR #847, opened 2025-04-07
Sha3Large test + bigger batch sizes

dorshmil:
- added a test for large size sha3
- increased the default batch size in all the Large hash tests

cuda-backend-branch: hash_rewrite


yshekel (approved):

LeonHibnik (approved):

jeremyfelder:
@dorshmil What is the difference between the `Large` and `LargeUgly` tests?

jeremyfelder:
Do we need large tests running on each PR? Would it be better to run these on some other cadence?

PR #849, opened 2025-04-08
[FEAT] Rust memset

LeonHibnik:
## Describe the changes

This PR adds support for using icicle_memset in icicle rust wrapper

## Linked Issues

Resolves #

### DIFF
--- a/wrappers/rust/icicle-runtime/src/memory.rs
+++ b/wrappers/rust/icicle-runtime/src/memory.rs
@@ -20,6 +20,9 @@ pub trait HostOrDeviceSlice<T> {
     unsafe fn as_mut_ptr(&mut self) -> *mut T;
     fn len(&self) -> usize;
     fn is_empty(&self) -> bool;
+
+    fn memset(&mut self, value: u8, size: usize) -> Result<(), eIcicleError>;
+    fn memset_async(&mut self, value: u8, size: usize, stream: &IcicleStream) -> Result<(), eIcicleError>;
 }
 
 impl<T> HostOrDeviceSlice<T> for HostSlice<T> {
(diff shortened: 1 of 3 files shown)


wrappers/rust/icicle-runtime/src/memory.rs
```diff
@@ -241,6 +241,30 @@ impl<T> DeviceSlice<T> {
             .wrap()
         }
     }
+
+    pub fn memset(&mut self, value: u8) -> Result<(), eIcicleError> {
+        if self.is_empty() {
+            return Ok(());
+        }
+        if !self.is_on_active_device() {
+            panic!("not allocated on an inactive device");
```
yshekel, line 250: the comment is wrong

wrappers/rust/icicle-runtime/src/tests.rs
```diff
@@ -146,4 +146,49 @@ mod tests {
         assert_eq!(config_ext.get_int("example_int"), 42);
         assert_eq!(config_ext.get_bool("example_bool"), true);
     }
+    #[test]
+    fn test_memset() {
```
yshekel, line 150: I think those tests do not execute in CI BTW

LeonHibnik: Solved in #836

wrappers/rust/icicle-runtime/src/memory.rs
```diff
@@ -284,6 +308,20 @@ impl<T> DeviceVec<T> {
             )))))
         }
     }
+
+    pub fn as_mut_slice(&mut self) -> &mut DeviceSlice<T> {
+        &mut self[..]
+    }
+
+    pub fn memset(&mut self, value: u8) -> Result<(), eIcicleError> {
```
yshekel, line 316: Given we can memset a deviceSlice why do we need it also for DeviceVec? 
And here also it's memset the entire vector but usually we want to to memset offset+:size.

Can we implement it on HostOrDeviceSlice instead? If yes then we can memset any memory with offest+size, regardless being host/device memory. We probably also would need to sub-slice HostOrDeviceSlice to be fully flexible.

nonam3e (approved):

wrappers/rust/icicle-runtime/src/memory.rs
```diff
@@ -81,6 +84,32 @@ impl<T> HostOrDeviceSlice<T> for HostSlice<T> {
     fn is_empty(&self) -> bool {
         self.len() == 0
     }
+
+    fn memset(&mut self, value: u8, size: usize) -> Result<(), eIcicleError> {
+        if size == 0 || self.is_empty() {
```
yshekel, line 89: This is a strange case, why not panic? should be super rare but why would someone do it?

yshekel (approved):

PR #861, opened 2025-04-16
Serialization

nonam3e:
This PR adds serialization support for proofs


icicle/include/icicle/serialization.h
```diff
@@ -0,0 +1,52 @@
...
+  public:
+    virtual ~Serializer() = default;
+
+    virtual eIcicleError serialized_size(size_t& size) const { return eIcicleError::API_NOT_IMPLEMENTED; }
+
+    virtual eIcicleError serialize(std::byte*& out) const { return eIcicleError::API_NOT_IMPLEMENTED; }
+
+    virtual eIcicleError deserialize(std::byte*& in, size_t& length) { return eIcicleError::API_NOT_IMPLEMENTED; }
+
+    eIcicleError serialize_to_file(const std::string& filename) const
```
ChickenLover, line 19: Maybe change to `path`

icicle/include/icicle/serialization.h
```diff
@@ -0,0 +1,52 @@
+#pragma once
+
+#include <cstddef>
+#include <fstream>
+#include <vector>
+
+namespace icicle {
+  class Serializer
```
ChickenLover, line 8: Maybe `Serializable` or `IsSerializable` would be a better name to highlight that this is an interface

icicle/src/sumcheck/sumcheck_c_api.cpp
```diff
@@ -260,5 +260,122 @@ eIcicleError CONCAT_EXPAND(ICICLE_FFI_PREFIX, sumcheck_proof_print)(SumcheckProo
...
   return eIcicleError::SUCCESS;
 }
+
+/**
+ * @brief Gets the serialized size of the given SumcheckProof instance.
+ * @param sumcheck_proof_handle Pointer to the SumcheckProof instance.
+ * @param size Pointer to store the serialized size of the SumcheckProof instance.
+ * @return eIcicleError indicating the success or failure of the operation.
+ */
+eIcicleError CONCAT_EXPAND(ICICLE_FFI_PREFIX, sumcheck_proof_get_serialized_size)(
```
LeonHibnik, line 270: IMHO if you are using it inside a SumcheckProof object, a user doesn't need to use that explicitly in the deserialization

icicle/include/icicle/fri/fri_proof.h
```diff
@@ -112,6 +115,125 @@ namespace icicle {
...
+      size += sizeof(uint64_t); // pow_nonce
+
+      return eIcicleError::SUCCESS;
+    }
+
+    eIcicleError serialize(std::byte*& out) const override
+    {
+      size_t m_query_proofs_size = m_query_proofs.size();
+      std::memcpy(out, &m_query_proofs_size, sizeof(size_t));
+      out += sizeof(size_t);
```
ChickenLover, line 143: There must be some kind of solution that will let you skip this out += each time.

icicle/include/icicle/fri/fri_proof.h
```diff
@@ -112,6 +115,125 @@ namespace icicle {
...
+      for (const std::vector<MerkleProof>& query_proofs : m_query_proofs) {
+        size_t query_proofs_size = query_proofs.size();
+        std::memcpy(out, &query_proofs_size, sizeof(size_t));
+        out += sizeof(size_t);
+        for (const MerkleProof& proof : query_proofs) {
+          eIcicleError err = proof.serialize(out);
+          if (err != eIcicleError::SUCCESS) { return err; }
```
ChickenLover, line 150: I think we have a macro for this

LeonHibnik (commented):
Nice!
I think that serialized_size can be implicit

icicle/include/icicle/fri/fri_proof.h
ChickenLover: even a dedicated function that will run memcpy + update the pointer would be fine. But I am pretty sure smth like this exists in std

icicle/src/sumcheck/sumcheck_c_api.cpp
nonam3e: right, this method is used only before serialization to allow user to allocate a buffer with the right size

icicle/include/icicle/fri/fri_proof.h
```diff
@@ -112,6 +115,125 @@ namespace icicle {
...
+      std::memcpy(out, m_final_poly.data(), final_poly_size * sizeof(F));
+      out += final_poly_size * sizeof(F);
+      std::memcpy(out, &m_pow_nonce, sizeof(uint64_t));
+      out += sizeof(uint64_t);
+      return eIcicleError::SUCCESS;
+    }
+
+    eIcicleError deserialize(std::byte*& in, size_t& length) override
+    {
+      auto advance = [&](size_t bytes) -> bool {
```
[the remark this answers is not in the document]
nonam3e: will add a method in `Serialize` class which will check the remaining size + copy + shift the pointer in one call

icicle/include/icicle/serialization.h
```diff
@@ -0,0 +1,52 @@
...
+  {
+  public:
+    virtual ~Serializer() = default;
+
+    virtual eIcicleError serialized_size(size_t& size) const { return eIcicleError::API_NOT_IMPLEMENTED; }
+
+    virtual eIcicleError serialize(std::byte*& out) const { return eIcicleError::API_NOT_IMPLEMENTED; }
+
+    virtual eIcicleError deserialize(std::byte*& in, size_t& length) { return eIcicleError::API_NOT_IMPLEMENTED; }
```
ChickenLover, line 17: It's not very convenient because

```
T obj; // calls empty constructor which might not be implemented!
obj.deserialize(in, length);
```

wrappers/rust/icicle-core/src/merkle/mod.rs
```diff
@@ -81,6 +136,37 @@ impl MerkleProof {
...
 
+    /// Create a new MerkleProof object with specified leaf, root, and path data.
+    pub fn create_with_data<T>(
+        is_pruned: bool,
+        leaf_idx: u64,
+        leaf: &[T],
+        root: &[T],
+        path: &[T],
+    ) -> Result<Self, eIcicleError> {
+        let leaf_bytes = unsafe { slice::from_raw_parts(leaf.as_ptr() as *const u8, leaf.len() * mem::size_of::<T>()) };
```
ChickenLover, line 147: So what if leaf/root or path will be dropped after the creation of the proof object?

nonam3e: is it better to take ownership here?

ChickenLover: Either that, or just copy

ChickenLover: It's a pretty small struct so maybe just copying is best

icicle/include/icicle/fri/fri_proof.h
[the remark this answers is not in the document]
nonam3e: added `memcpy_shift_destination` and `memcpy_shift_source` as protected serialization methods

icicle/include/icicle/errors.h
```diff
@@ -71,6 +71,13 @@ namespace icicle {
     }
   }
 
+#define ICICLE_CHECK_IF_RETURN(api_call)                                                                               \
```
yshekel, line 74: maybe rename to "ICICLE_RETURN_IF_ERR"? 
I know this is the macro in cuda backend too but I think it's less clear

icicle/include/icicle/serialization.h
```diff
@@ -0,0 +1,87 @@
...
+#include <vector>
+#include <cstring>
+
+namespace icicle {
+  class Serializable
+  {
+  public:
+    virtual ~Serializable() = default;
+
+    virtual eIcicleError serialized_size(size_t& size) const { return eIcicleError::API_NOT_IMPLEMENTED; }
```
yshekel, line 14: why not make it pure virtual? with =0? don't you need this implemented always?
Same for following APIs

icicle/include/icicle/serialization.h
```diff
@@ -0,0 +1,87 @@
...
+
+    virtual eIcicleError serialize(std::byte*& buffer, size_t& buffer_length) const
```
yshekel, line 16: why is buffer length passed by ref? isn't in an input?

icicle/include/icicle/serialization.h
```diff
@@ -0,0 +1,87 @@
...
+    {
+      return eIcicleError::API_NOT_IMPLEMENTED;
+    }
+
+    virtual eIcicleError deserialize(std::byte*& buffer, size_t& buffer_length)
```
yshekel, line 21: same regarding buffer_length here

icicle/include/icicle/serialization.h
```diff
@@ -0,0 +1,87 @@
...
+    {
+      std::ofstream file(path, std::ios::binary);
+      if (!file.is_open()) { return eIcicleError::INVALID_ARGUMENT; }
+
+      size_t buffer_length;
+      ICICLE_CHECK_IF_RETURN(serialized_size(buffer_length));
+      std::vector<std::byte> buffer(buffer_length);
+      std::byte* ptr = buffer.data();
+      size_t remaining_length = buffer_length;
+      ICICLE_CHECK_IF_RETURN(serialize(ptr, remaining_length));
```
yshekel, line 36: why do you need this remaining length? you don't use it

icicle/include/icicle/serialization.h
```diff
@@ -0,0 +1,87 @@
...
+      return eIcicleError::SUCCESS;
+    }
+
+    eIcicleError deserialize_from_file(const std::string& path)
+    {
+      std::ifstream file(path, std::ios::binary | std::ios::ate);
+      if (!file.is_open()) { return eIcicleError::INVALID_ARGUMENT; }
+
+      std::streamsize size = file.tellg();
+      if (size <= 0) { return eIcicleError::INVALID_ARGUMENT; }
```
yshekel, line 47: I think you need to check it is exactly the expected size based on serialized_size(). Otherwise it's probably an error

icicle/include/icicle/sumcheck/sumcheck_proof.h
```diff
@@ -17,7 +17,7 @@ namespace icicle {
    */
 
   template <typename S>
-  class SumcheckProof
+  class SumcheckProof : public Serializable
```
yshekel, line 20: I believe using inheritance in this case is a design mistake for several reasons:

(1) Polymorphism is likely unnecessary — the exact type of the proof object to be serialized or deserialized is typically known at the call site. There’s no need to store or manipulate heterogeneous Serializable* pointers in a shared structure.

(2) More critically, inheritance couples the proof object directly with the serialization logic. This makes it difficult to support multiple serialization formats (e.g., binary, JSON, or others). You’d eventually need to modify the proof object’s API just to accommodate different formats — perhaps by adding format enums or conditional logic, which quickly becomes brittle and hard to maintain.

A better approach is to use a Visitor-like design, where an external serializer “visits” the proof object. This pattern keeps serialization logic completely separate from the data structure, allowing for multiple independent serializers and clean, decoupled code.

I suggest making the Serializer a template class, for example something like:

`template<typename T>
struct BinarySerializer {
    static void size();
    static void serialize(const T& obj, std::byte* buffer);
    static void deserialize(const std::byte* buffer , T& obj);
};`

Then you specialize it per proof type T:

`template<>
struct BinarySerializer<SumcheckProof> {
    static size_t size(const SumcheckProof& obj) {
        // TODO implement
    }

    static void serialize(const SumcheckProof& obj, std::byte* buffer) {
        // TODO implement
    }

    static void deserialize(const std::byte* buffer, SumcheckProof& obj) {
              // TODO implement
    }
};`

This approach makes it easy to extend support for new formats (e.g., JsonSerializer<ProofA>) without touching the ProofA definition itself. It adheres to the **Single Responsibility Principle.**
Separation of concerns and decoupling are key principles in software design — they make evolving, debugging, and reusing components significantly easier.

icicle/include/icicle/merkle/merkle_proof.h
```diff
@@ -17,11 +17,35 @@ namespace icicle {
    * It provides functionality to allocate, copy, and access these components, supporting both
    * raw byte manipulation and type-safe access via templates.
    */
-  class MerkleProof
+  class MerkleProof : public Serializable
```
yshekel, line 20: same comment here like SumcheckProof

icicle/src/hash/merkle_c_api.cpp
```diff
@@ -148,4 +165,86 @@ eIcicleError icicle_merkle_tree_verify(icicle::MerkleTree* tree, const icicle::M
...
+    ICICLE_LOG_ERROR << "Cannot deserialize from a null buffer or size is 0.";
+    return eIcicleError::INVALID_POINTER;
+  }
+
+  *proof = new icicle::MerkleProof();
+  return (*proof)->deserialize(buffer, size);
+}
+
+// Serialize the MerkleProof object to a file
+eIcicleError icicle_merkle_proof_serialize_to_file(MerkleProofHandle proof, const char* filename, size_t filename_len)
```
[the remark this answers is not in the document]
yshekel: Do we even need it? I mean given that we serialize to a buffer, users can easily write to file to socket or whatever.   Don't even need FFI actually if you serialize to buffer and let user do whatever he wants with it

wrappers/rust/icicle-core/src/merkle/mod.rs
```diff
@@ -81,6 +136,37 @@ impl MerkleProof {
         }
     }
 
+    /// Create a new MerkleProof object with specified leaf, root, and path data.
+    pub fn create_with_data<T>(
```
yshekel, line 140: When is this required? Do you assume users will use ICICLE verifier without ICICLE prover? if not then when will they use this?

wrappers/rust/icicle-core/src/traits.rs
```diff
@@ -43,3 +43,14 @@ pub trait Arithmetic: Sized + Add<Output = Self> + Sub<Output = Self> + Mul<Outp
 pub trait Handle {
     fn handle(&self) -> *const c_void;
 }
+
+pub trait Serialization
+where
+    Self: Sized,
+{
+    fn get_serialized_size(&self) -> Result<usize, eIcicleError>;
+    fn serialize(&self) -> Result<Vec<u8>, eIcicleError>;
```
yshekel, line 52: (1 )If you build a vector here then do you also need get_serialized_size()?
(2) Maybe better to let the user allocate the memory and give a mutable slice so that he has more control. Maybe we can have both. This can avoid users to copy again and possibly use mapped memory directly (not sure it possible in rust slices, like in C++ but maybe)

yshekel (changes requested):
I left a few comments.

My main issue with this PR is that I believe using inheritance in this case is a design mistake and you should use the visitor pattern instead.

Also Is there a reason not to use the Serialize/Deserialize traits from the serde crate? It seems to be the standard way in Rust to do it

icicle/include/icicle/sumcheck/sumcheck_proof.h
nonam3e: It's a good point, will update the pr

icicle/include/icicle/serialization.h
nonam3e: it's a mutable reference that is used inside to track the remaining length of the proof, maybe it's better to hide it from user, wdyt?

wrappers/rust/icicle-core/src/merkle/mod.rs
nonam3e: yeah, assume users can create it, also it's used it https://github.com/ingonyama-zk/icicle/pull/813

icicle/include/icicle/serialization.h
```diff
@@ -0,0 +1,296 @@
...
+      size_t size;
+      ICICLE_CHECK_IF_RETURN(BinarySerializer<T>::serialized_size(obj, size));
+      if (buffer_length != size) {
+        ICICLE_LOG_ERROR << "Serialization failed: buffer_length != size: " << buffer_length << " != " << size;
+        return eIcicleError::INVALID_ARGUMENT;
+      }
+      return BinarySerializer<T>::pack_and_advance(buffer, buffer_length, obj);
+    }
+
+    static eIcicleError deserialize(std::byte* buffer, size_t buffer_length, T* obj)
```
yshekel, line 62: const std::byte* buffer.
And better make it T& or check that obj!=nullptr. As we discussed if you allocate T internally then the user won't know that he needs to release it and possibly how to release

icicle/include/icicle/serialization.h
```diff
@@ -0,0 +1,296 @@
...
+      std::byte* ptr = buffer.data();
+      ICICLE_CHECK_IF_RETURN(BinarySerializer<T>::serialize(ptr, buffer_length, obj));
+      file.write(reinterpret_cast<const char*>(buffer.data()), buffer_length);
+      return eIcicleError::SUCCESS;
+    }
+
+    static eIcicleError deserialize_from_file(const std::string& path, T* obj)
+    {
+      std::ifstream file(path, std::ios::binary | std::ios::ate);
+      if (!file.is_open()) { return eIcicleError::INVALID_ARGUMENT; }
```
yshekel, line 84: can also print an error explaining what went wrong. Actually in all places you return an error

icicle/include/icicle/serialization.h
```diff
@@ -0,0 +1,296 @@
...
+        size_t round_poly_size;
+        ICICLE_CHECK_IF_RETURN(memcpy_shift_source(&round_poly_size, buffer_length, buffer, sizeof(size_t)));
+
+        size_t byte_size = round_poly_size * sizeof(S);
+        round_polynomials[i].resize(round_poly_size);
+        ICICLE_CHECK_IF_RETURN(memcpy_shift_source(round_polynomials[i].data(), buffer_length, buffer, byte_size));
+      }
+
+      SumcheckProof<S> proof = SumcheckProof<S>(std::move(round_polynomials));
+      *obj = std::move(proof);
```
yshekel, line 141: seems that you can easily take a reference then as I commented above

icicle/include/icicle/serialization.h
```diff
@@ -0,0 +1,296 @@
...
+      std::vector<std::byte> buffer(static_cast<size_t>(size));
+      std::byte* buffer_ptr = buffer.data();
+      if (!file.read(reinterpret_cast<char*>(buffer_ptr), size)) { return eIcicleError::INVALID_ARGUMENT; }
+
+      return BinarySerializer<T>::deserialize(buffer_ptr, static_cast<size_t>(size), obj);
+    }
+  };
+
+  template <typename S>
+  struct BinarySerializer<SumcheckProof<S>> : BinarySerializerBase<SumcheckProof<S>> {
```
yshekel, line 99: move those to separate files, I don't see why have them all in one file here. Users should include what they need and use it.

nonam3e: like sumcheck/sumcheck_proof_serialization.h ?

icicle/include/icicle/serialization.h
```diff
@@ -0,0 +1,296 @@
...
+
+  template <typename T>
+  struct BinarySerializer {
+    static eIcicleError serialized_size(const T& obj, size_t& size);
+    static eIcicleError pack_and_advance(std::byte*& buffer, size_t& buffer_length, const T& obj);
+    static eIcicleError unpack_and_advance(std::byte*& buffer, size_t& buffer_length, T* obj);
+  };
+
+  template <typename T>
+  struct BinarySerializerBase {
```
yshekel, line 50: I see the structure you’re going for, and while it’s not bad, I do find the naming and use of inheritance a bit confusing. I’d like to suggest a slightly different approach that I think improves clarity:

1. Public Interface – Users will interact with a simple templated interface, instantiated with the type T they care about:
```
 template <typename T>
  struct BinarySerializer {
    static size_t size() { ...}
    static eIcicleError serialize(const T& obj, std::byte* buffer, size_t buffer_len) {...}
    static eIcicleError deserialize(const std::byte* buffer, size_t buffer_len, T& obj) {...}
  };
```

2. Internal Implementation – For each supported type (MerkleProof, SumcheckProof, FriProof, etc.), we define a specialized implementation struct, similar to what you already have. Let’s name this BinarySerializerImpl<T>, which holds the actual logic.
3. Bridging the Two – The BinarySerializer<T> interface delegates to BinarySerializerImpl<T> internally, preserving the separation between interface and implementation and avoiding inheritance altogether.


4. the 'BinarySerializer' I defined in (1) is implemented by calling BinarySerializerImpl<T> exactly like you did here.

wrappers/rust/icicle-core/Cargo.toml
```diff
@@ -12,6 +12,9 @@ hex = "0.4"
 once_cell = "1.10.0" # how to have it in dev-deps?
 rayon = "1.8.1"
 rand = "0.8" # TODO move to dev-dependencies
+serde = "1.0.219"
+serde_json = "1.0.140"
```
yshekel, line 16: do we need serde_json?

wrappers/rust/icicle-core/src/sumcheck/mod.rs
```diff
@@ -244,6 +247,38 @@ macro_rules! impl_sumcheck {
...
+            ) -> eIcicleError;
+
+            #[link_name = concat!($field_prefix, "_sumcheck_proof_deserialize")]
+            fn icicle_sumcheck_proof_deserialize(
+                handle: *mut SumcheckProofHandle,
+                buffer: *const u8,
+                size: usize,
+            ) -> eIcicleError;
+
+            #[link_name = concat!($field_prefix, "_sumcheck_proof_serialize_to_file")]
```
yshekel, line 269: I think those two are unused now

wrappers/rust/icicle-core/src/merkle/mod.rs
```diff
@@ -66,6 +112,19 @@ extern "C" {
         out_leaf_idx: *mut u64,
     ) -> *const u8;
     fn icicle_merkle_proof_get_root(proof: MerkleProofHandle, out_size: *mut usize) -> *const u8;
+    fn icicle_merkle_proof_get_serialized_size(proof: MerkleProofHandle, out_size: *mut usize) -> eIcicleError;
+    fn icicle_merkle_proof_serialize(proof: MerkleProofHandle, buffer: *mut u8, size: usize) -> eIcicleError;
+    fn icicle_merkle_proof_deserialize(proof: *mut MerkleProofHandle, buffer: *const u8, size: usize) -> eIcicleError;
+    fn icicle_merkle_proof_serialize_to_file(
```
yshekel, line 118: I think this is unused since you use serde now

wrappers/rust/icicle-core/src/sumcheck/tests.rs
```diff
@@ -282,3 +286,87 @@ where
...
+    );
+    /****** End CPU Proof ******/
+
+    let proof_round_polys =
+        <<SW as Sumcheck>::Proof as SumcheckProofOps<<SW as Sumcheck>::Field>>::get_round_polys(&proof).unwrap();
+
+    let proof_as_sumcheck_proof: <SW as Sumcheck>::Proof = <SW as Sumcheck>::Proof::from(proof_round_polys);
+
+    // === Serialize ===
+    let serialized_proof = to_string(&proof_as_sumcheck_proof).unwrap();
```
yshekel, line 361: to_string/from_string are the serde trait?

icicle/src/sumcheck/sumcheck_c_api.cpp
```diff
@@ -260,5 +261,123 @@ eIcicleError CONCAT_EXPAND(ICICLE_FFI_PREFIX, sumcheck_proof_print)(SumcheckProo
...
+  return BinarySerializer<SumcheckProof<scalar_t>>::deserialize(buffer, size, *sumcheck_proof_handle);
+}
+
+/**
+ * @brief Serializes the given SumcheckProof instance to a file.
+ * @param sumcheck_proof_handle Pointer to the SumcheckProof instance to serialize.
+ * @param filename The name of the file to serialize the SumcheckProof to.
+ * @return eIcicleError indicating the success or failure of the operation.
+ */
+eIcicleError CONCAT_EXPAND(ICICLE_FFI_PREFIX, sumcheck_proof_serialize_to_file)(
```
yshekel, line 344: I suggest removing those if unused in ffi

yshekel (commented):
Looks quite good now. I have a few comments about the C++ structure still but small restructure this time. I think it will simplify and clarify the interface and should be easy to do.

wrappers/rust/icicle-core/src/sumcheck/tests.rs
nonam3e: it's from serde_json, I was thinking about serializing with a generic serializer, but I spent too much time on adding a generic here instead and got nothing

yshekel (approved):

LeonHibnik (approved):

icicle/include/icicle/errors.h
```diff
@@ -71,6 +71,13 @@ namespace icicle {
     }
   }
 
+#define ICICLE_RETURN_IF_ERR(api_call)                                                                                 \
+  do {                                                                                                                 \
+    using namespace icicle;                                                                                            \
```
ChickenLover, line 76: Why do we need using here? Just for eIcicleError name?

wrappers/rust/icicle-runtime/src/errors.rs
```diff
@@ -31,3 +33,23 @@ impl eIcicleError {
         }
     }
 }
+
+impl Display for eIcicleError {
```
ChickenLover, line 37: 👍

wrappers/rust/icicle-core/src/sumcheck/tests.rs
```diff
@@ -282,3 +285,89 @@ where
...
+        combine_func,
+        &config,
+        &sumcheck_config,
+    );
+    /****** End CPU Proof ******/
+
+    let proof_round_polys =
+        <<SW as Sumcheck>::Proof as SumcheckProofOps<<SW as Sumcheck>::Field>>::get_round_polys(&proof).unwrap();
+
+    let proof_as_sumcheck_proof: <SW as Sumcheck>::Proof = <SW as Sumcheck>::Proof::from(proof_round_polys);
```
ChickenLover, line 359: Is it possible to move this to extract the copypaste code here? It has nothing to do with serialization

nonam3e: yeah we can add a separate function here for the test

ChickenLover (approved):
good

PR #863, opened 2025-04-16
empty device vec

emirsoyturk:
## Describe the changes

This PR...

## Linked Issues

Resolves #

### DIFF
--- a/icicle/src/runtime.cpp
+++ b/icicle/src/runtime.cpp
@@ -47,6 +47,11 @@ extern "C" eIcicleError icicle_get_device_count(int& device_count /*OUT*/)
 
 extern "C" eIcicleError icicle_malloc(void** ptr, size_t size)
 {
+  if (size == 0) {
+    *ptr = reinterpret_cast<void*>(0x1);
+    return eIcicleError::SUCCESS;
+  }
+  
   auto err = DeviceAPI::get_thread_local_deviceAPI()->allocate_memory(ptr, size);
   if (err == eIcicleError::SUCCESS) {
     DeviceAPI::get_global_memory_tracker().add_allocation(*ptr, size, DeviceAPI::get_thread_local_device());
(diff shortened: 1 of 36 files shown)


LeonHibnik (approved):

LeonHibnik:
Looks good, Update the PR title and description

icicle/src/runtime.cpp
```diff
@@ -47,6 +47,11 @@ extern "C" eIcicleError icicle_get_device_count(int& device_count /*OUT*/)
 
 extern "C" eIcicleError icicle_malloc(void** ptr, size_t size)
 {
+  if (size == 0) {
+    *ptr = reinterpret_cast<void*>(0x1);
```
yshekel, line 51: Why isn't this an error? I think cuda fails in that case and malloc probably return nullptr.
When does it make sense to allocate 0 bytes and expect icicle_malloc() to succeed?

wrappers/golang/runtime/memory.go
```diff
@@ -6,10 +6,6 @@ import "C"
 import "unsafe"
 
 func Malloc(size uint) (unsafe.Pointer, EIcicleError) {
-	if size == 0 {
-		return nil, AllocationFailed
-	}
-
```
jeremyfelder, line 12: This together with the change made in `icicle_malloc` in cpp, its a breaking change.

icicle/src/runtime.cpp
emirsoyturk: i need that for icicle-halo2. i have a code that initialize DeviceVec for both advice and instance columns (private and public parts of witness) but when a circuit doesn't have public or private input then I need to initialize empty DeviceVec

jeremyfelder: I think for now, it would be better if you include some branching logic in icicle-halo2 for this case and create an empty DeviceVec directly there.

jeremyfelder (changes requested):
We will add this to a v4 roadmap for rust redesign

PR #864, opened 2025-04-17
fix: ensure `s` stays within valid range `[0, p²)`

mdqst:
## Describe the changes

in certain edge cases, `s` could fall outside the expected range.
this change guarantees that `s` is always within `[0, p²)`, which is the correct behavior across all scenarios.


jeremyfelder:
@HadarIngonyama Can you please take a look. I see that you have the change commented out above the current code

jeremyfelder:
@mdqst Thanks for the work here! There are many other changes that can be made to make this script fully representative. For now, it is ok for our needs.

PR #865, opened 2025-04-17
fix: digit comparison for balanced representation

famouswizard:
## Describe the changes

Fixed the issue in the function where the comparison `if digit >= base // 2:` was incorrect.

Now, the value `digit == base // 2` is handled properly, ensuring that the function works as intended for balanced digit representation.


yshekel:
@Tomer-Solberg can you please confirm this is correct? thanks

jeremyfelder:
@famouswizard Thanks for your interest and work on this. The code is correct but the comment is incorrect and should be changed to reflect working with (-b/2,b/2] instead of [-b/2,b/2).

jeremyfelder:
recheck

PR #871, opened 2025-04-22
[FEAT] MatMul API 

lnm-at-work:
## Describe the changes

This PR...

## Linked Issues

Resolves #

### DIFF
--- a/icicle/backend/cpu/src/field/cpu_matrix_ops.cpp
+++ b/icicle/backend/cpu/src/field/cpu_matrix_ops.cpp
@@ -1,0 +1,12 @@
+#include "icicle/backend/vec_ops_backend.h"
+
+template <typename T>   
+static eIcicleError cpu_matrix_mul(
+  const Device& device,
+  const T* mat_a, uint32_t nof_rows_a, uint32_t nof_cols_a,
+  const T* mat_b, uint32_t nof_rows_b, uint32_t nof_cols_b, const VecOpsConfig& config, T* mat_out)
+{
+  return eIcicleError::SUCCESS;
+}   
+
+REGISTER_MATRIX_MUL_BACKEND("CPU", cpu_matrix_mul<scalar_t>);
--- a/icicle/backend/cpu/src/field/cpu_vec_ops.cpp
+++ b/icicle/backend/cpu/src/field/cpu_vec_ops.cpp
@@ -36,6 +36,7 @@ enum VecOperation {
   SLICE,
   REPLACE_ELEMENTS,
   OUT_OF_PLACE_MATRIX_TRANSPOSE,
+  MATRIX_MULTIPLICATION,
 
   NOF_VECTOR_OPERATIONS
 };
@@ -173,6 +174,23 @@ public:
     m_output = mat_out;
     dispatch();
   }
+
+  void send_matrix_multiplication_task(
+    VecOperation operation,
+    const T* mat_a,
+    const uint32_t nof_rows_a,
+    const uint32_t nof_cols_a,
+    const T* mat_b,   
+    const uint32_t nof_rows_b,
+    const uint32_t nof_cols_b,
+    const uint32_t stride,
+    T* mat_out)
+  {
+    m_operation = operation;
+
+    dispatch();
+  }    
+  
 
   // Execute the selected function based on m_operation
   virtual void execute() { (this->*functionPtrs[static_cast<size_t>(m_operation)])(); }
@@ -751,6 +769,24 @@ eIcicleError cpu_matrix_transpose(
 }
 
 REGISTER_MATRIX_TRANSPOSE_BACKEND("CPU", cpu_matrix_transpose<scalar_t>);
+
+
+/*********************************** MATRIX MULTIPLICATION ***********************************/
+
+
+template <typename T>   
+static eIcicleError cpu_matrix_mult(
+  const Device& device,
+  const T* mat_a, uint32_t nof_rows_a, uint32_t nof_cols_a,
+  const T* mat_b, uint32_t nof_rows_b, uint32_t nof_cols_b, const VecOpsConfig& config, T* mat_out)
+{
+    //TODO: Implement matrix multiplication
+  return eIcicleError::SUCCESS;
+}   
+
+REGISTER_MATRIX_MUL_BACKEND("CPU", cpu_matrix_mult<scalar_t>);
+
+
 
 /*********************************** BIT REVERSE ***********************************/
 template <typename T>
--- a/icicle/include/icicle/backend/vec_ops_backend.h
+++ b/icicle/include/icicle/backend/vec_ops_backend.h
@@ -29,13 +29,26 @@ namespace icicle {
     const VecOpsConfig& config,
     scalar_t* output)>;
 
-  using scalarMatrixOpImpl = std::function<eIcicleError(
+  using scalarUnaryMatrixOpImpl = std::function<eIcicleError(
     const Device& device,
     const scalar_t* in,
     uint32_t nof_rows,
     uint32_t nof_cols,
     const VecOpsConfig& config,
     scalar_t* out)>;
+
+  using scalarBinaryMatrixOpImpl = std::function<eIcicleError(
+    const Device& device,
+    const scalar_t* in_a,
+    uint32_t nof_rows_a,
+    uint32_t nof_cols_a,
+    const scalar_t* in_b,
+    uint32_t nof_rows_b,
+    uint32_t nof_cols_b,
+    const VecOpsConfig& config,
+    scalar_t* out)>;
+
+
 
   using scalarBitReverseOpImpl = std::function<eIcicleError(
     const Device& device, const scalar_t* input, uint64_t size, const VecOpsConfig& config, scalar_t* output)>;
@@ -200,7 +213,7 @@ namespace icicle {
     }();                                                                                                               \
   }
 
-  void register_matrix_transpose(const std::string& deviceType, scalarMatrixOpImpl impl);
+  void register_matrix_transpose(const std::string& deviceType, scalarUnaryMatrixOpImpl impl);
 
 #define REGISTER_MATRIX_TRANSPOSE_BACKEND(DEVICE_TYPE, FUNC)                                                           \
   namespace {                                                                                                          \
@@ -209,6 +222,20 @@ namespace icicle {
       return true;                                                                                                     \
     }();                                                                                                               \
   }
+
+
+  void register_matrix_mult(const std::string& deviceType, scalarBinaryMatrixOpImpl impl);
+
+#define REGISTER_MATRIX_MUL_BACKEND(DEVICE_TYPE, FUNC)                                                                 \
+  namespace {                                                                                                          \
+    static bool UNIQUE(_reg_matrix_mul) = []() -> bool {                                                               \
+      register_matrix_mult(DEVICE_TYPE, FUNC);                                                                          \
+      return true;                                                                                                     \
+    }();                                                                                                               \
+  } 
+
+
+
 
   void register_scalar_bit_reverse(const std::string& deviceType, scalarBitReverseOpImpl);
 
--- a/icicle/include/icicle/vec_ops.h
+++ b/icicle/include/icicle/vec_ops.h
@@ -318,6 +318,28 @@ namespace icicle {
   eIcicleError
   matrix_transpose(const T* mat_in, uint32_t nof_rows, uint32_t nof_cols, const VecOpsConfig& config, T* mat_out);
 
+
+  /**
+   * @brief Multiplies two matrices.
+   *
+   * @tparam T Type of the elements in the matrices.
+   * @param mat_a Pointer to the first input matrix.
+   * @param nof_rows_a Number of rows in the first input matrix.
+   * @param nof_cols_a Number of columns in the first input matrix.
+   * @param mat_b Pointer to the second input matrix.
+   * @param nof_rows_b Number of rows in the second input matrix.
+   * @param nof_cols_b Number of columns in the second input matrix.
+   * @param config Configuration for the operation.
+   * @param mat_out Pointer to the output matrix where the results will be stored.
+   * @return eIcicleError Error code indicating success or failure.
+   * @note The input matrices are assumed to be stored in row-major order.
+   *       This function multiplies an matrix A  or a batch of matrices with matrix B.
+   */ 
+  template <typename T>
+  eIcicleError matrix_mult(
+    const T* mat_a, uint32_t nof_rows_a, uint32_t nof_cols_a,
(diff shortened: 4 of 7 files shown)


icicle/backend/cpu/src/field/cpu_vec_ops.cpp
```diff
@@ -174,6 +175,23 @@ class VectorOpTask : public TaskBase
     dispatch();
   }
 
+  void send_matrix_multiplication_task(
```
yshekel, line 178: what is going to be a task? single output element?

yshekel (commented):
Structure looks good for generalized mat-mul
I think not what we need in labrador (its Rq elements there in NTT domain) but we can start from that, @LeonHibnik ?

icicle/tests/test_matrix_api.h
```diff
@@ -0,0 +1,452 @@
...
+      }
+    }
+  }
+
+
+};
+
+
+// Batched matrix multiplication
+TEST_F(MatrixTestBase, MatrixMultiplicationBatched)
```
yshekel, line 110: (1) Please check other tests that loop over all registered devices and test all devices.
(2) below you allocate memory via std::vector but then you set `cfg.is_data_on_device=true` which is wrong. For CPU it works but for GPU will crash. Device memory is allocated with icicle_malloc(). It's true for all tests I think

icicle/tests/test_matrix_api.h
```diff
@@ -0,0 +1,452 @@
...
+
+  // Process each matrix A in the batch with the same B matrix
+  for (size_t i = 0; i < batch_size; i++) {
+    ICICLE_CHECK(matrix_mult(batch_a[i].data(), matrix_size, matrix_size,
+                            single_b.data(), matrix_size, matrix_size,
+                            cfg, icicle_output[i].data()));
+  }
+
+  // Compare results for each matrix in the batch
+  for (size_t i = 0; i < batch_size; i++) {
```
yshekel, line 153: consider using memcmp to avoid this slow loop

icicle/tests/test_matrix_api.h
```diff
@@ -0,0 +1,452 @@
...
+                         valid_matrix.data(), matrix_size, matrix_size,
+                         cfg, result.data());
+  ASSERT_NE(error, eIcicleError::SUCCESS);
+
+  // Test zero columns
+  error = matrix_mult(valid_matrix.data(), matrix_size, 0,
+                     valid_matrix.data(), matrix_size, matrix_size,
+                     cfg, result.data());
+  ASSERT_NE(error, eIcicleError::SUCCESS);
+}
```
yshekel, line 452: no test for rq_matrix_mult?

icicle/backend/cpu/src/field/cpu_matrix_ops.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+
+template <typename T>   
+static eIcicleError cpu_matrix_mult(
+  const Device& device,
+  const T* mat_a, uint32_t nof_rows_a, uint32_t nof_cols_a,
+  const T* mat_b, uint32_t nof_rows_b, uint32_t nof_cols_b, const VecOpsConfig& config, T* mat_out)
+{
+  // Check for null pointers
+  if (mat_a == nullptr || mat_b == nullptr || mat_out == nullptr) {
```
yshekel, line 32: please add ICICLE_LOG_ERROR for each such failure case

icicle/backend/cpu/src/field/cpu_matrix_ops.cpp
```diff
@@ -0,0 +1,207 @@
...
+  
+  // Naive algorithm using taskflow
+
+
+  // Divide the problem among workers
+  const int nof_workers = get_nof_workers(config);
+  const uint32_t rows_per_task = std::max(1U, (nof_rows_a + nof_workers - 1) / nof_workers); // ceil division
+  
+  tf::Taskflow taskflow; // Accumulate tasks
+  tf::Executor executor; // execute all tasks accumulated on multiple threads
```
yshekel, line 58: I think `nof_workers` should be the input of the constructor here

icicle/include/icicle/vec_ops.h
```diff
@@ -318,6 +318,51 @@ namespace icicle {
...
+   * @param nof_rows_b Number of rows in the second input matrix.
+   * @param nof_cols_b Number of columns in the second input matrix.
+   * @param config Configuration for the operation.
+   * @param mat_out Pointer to the output matrix where the results will be stored.
+   * @return eIcicleError Error code indicating success or failure.
+   * @note The input matrices are assumed to be stored in row-major order.
+   *       This function multiplies a matrix A or a batch of matrices with matrix B in RQ mode.
+   */ 
+  template <typename T>
+  eIcicleError rq_matrix_mult(
```
yshekel, line 361: Isn't actually Tq, the NTT domain? rather than Rq I mean?
and BTW I consider adding a simple Rq/Tq structs eventually, so we might reuse matrix_mult<Tq>(...) and matrix_mult<Zq>(...) but anyway internally the implementation is the same.

yshekel (changes requested):
Good work, I left a few comments.

icicle/backend/cpu/src/field/cpu_matrix_ops.cpp
```diff
@@ -0,0 +1,207 @@
...
+  
+  // For each batch
+  for (uint32_t batch_idx = 0; batch_idx < config.batch_size; batch_idx++) {
```
yshekel, line 154: I don't know if batch is really required but ok

icicle/backend/cpu/src/field/cpu_matrix_ops.cpp
```diff
@@ -0,0 +1,207 @@
...
+      taskflow.emplace([=]() {
+        // Compute a block of the output matrix
+        for (uint32_t i = row_start; i < row_end; i++) {
+          for (uint32_t j = 0; j < nof_cols_b; j++) {
+            // Initialize result element to zero
+            T sum = T::zero();
+            
+            // Compute dot product of row i from A and column j from B
+            for (uint32_t k = 0; k < nof_cols_a; k++) {
+              uint64_t a_idx = config.columns_batch ?
```
yshekel, line 179: what do you refer to as column batch? It is originally defined for vecops but here it's not that clear.
the Rq polynomials in B are along the cols right?

icicle/tests/test_matrix_api.h
```diff
@@ -0,0 +1,406 @@
...
+  cfg.is_b_on_device = false;
+  cfg.is_result_on_device = false;
+  cfg.batch_size = batch_size;
+
+  for (const auto& device : s_registered_devices) {
+    ICICLE_CHECK(icicle_set_device(device));
+
+    // Process each matrix A in the batch with the same B matrix
+    for (size_t i = 0; i < batch_size; i++) {
+      ICICLE_CHECK(matrix_mult(
```
yshekel, line 108: Why do you need a batch in this test? The API doesn't support it (and it is not required)

yshekel (commented):
Looks good, great tests but two problems:
(1) missing cpu implementation
(2) missing tests and code for PolyRing type

icicle/backend/cpu/src/field/cpu_matrix_ops.cpp
```diff
@@ -0,0 +1,151 @@
...
+    // Check if inner dimensions match for matrix multiplication
+    if (nof_cols_a != nof_rows_b) { return eIcicleError::INVALID_ARGUMENT; }
+
+    const uint64_t stride = config.columns_batch ? config.batch_size : 1;
+    const uint64_t total_elements_one_mat = static_cast<uint64_t>(nof_rows_a) * nof_cols_b;
+
+    // Naive algorithm using taskflow
+
+    // Divide the problem among workers
+    const int nof_workers = 10; // get_nof_workers(config);
```
LeonHibnik, line 82: Why is the nof_workers hardcoded?

icicle/backend/cpu/src/field/cpu_matrix_ops.cpp
```diff
@@ -0,0 +1,151 @@
...
+      for (uint32_t row_start = 0; row_start < nof_rows_a; row_start += rows_per_task) {
+        uint32_t row_end = std::min(row_start + rows_per_task, nof_rows_a);
+
+        taskflow.emplace([=]() {
+          // Compute a block of the output matrix
+          for (uint32_t i = row_start; i < row_end; i++) {
+            for (uint32_t j = 0; j < nof_cols_b; j++) {
+              // Initialize result element
+              uint64_t a_idx = config.columns_batch ? (i * nof_cols_a + 0) * stride : i * nof_cols_a + 0;
+              uint64_t b_idx = config.columns_batch ? (0 * nof_cols_b + j) * stride : 0 * nof_cols_b + j;
```
LeonHibnik, line 105: Why is `nof_cols_b` here?

icicle/backend/cpu/src/field/cpu_matrix_ops.cpp
```diff
@@ -0,0 +1,151 @@
...
+      // Divide rows among tasks
+      for (uint32_t row_start = 0; row_start < nof_rows_a; row_start += rows_per_task) {
+        uint32_t row_end = std::min(row_start + rows_per_task, nof_rows_a);
+
+        taskflow.emplace([=]() {
+          // Compute a block of the output matrix
+          for (uint32_t i = row_start; i < row_end; i++) {
+            for (uint32_t j = 0; j < nof_cols_b; j++) {
+              // Initialize result element
+              uint64_t a_idx = config.columns_batch ? (i * nof_cols_a + 0) * stride : i * nof_cols_a + 0;
```
LeonHibnik, line 104: Remove the + 0 and add a comment explaining it

LeonHibnik (commented):
Looks good overall, left some comments.
I think we should move the matmul and transpose implementations from vecops to matops, open to a discussion

icicle/backend/cpu/src/field/cpu_matrix_ops.cpp
```diff
@@ -0,0 +1,151 @@
...
+  //     const Zq* b_zq = reinterpret_cast<const Zq*>(&b);
+  //     Zq* c_zq = reinterpret_cast<Zq*>(&c);
+  //     auto config = default_vec_ops_config();
+  //     vector_add(a_zq, b_zq, degree, config, c_zq);
+  //     return c;
+  //   }
+  // #endif
+
+  template <typename T>
+  static eIcicleError cpu_tq_matrix_mult(
```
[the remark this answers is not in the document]
lnm-at-work: I tried to use PolyRing in the backend earlier. Problem: vec_ops.h and poly_ring_vec_ops.cpp are separate. I could either

 * change how rings work to make this type work with the generic methods in vec_ops, 
 * mirror Yuval's solution for the PolyRing vecops. 
 
I went with the latter because there is no operators for PolyRing anyway

icicle/include/icicle/backend/vec_ops_backend.h
```diff
@@ -818,6 +839,27 @@ namespace icicle {
       }();                                                                                                             \
     }
 
+  using polyRingBinaryMatrixOpImpl = std::function<eIcicleError(
```
yshekel, line 842: Since we now have a type for polynomial rings, you should use that instead of a flat (Zq,d).
 This is also true for the backend implementation. See here https://github.com/ingonyama-zk/icicle/blob/a2bc757d441ac461aa61f19c8945cbecfca54af5/icicle/include/icicle/backend/vec_ops_backend.h#L779

icicle/include/icicle/backend/vec_ops_backend.h
```diff
@@ -38,6 +38,17 @@ namespace icicle {
     const VecOpsConfig& config,
     scalar_t* out)>;
 
+  using scalarBinaryMatrixOpImpl = std::function<eIcicleError(
```
yshekel, line 41: what do you mean by binary?

icicle/include/icicle/vec_ops.h
```diff
@@ -318,6 +318,45 @@ namespace icicle {
...
+    uint32_t nof_rows_a,
+    uint32_t nof_cols_a,
+    const T* mat_b,
+    uint32_t nof_rows_b,
+    uint32_t nof_cols_b,
+    const VecOpsConfig& config,
+    T* mat_out);
+
+  template <typename T>
+  eIcicleError poly_ring_matrix_mult(
```
yshekel, line 349: Please remove this api and use the matrix_mult() above only. It should be instantiated for Zq and PolyRing so both can be called with their corresponding implementations

icicle/src/matrix_ops.cpp
```diff
@@ -2,6 +2,73 @@
...
+    uint32_t nof_cols_b,
+    const VecOpsConfig* config,
+    scalar_t* mat_out)
+  {
+    return PolyRingMatrixMultiplicationDispatcher::execute(
+      degree, mat_a, nof_rows_a, nof_cols_a, mat_b, nof_rows_b, nof_cols_b, *config, mat_out);
+  }
+
+  template <>
+  eIcicleError poly_ring_matrix_mult(
```
yshekel, line 56: Following previous comments, here you would have to make it :
`  template <>
  eIcicleError matrix_mult(
    const PolyRing* mat_a,
    uint32_t nof_rows_a,
    uint32_t nof_cols_a,
    const PolyRing* mat_b,
    uint32_t nof_rows_b,
    uint32_t nof_cols_b,
    const VecOpsConfig& config,
    scalar_t* mat_out)`

As we do have such as type now.

icicle/src/rings/polyring_vec_ops.cpp
```diff
@@ -91,4 +91,23 @@ namespace icicle {
     return vector_sum(zq_in, size, sum_cfg, zq_out);
   }
 
+  template <>
+  eIcicleError matrix_mult(
```
yshekel, line 95: This is the correct way, I now see, but please put it in matrix_ops.cpp instead of the 'poly_ring_matrix_mult'

icicle/tests/test_matrix_api.h
```diff
@@ -0,0 +1,475 @@
...
+  const size_t M = 1 << 2;
+
+  auto degree = PolyRing::d;
+  auto direct_input_a = std::vector<PolyRing>(N);
+  auto direct_input_b = std::vector<PolyRing>(M);
+  auto direct_output = std::vector<PolyRing>(N);
+  auto icicle_output = std::vector<PolyRing>(N);
+
+  // Initialize input with N random vectors of M elements
+  direct_input_a.resize(N * M * degree);
```
yshekel, line 448: Why *degree? I mean it's a N*M matrix of PolyRing elements right?

yshekel: Please test (1) mat*vec, (2) vec*mat, (3) non-square cases too

yshekel (approved):

STARRGAXE:
Nice

PR #893, opened 2025-05-04
[FEAT] Download taskflow if not found on system

jeremyfelder:
## Describe the changes

This PR updates the cmake build process for CPU backend to only fetch Taskflow if it cannot be found on the system already.

## Describe the rational

Why is this change necessary?

This reduces build times for CPU backends considerably, especially in the CI

Does it increase performance?

Build performance

### DIFF
--- a/icicle/backend/cpu/CMakeLists.txt
+++ b/icicle/backend/cpu/CMakeLists.txt
@@ -1,33 +1,41 @@
 cmake_minimum_required(VERSION 3.18)
 
-message(STATUS "Fetching Taskflow v3.8.0 (CPU backend)")
+# Find Taskflow package
+message(STATUS "Checking for Taskflow v3.8.0")
+find_package(Taskflow 3.8.0 EXACT QUIET)
 
+if(Taskflow_FOUND)
+  message(STATUS "Found Taskflow v${Taskflow_VERSION}, using existing installation.")
+  # Use icicle_device as interface for TaskFlow headers
+  target_include_directories(icicle_device INTERFACE ${Taskflow_INCLUDE_DIRS})
+else()
+  message(STATUS "Taskflow not found locally. Fetching Taskflow v3.8.0 (CPU backend)")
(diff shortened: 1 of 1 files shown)


icicle/backend/cpu/CMakeLists.txt
```diff
@@ -1,33 +1,41 @@
...
+  set(CMAKE_MESSAGE_LOG_LEVEL "WARNING")
 
-include(FetchContent)
-# Temporarily redefine set message log level to WARNING for fetched content
-set(ORIG_CMAKE_MESSAGE_LOG_LEVEL "${CMAKE_MESSAGE_LOG_LEVEL}")
-set(CMAKE_MESSAGE_LOG_LEVEL "WARNING")
+  FetchContent_Declare(
+    Taskflow
+    GIT_REPOSITORY https://github.com/taskflow/taskflow.git
+    GIT_TAG d8c49c64b4ee5015a3f1c0a42748fa7a2bf5529e # v3.8.0
```
LeonHibnik, line 21: Should we use fetch the code or download the binary?

LeonHibnik (approved):
Looks good

icicle/backend/cpu/CMakeLists.txt
jeremyfelder: I haven't seen any prebuilt binaries that they release, but I might have missed it.

LeonHibnik: I meant the zipped source code

jeremyfelder: I don't think there is a difference. Github zipped source code from a release's asset and the source code form the commit sha are the same as far as I know.

yshekel: I think taskflow is header only and cmake is useful to pull the headers to build dir.

yshekel (approved):

PR #896, opened 2025-05-12
Feat/fri poseidon

ChickenLover:
## Describe the changes

## Describe the rational

Why is this change necessary?

Does it increase performance?

## Backend branches

cuda-backend-branch: feat/fri-poseidon

metal-backend-branch: main

### DIFF
--- a/wrappers/rust/icicle-core/src/fri/tests.rs
+++ b/wrappers/rust/icicle-core/src/fri/tests.rs
@@ -119,8 +119,8 @@ pub fn check_fri_on_device<F: FieldImpl>(
     };
     test_utilities::test_set_main_device();
     check();
-    // test_utilities::test_set_ref_device();
-    // check();
+    test_utilities::test_set_ref_device();
+    check();
 }
 
 pub fn check_fri_proof_serialization<F: FieldImpl, S, D, T>(
--- a/wrappers/rust/icicle-core/src/poseidon2/mod.rs
+++ b/wrappers/rust/icicle-core/src/poseidon2/mod.rs
@@ -8,7 +8,9 @@ use icicle_runtime::errors::eIcicleError;
 /// This allows the implementation of Poseidon2 hashing for various field types that implement `FieldImpl`.
 pub trait Poseidon2Hasher<F: FieldImpl> {
     /// Method to create a new Poseidon2 hasher for a given t (branching factor).
-    fn new(t: u32, domain_tag: Option<&F>) -> Result<Hasher, eIcicleError>;
+    fn new(t: u32, domain_tag: Option<&F>) -> Result<Hasher, eIcicleError> {
+        Self::new_with_input_size(t, domain_tag, 0)
+    }
 
     /// Method to create a new Poseidon2 hasher for a given t, overriding input size.
     fn new_with_input_size(t: u32, domain_tag: Option<&F>, input_size: u32) -> Result<Hasher, eIcicleError>;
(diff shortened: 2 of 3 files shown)


wrappers/rust/icicle-curves/icicle-bn254/src/fri/mod.rs
```diff
@@ -6,9 +6,10 @@ impl_fri!("bn254", bn254_fri, ScalarField, ScalarCfg);
 
 #[cfg(test)]
 mod tests {
-    use icicle_core::impl_fri_tests;
-
-    use crate::curve::ScalarField;
-
-    impl_fri_tests!(bn254_fri_test, ScalarField, ScalarField);
+    mod bn254_fri_test {
```
jeremyfelder, line 9: I don't think this extra mod is needed. Its already separated from others by the directory path

wrappers/rust/icicle-core/src/poseidon2/mod.rs
```diff
@@ -53,20 +64,41 @@ macro_rules! impl_poseidon2 {
...
             // Implement the `Poseidon2Hasher` trait for the given field configuration.
             impl Poseidon2Hasher<$field> for $field_cfg {
                 fn new(t: u32, domain_tag: Option<&$field>) -> Result<Hasher, eIcicleError> {
                     let handle: HasherHandle = unsafe {
-                        create_poseidon2_hasher(t, domain_tag.map_or(std::ptr::null(), |tag| tag as *const $field))
+                        create_poseidon2_hasher(
+                            t,
+                            domain_tag.map_or(std::ptr::null(), |tag| tag as *const $field),
+                            0,
+                        )
```
jeremyfelder, line 78: Can we change this to call the `new_with_input_size` function internally with an input size of 0.

wrappers/rust/icicle-core/src/fri/tests.rs
```diff
@@ -119,8 +119,8 @@ pub fn check_fri_on_device<F: FieldImpl>(
     };
     test_utilities::test_set_main_device();
     check();
-    test_utilities::test_set_ref_device();
-    check();
+    // test_utilities::test_set_ref_device();
+    // check();
```
jeremyfelder, line 123: Why are we removing tests on the ref device?

wrappers/rust/icicle-core/src/fri/mod.rs
```diff
@@ -216,83 +216,118 @@ macro_rules! impl_fri {
...
+        use icicle_hash::keccak::Keccak256;
+        use icicle_runtime::test_utilities;
+        use icicle_runtime::{device::Device, runtime};
+        use serial_test::parallel;
+        use std::sync::Once;
 
-            const MAX_SIZE: u64 = 1 << 10;
-            static INIT: Once = Once::new();
-            const FAST_TWIDDLES_MODE: bool = false;
+        const MAX_SIZE: u64 = 1 << 10;
```
jeremyfelder, line 233: Do we want to keep this size or make is smaller?

icicle/src/hash/poseidon2.cpp
```diff
@@ -7,7 +7,7 @@ namespace icicle {
   ICICLE_DISPATCHER_INST(CreatePoseidon2HasherDispatcher, create_poseidon2, CreatePoseidon2Impl);
 
   template <>
-  Hash create_poseidon2_hash<scalar_t>(unsigned t, const scalar_t* domain_tag)
+  Hash create_poseidon2_hash<scalar_t>(unsigned t, const scalar_t* domain_tag, unsigned input_size)
```
jeremyfelder, line 10: Isn't this a breaking change in cpp API?

wrappers/rust/icicle-core/src/fri/tests.rs
ChickenLover: Oops. I was 100% sure I uncommented it

wrappers/rust/icicle-core/src/fri/mod.rs
ChickenLover: I actually changed it in reduce fri test PR. But 1 << 10 should be also fine

icicle/src/hash/poseidon2.cpp
ChickenLover: yes, but I added the default value to the header. So I think if user includes this function from the header - it should work fine

jeremyfelder: Nice, I missed that. I think it should be ok then.

jeremyfelder (approved):

nonam3e (approved):
👍

PR #897, opened 2025-05-12
G2 multiplier optimization

mickeyasa:
removing karatsube from G2 multiplier and add early return to mul_unsigned


yshekel:
I honestly don't understand this logic so I will it to @HadarIngonyama

icicle/include/icicle/fields/complex_extension.h
```diff
@@ -193,10 +193,11 @@ class ComplexExtensionField
   {
     FWide real_prod = FF::mul_wide(xs.c0, ys.c0);
     FWide imaginary_prod = FF::mul_wide(xs.c1, ys.c1);
-    FWide prod_of_sums = FF::mul_wide(xs.c0 + xs.c1, ys.c0 + ys.c1);
+    FWide ab = FF::mul_wide(xs.c0, ys.c1);
+    FWide ba = FF::mul_wide(xs.c1, ys.c0);
     FWide nonresidue_times_im = mul_by_nonresidue(imaginary_prod);
     nonresidue_times_im = CONFIG::nonresidue_is_negative ? FWide::neg(nonresidue_times_im) : nonresidue_times_im;
-    return Wide{real_prod + nonresidue_times_im, prod_of_sums - real_prod - imaginary_prod};
+    return Wide{real_prod + nonresidue_times_im, ab + ba};
```
jeremyfelder, line 200: Is this faster on all curves that support G2?
How does it affect the supported extension fields?

mickeyasa: excellent point. the bigger the field size and the smaller the hw-multiplier, that optimizationis more affective.
there are some scenario (gpu(32b multiplier)  with small fields) that this code is slower.
@HadarIngonyama  will exclude this functionality for those cases

HadarIngonyama: fixed it

HadarIngonyama (approved):

jeremyfelder (approved):

LeonHibnik (approved):

PR #898, opened 2025-05-13
fix: update build script link for Go wrapper

futreall:
## Describe the changes
The old path (`wrappers/golang/build.sh`) vanished after the repo restructure.  
Switched it to `scripts/release/build_all.sh` so the build steps work again without a 404.


jeremyfelder:
@futreall thanks for the contribution! I believe that script was only updated at v3.7. Can you remove the changes for all of the versions before that.

PR #899, opened 2025-05-14
changed sm preview

Otsar-Raikou:
changed metadata social media previewto new icicle logo

## Describe the changes

## Describe the rational

Why is this change necessary?

Does it increase performance?

## Backend branches

cuda-backend-branch: main

metal-backend-branch: main


jeremyfelder (changes requested):
Two things that need changes here:
1. Docusaurus only has access to the docs folder. You can move the image to `docs/static/img/ICICLENEWLOGO.png`.
2. The README on [line 9](https://github.com/ingonyama-zk/icicle/blob/main/README.md?plain=1#L9) is trying to find the same image so that `src` needs to be updated

jeremyfelder (approved):

LeonHibnik (approved):

PR #907, opened 2025-05-18
Yshekel/jl projection

yshekel:
API for JL-projection, including CPU backend

cuda-backend-branch: main
metal-backend-branch: main

### DIFF
--- a/icicle/backend/cpu/src/ring/cpu_jl_projection.cpp
+++ b/icicle/backend/cpu/src/ring/cpu_jl_projection.cpp
@@ -1,6 +1,7 @@
 #include "icicle/jl_projection.h"
 #include "icicle/backend/vec_ops_backend.h"
 #include "icicle/hash/keccak.h"
+#include "taskflow/taskflow.hpp"
 
 /// @brief CPU backend implementation of the JL projection algorithm.
 /// Projects an input vector of size N (`input_size`) into a lower-dimensional
@@ -15,16 +16,29 @@
 ///     11 →  0
 ///
 
+// extract the number of threads to run from config
+int get_nof_workers(const VecOpsConfig& config); // defined in cpu_vec_ops.cpp
+
 static eIcicleError cpu_jl_projection(
   const Device& device,
   const field_t* input,
   size_t input_size,
   const std::byte* seed,
   size_t seed_len,
-  const VecOpsConfig& cfg,
+  const VecOpsConfig& config,
   field_t* output,
   size_t output_size)
 {
+  if (!input || !output || !seed || input_size == 0 || output_size == 0 || seed_len == 0) {
+    ICICLE_LOG_ERROR << "Invalid argument: null pointer or zero size.";
+    return eIcicleError::INVALID_ARGUMENT;
+  }
+
+  if (config.batch_size != 1) {
+    ICICLE_LOG_ERROR << "Unsupported config: JL projection does not support batch.";
+    return eIcicleError::INVALID_ARGUMENT;
+  }
+
   const size_t num_rows = output_size; // Number of output values
   const size_t num_cols = input_size;  // Length of input vector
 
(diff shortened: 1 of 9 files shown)


icicle/backend/cpu/src/ring/cpu_jl_projection.cpp
```diff
@@ -0,0 +1,105 @@
...
+  const std::byte* seed,
+  size_t seed_len,
+  const VecOpsConfig& config,
+  field_t* output,
+  size_t output_size)
+{
+  if (!input || !output || !seed || input_size == 0 || output_size == 0 || seed_len == 0) {
+    ICICLE_LOG_ERROR << "Invalid argument: null pointer or zero size.";
+    return eIcicleError::INVALID_ARGUMENT;
+  }
```
jeremyfelder, line 35: I'd prefer splitting this into two separate errors: INVALID_ARG and INVALID_POINTER.

icicle/backend/cpu/src/ring/cpu_jl_projection.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+  if (config.batch_size != 1) {
+    ICICLE_LOG_ERROR << "Unsupported config: JL projection does not support batch.";
+    return eIcicleError::INVALID_ARGUMENT;
+  }
+
+  const size_t num_rows = output_size; // Number of output values
```
jeremyfelder, line 42: Isn't this always 256? Only the columns are based on user input?

icicle/backend/cpu/src/ring/cpu_jl_projection.cpp
```diff
@@ -0,0 +1,105 @@
...
+  const size_t num_cols = input_size;  // Length of input vector
+
+  auto keccak512 = Keccak512::create();
+  constexpr uint32_t bits_per_entry = 2;                                        // 2 bits per matrix entry
+  const size_t entries_per_hash = keccak512.output_size() * 8 / bits_per_entry; // 4 × hash bytes
```
jeremyfelder, line 47: Since we are using Keccak512, isn't this always going to be 256?

icicle/backend/cpu/src/ring/cpu_jl_projection.cpp
```diff
@@ -0,0 +1,105 @@
...
+///
+/// Matrix entries are generated with the following 2-bit encoding:
+///     00 →  0
+///     01 → +1
+///     10 → -1
+///     11 →  0
+///
+
+// extract the number of threads to run from config
+int get_nof_workers(const VecOpsConfig& config); // defined in cpu_vec_ops.cpp
```
jeremyfelder, line 20: We had this also in the norm implementation. Why is it needed? Can we use the vecops header file instead?

icicle/tests/test_ring_api.cpp
```diff
@@ -230,4 +231,74 @@ TEST_F(RingTestBase, BalancedDecompositionErrorCases)
...
+
+  auto norm_squared = [&](const std::vector<field_t>& v) -> double {
+    double sum = 0.0;
+    for (const auto& x : v) {
+      const int64_t val = *reinterpret_cast<const int64_t*>(&x);
+      const int64_t centered = (val > q / 2) ? val - q : val; // Convert to signed representative
+      sum += static_cast<double>(centered) * static_cast<double>(centered);
+    }
+    return sum;
+  };
```
jeremyfelder, line 267: Norm is also using similar test logic, whichever of the two is merged first the other should update their tests.

icicle/tests/test_ring_api.cpp
```diff
@@ -230,4 +231,74 @@ TEST_F(RingTestBase, BalancedDecompositionErrorCases)
...
+TEST_F(RingTestBase, JLProjectionTest)
+{
+  static_assert(field_t::TLC == 2, "Decomposition assumes q ~64b");
+  constexpr auto q_storage = field_t::get_modulus();
+  const int64_t q = *(int64_t*)&q_storage; // Note this is valid since TLC == 2
+  ICICLE_ASSERT(q > 0) << "Expecting at least one slack bit to use int64 arithmetic";
+
+  const size_t N = 1 << 16;       // Input vector size
+  const size_t output_size = 256; // JL projected size
+  const int max_trials = 10; // JL projection output is bound by sqrt(128)*norm(input) with probability 0.5. Therefore
```
jeremyfelder, line 245: Would a user need to redo the JL projection if this bound isn't held? If so, why don't we perform this check and redo internally?

icicle/backend/cpu/src/ring/cpu_jl_projection.cpp
yshekel: yes, we can. It's not exposed but I think Emir is doing it in this PR and I will use that

icicle/backend/cpu/src/ring/cpu_jl_projection.cpp
yshekel: yes it is. Just wanted the computation to be clear when reading

icicle/tests/test_ring_api.cpp
yshekel: The user will have to retry with another seed until he finds a seed that works. He will later use this seed again for another computation.
I can do it internally but I don't want to because it means defining the logic to choose another seed. Typically the seed is based on fiat shamir and I want to have the user control it entirely and not be responsible for this.
Also internally I would have to take the output bound as parameter and it makes the api not project() but more than that. We can have such an API if we wanted but in addition, not part of this one.

icicle/backend/cpu/src/ring/cpu_jl_projection.cpp
yshekel: fixed

jeremyfelder (approved):

emirsoyturk (approved):

PR #908, opened 2025-05-19
Labrador example

yshekel:
Labrador C++ example

cuda-backend-branch: main
metal-backend-branch: main


examples/c++/labrador/example.cpp
```diff
@@ -0,0 +1,967 @@
...
+  for (size_t i = 0; i < r; i++) {
+    // Create seed for P_i matrix (same as in step 12)
+    std::vector<std::byte> base_jl_seed(seed1);
+    base_jl_seed.push_back(std::byte(JL_i));
+    base_jl_seed.push_back(std::byte(i));
+
+    for (size_t j = 0; j < JL_out; j++) {
+      // Get row j of matrix P_i
+      std::vector<Zq> R_ij(n * d);
+      ICICLE_CHECK(get_jl_matrix_row(
```
yshekel, line 202: @goforashutosh we can have the api return a vector of Rq polynomials, conjugated. What do you think?
Also maybe we can generate a bunch of rows, possibly the entire matrix at once, or a subset of rows as a vector of conjugated Rq polyonmials. Makes sense? instead of this API and the conjugate one.

PR #916, opened 2025-05-24
docs: Complete Go example for multi-device setup

CreeptoGengar:
## Describe the changes

This PR completes the Go section in the multi-device documentation by removing the TODO placeholder and adding a comprehensive code example. The addition includes proper import statements, device creation for both CUDA and CPU, error handling, and follows the same pattern established by C++ and Rust examples.

## Describe the rational

Why is this change necessary?
The TODO in the Go section left Go developers without guidance on how to implement multi-device functionality using ICICLE's Go bindings. This creates documentation gaps and reduces developer experience for Go users.

Does it increase performance?
No, this is a documentation improvement that enables developers to properly utilize existing multi-device performance features in Go applications.

## Backend branches

cuda-backend-branch: main

metal-backend-branch: main

### DIFF
--- a/docs/docs/start/architecture/multi-device.md
+++ b/docs/docs/start/architecture/multi-device.md
@@ -48,7 +48,25 @@ let device = icicle_runtime::Device::new("CUDA", 0 /* =device_id*/);
 icicle_runtime::set_device(&device).unwrap();
 ```
 
-### Go (TODO)
+### Go
+
+```go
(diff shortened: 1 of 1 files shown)


jeremyfelder (changes requested):
Hey @CreeptoGengar! Thanks for taking the time to contribute. The example you added in the Go docs for multi-device use is not correct. Using `runtime.SetDevice` alone is problematic due to the way Go handles concurrency via goroutines. In order to properly use multiple devices, `runtime.RunOnDevice` should be called.

PR #921, opened 2025-05-29
[FEAT] add method to get challenges from sumcheck's Fiat-Shamir

idanfr-ingo:
In sumcheck - add get_challenge_vector method to let users get the challenges generated by the transcript

cuda-backend-branch: get_sumchecks_challanges

metal-backend-branch: main

### DIFF
--- a/icicle/include/icicle/backend/sumcheck_backend.h
+++ b/icicle/include/icicle/backend/sumcheck_backend.h
@@ -51,6 +51,7 @@ namespace icicle {
       const CombineFunction<F>& combine_function,
       SumcheckTranscriptConfig<F>&& transcript_config,
       const SumcheckConfig& sumcheck_config,
+      F* challenge_vector /*out*/,
       SumcheckProof<F>& sumcheck_proof /*out*/) = 0;
   };
 
--- a/icicle/include/icicle/sumcheck/sumcheck.h
+++ b/icicle/include/icicle/sumcheck/sumcheck.h
@@ -110,7 +110,7 @@ namespace icicle {
         mle_polynomials, mle_polynomial_size, claimed_sum, combine_function, std::move(transcript_config),
         sumcheck_config, const_cast<F*>(m_challenge_vector.data()), sumcheck_proof);
     }
-    F get_challenge_vector(int round_idx) { return m_challenge_vector[round_idx]; }
+    std::vector<F> get_challenge_vector() { return m_challenge_vector; }
 
     /**
      * @brief Verify an element against the Sumcheck round polynomial.
--- a/icicle/include/icicle/sumcheck/sumcheck.h
+++ b/icicle/include/icicle/sumcheck/sumcheck.h
@@ -102,10 +102,15 @@ namespace icicle {
                          << ", but " << total_nof_vars << " were given";
         return eIcicleError::INVALID_ARGUMENT;
       }
+
+      // Initialize challenge vector with the correct size based on MLE polynomial size
+      m_challenge_vector.resize(std::log2(mle_polynomial_size));
+
       return m_backend->get_proof(
         mle_polynomials, mle_polynomial_size, claimed_sum, combine_function, std::move(transcript_config),
-        sumcheck_config, sumcheck_proof);
+        sumcheck_config, const_cast<F*>(m_challenge_vector.data()), sumcheck_proof);
     }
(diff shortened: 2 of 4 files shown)


icicle/tests/test_field_api.cpp
```diff
@@ -881,6 +881,91 @@ TYPED_TEST(FieldTest, SumcheckSingleInputProgram)
...
+      verifier_sumcheck.verify(sumcheck_proof, claimed_sum, std::move(verifier_transcript_config), verification_pass));
+
+    ASSERT_EQ(true, verification_pass);
+
+    ASSERT_EQ(prover_sumcheck.get_challenge_vector(0), TypeParam::zero());
+
+    for (int i = 0; i < std::log2(mle_poly_size); i++) {
+      TypeParam challenge = prover_sumcheck.get_challenge_vector(i);
+      ICICLE_LOG_INFO << "challenge_vector[" << i << "] = " << challenge;
+    }
```
jeremyfelder, line 959: How does this test that the API functions as intended?

jeremyfelder (approved):

icicle/tests/test_field_api.cpp
idanfr-ingo: At line 954 it checks that the first challenge is 0, here it checks that it can access all challenges.

icicle/include/icicle/sumcheck/sumcheck.h
```diff
@@ -102,10 +102,15 @@ namespace icicle {
...
+
+      // Initialize challenge vector with the correct size based on MLE polynomial size
+      m_challenge_vector.resize(std::log2(mle_polynomial_size));
+
       return m_backend->get_proof(
         mle_polynomials, mle_polynomial_size, claimed_sum, combine_function, std::move(transcript_config),
-        sumcheck_config, sumcheck_proof);
+        sumcheck_config, const_cast<F*>(m_challenge_vector.data()), sumcheck_proof);
     }
+    F get_challenge_vector(int round_idx) { return m_challenge_vector[round_idx]; }
```
jeremyfelder, line 113: The function name is misleading since the function only returns a single challenge and not the entire vector of challenges.
Does it make sense to provide two functions, 1 for obtaining the entire vector of challenges and 1 for obtaining a single challenge? (alternatively, if the number of rounds is usually very small then why not always return the entire vector)

idanfr-ingo: changed method to return entire vector

jeremyfelder (approved):

icicle/include/icicle/sumcheck/sumcheck.h
```diff
@@ -102,10 +102,15 @@ namespace icicle {
...
+
+      // Initialize challenge vector with the correct size based on MLE polynomial size
+      m_challenge_vector.resize(std::log2(mle_polynomial_size));
+
       return m_backend->get_proof(
         mle_polynomials, mle_polynomial_size, claimed_sum, combine_function, std::move(transcript_config),
-        sumcheck_config, sumcheck_proof);
+        sumcheck_config, const_cast<F*>(m_challenge_vector.data()), sumcheck_proof);
     }
+    std::vector<F> get_challenge_vector() { return m_challenge_vector; }
```
yshekel, line 113: Note that you copy here rather than return a reference

icicle/include/icicle/sumcheck/sumcheck.h
```diff
@@ -158,6 +163,7 @@ namespace icicle {
   private:
     std::shared_ptr<SumcheckBackend<F>>
       m_backend; ///< Shared pointer to the backend responsible for Sumcheck operations.
+    mutable std::vector<F> m_challenge_vector;
```
yshekel, line 166: Why is this vector here and not inside the backend? everything is there and this breaks the design that relies on a backend to store and compute

icicle/include/icicle/backend/sumcheck_backend.h
```diff
@@ -51,6 +51,7 @@ namespace icicle {
       const CombineFunction<F>& combine_function,
       SumcheckTranscriptConfig<F>&& transcript_config,
       const SumcheckConfig& sumcheck_config,
+      F* challenge_vector /*out*/,
```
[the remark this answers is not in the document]
yshekel: Alternatively, at least pass the size of the vector, but even better, the vector should be stored in the backend and have it return a reference when needed

PR #928, opened 2025-06-03
fri final polynomial degree check

krakhit:
## Describe the changes

This PR adds a fix to the FRI verifier, following a vulnerability discovered in [Plonky3 verifier](https://github.com/Plonky3/Plonky3/commit/e784f44924e12a5a6799f3b03c18d1fa6b1a111e) 
 

1. FRI Verifier: Early Degree/Length Check: The FRI verifier now checks that the final polynomial’s length is exactly stopping_degree + 1 at the very start of verification, before any transcript, Merkle, or collinearity checks. 
2. Improved Error Message: The error message now includes both the actual and expected degree, and explicitly notes the security context (potential degree attack).
3. Test: Malicious Final Polynomial:  Added a test that constructs a valid FRI proof, then appends one or more extra coefficients to the final polynomial. The test verifies that the verifier rejects such proofs with eIcicleError::INVALID_ARGUMENT and valid == false.
4. Test Utility: Accessor for Query Proofs
Added a get_query_proofs() accessor to FriProof to facilitate test construction of modified proofs. This is only used in tests and does not affect normal operation. 

## Describe the rational

Why is this change necessary?

**Security Impact**

- The FRI verifier is now robust against degree attacks, matching the soundness guarantees of Plonky3.
- Any proof with a final polynomial of higher degree than allowed is immediately and reliably rejected.

Does it increase performance? No

## Backend branches

Not relevant, only for the FRI verifier which is CPU backend

Sample test output
```
Note: Google Test filter = *FriRejectsHighDegreeFinalPoly*
[==========] Running 2 tests from 2 test suites.
[----------] Global test environment set-up.
[----------] 1 test from FieldTest/0, where TypeParam = Field<babybear::fp_config>
[INFO] Failed to load backend from any known directory.
[INFO] Main-device=UNKNOWN, Reference-device=CPU
[ RUN      ] FieldTest/0.FriRejectsHighDegreeFinalPoly
[INFO] Seed for test is: 1748955555
[ERROR] Security check failed: FRI final polynomial degree is 17, expected degree is 15. This may indicate a malicious proof (degree attack).
```

``

### DIFF
--- a/icicle/backend/cpu/include/cpu_sumcheck.h
+++ b/icicle/backend/cpu/include/cpu_sumcheck.h
@@ -24,6 +24,7 @@ namespace icicle {
       const CombineFunction<F>& combine_function,
       SumcheckTranscriptConfig<F>&& transcript_config,
       const SumcheckConfig& sumcheck_config,
+      F* challenge_vector /*out*/,
       SumcheckProof<F>& sumcheck_proof /*out*/) override
     {
       if (sumcheck_config.use_extension_field) {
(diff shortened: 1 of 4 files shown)


jeremyfelder (approved):

krakhit (commented):
> In other cases where checks fail, we are returning `eIcicleError::Success` BUT with `valid = false`. I suggest we stay with this style for now 

Sure, makes sense. done

LeonHibnik (approved):

PR #945, opened 2025-06-15
Decompose Polynomial-rings

yshekel:
- Fix the decomposition rust API to execute on rings, not only prime-fields
- Implement the decomposition for polynomial rings instead of integer-rings (can do both in v4 but currently they do not match in design due to the Config type)
- update decomposition layout to be digit-major

TODO Yuval: PR to main when have polyring

cuda-backend-branch: yshekel/decompose_digit_major_layout
metal-backend-branch: main


ChickenLover (approved):

LeonHibnik (approved):
Nice!

PR #950, opened 2025-06-18
[FEATURE] Pseudorandom field element sampling from seed

ChickenLover:
## Describe the changes

This PR adds random sampling for ring and PolyRing

## Backend branches

cuda-backend-branch: feat/random-sampling

metal-backend-branch: main

### DIFF
--- a/icicle/tests/test_ring_api.cpp
+++ b/icicle/tests/test_ring_api.cpp
@@ -1097,7 +1097,7 @@ TEST_F(RingTestBase, RandomSampling)
   }
 
   auto test_random_sampling = [&](bool fast_mode) {
-    const int N = 10;
+    const int N = 15;
     for (int i = 0; i < N; ++i) {
       for (size_t device_index = 0; device_index < s_registered_devices.size(); ++device_index) {
         ICICLE_CHECK(icicle_set_device(s_registered_devices[device_index]));


icicle/cmake/target_editor.cmake
```diff
@@ -25,6 +25,7 @@ function(handle_ring TARGET)
     src/vec_ops.cpp
     src/rings/rns_vec_ops.cpp
     src/rings/polyring_vec_ops.cpp
+    src/rings/vec_ops.cpp
```
yshekel, line 28: why is named vec_ops and not random_sampling?

icicle/include/icicle/backend/vec_ops_backend.h
```diff
@@ -521,7 +521,46 @@ namespace icicle {
...
+  #define REGISTER_RING_ZQ_RANDOM_SAMPLING_BACKEND(DEVICE_TYPE, FUNC)                                                  \
+    namespace {                                                                                                        \
+      static bool UNIQUE(_reg_ring_zq_random_sampling) = []() -> bool {                                                \
+        register_ring_zq_random_sampling(DEVICE_TYPE, FUNC);                                                           \
+        return true;                                                                                                   \
+      }();                                                                                                             \
+    }
+
+  // for Rq type
+  using ringRqRandomSamplingImpl = std::function<eIcicleError(
```
yshekel, line 545: Why do you need two dispatchers and two backend APIs? Can't you implement the Rq case on top of the Zq in the frontend? That way backends don't need to implement the same thing twice.
BTW I don't think we need this API for Rq.

icicle/include/icicle/math/modular_arithmetic.h
```diff
@@ -478,6 +478,15 @@ class ModArith
     return reduce(Wide{res}); // finally, use barret reduction
   }
 
+  static constexpr HOST_DEVICE_INLINE Derived reduce_from_bytes(const std::byte* in)
```
yshekel, line 481: (1) How is that different from 'from(const std::byte* in, unsigned nof_bytes)' just above this one?
(2) I think raw pointer without size is a mistake that can cause segfaults
(3) replace 4 with sizeof(limb)

icicle/include/icicle/rings/random_sampling.h
```diff
@@ -0,0 +1,13 @@
...
+
+#include "icicle/errors.h"
+#include "icicle/vec_ops.h"
+
+namespace icicle {
+  // This should be the same for all the devices to get a deterministic result
+  const uint64_t RANDOM_SAMPLING_FAST_MODE_NUMBER_OF_TASKS = 256;
+
+  template <typename T>
+  eIcicleError ring_random_sampling(
```
yshekel, line 11: I suggest renaming to 'random_sampling' as it doesn't have to be for rings only.
Also not sure why it is under icicle/rings since it is not consistent with other APIs.

icicle/src/rings/vec_ops.cpp
```diff
@@ -0,0 +1,44 @@
+#include "icicle/backend/vec_ops_backend.h"
```
yshekel, line 1: I think this file is named incorrectly. Should be random_sampling.cpp

icicle/backend/cpu/src/ring/cpu_random_sampling.cpp
```diff
@@ -0,0 +1,148 @@
...
+    taskflow.clear();
+  }
+}
+
+void slow_mode_random_sampling(
+  uint64_t size, const std::byte* seed, uint64_t seed_len, const VecOpsConfig& cfg, field_t* output)
+{
+  // Use keccak to get deterministic uniform distribution
+  auto keccak512 = Keccak512::create();
+  const size_t element_size = field_t::TLC * 4;
```
yshekel, line 62: same about 4

icicle/tests/test_ring_api.cpp
```diff
@@ -1084,3 +1087,34 @@ TEST_F(RingTestBase, NegacyclicNTT)
...
+
+  ICICLE_CHECK(icicle_set_device(IcicleTestBase::main_device()));
+  const int N = 10;
+
+  auto test_random_sampling = [&](bool fast_mode, auto& a, auto& b) {
+    for (int i = 0; i < N; ++i) {
+      ICICLE_CHECK(ring_random_sampling(size, fast_mode, seed.data(), seed_len, VecOpsConfig{}, a.data()));
+      ICICLE_CHECK(ring_random_sampling(size, fast_mode, seed.data(), seed_len, VecOpsConfig{}, b.data()));
+      for (size_t i = 0; i < size; ++i) {
+        ASSERT_EQ(a[i], b[i]);
```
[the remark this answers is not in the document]
ChickenLover: Actually I was testing consistency between backends but it got accidently removed when I change the tests for rq

icicle/backend/cpu/src/ring/cpu_random_sampling.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+void fast_mode_random_sampling(
+  uint64_t size, const std::byte* seed, uint64_t seed_len, const VecOpsConfig& cfg, field_t* output)
+{
+  // Use keccak to get deterministic uniform distribution
+  auto keccak512 = Keccak512::create();
+  const size_t element_size = field_t::TLC * 4;
+  const size_t elements_per_hash = std::max(keccak512.output_size() / element_size, size_t(1));
+  // To support elements that are larger than 32 bytes
+  const size_t hashes_per_element = std::max(element_size / keccak512.output_size(), size_t(1));
```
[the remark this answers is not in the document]
ChickenLover: yes, good catch!

LeonHibnik (commented):
Looks good, left a question

LeonHibnik (approved):

LeonHibnik (approved):

PR #952, opened 2025-06-18
[FEAT] Add flattening utilities for polynomial slices into scalar slices

yshekel:
Some APIs operate on scalar elements rather than polynomials. To use them with polynomial data, we need to reinterpret polynomial slices as flat slices of their scalar coefficients.

This PR introduces four functions to safely flatten HostSlice and DeviceSlice—both mutable and immutable—into corresponding slices of scalar elements.

cuda-backend-branch: main
metal-backend-branch: main

### DIFF
--- a/wrappers/rust/icicle-core/src/tests.rs
+++ b/wrappers/rust/icicle-core/src/tests.rs
@@ -1,11 +1,11 @@
-use crate::polynomial_ring::{flatten_polynomials_host_slice, PolynomialRing};
+use crate::polynomial_ring::{flatten_device_polynomials, flatten_host_polynomials, PolynomialRing};
 use crate::{
     curve::{Affine, Curve, Projective},
     field::Field,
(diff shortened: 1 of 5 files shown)


wrappers/rust/icicle-core/src/tests.rs
```diff
@@ -209,10 +209,78 @@ where
     assert_eq!(poly.values(), input.as_slice());
 }
 
-pub fn check_vector_alloc<P: PolynomialRing>()
-where
-    P: Clone,
-{
+pub fn check_vector_alloc<P: PolynomialRing>() {
```
ChickenLover, line 212: What is this for?

yshekel: I just removed the trait bound but actually it's a silly test. I will remove it

yshekel:
> Nice, we should refactor the icicle-core/src/tests file in the future @yshekel

agreed

PR #953, opened 2025-06-18
HostDevice memory API tweaks

vhnatyk:
## 📌 Describe the Changes

This early draft PR introduces the following improvements:

1. **NTT Example Consistency**  
   Aligns the `ntt` example file behavior with the README instructions to ensure it behaves exactly as documented.

2. **Simplified Rust API for `HostOrDeviceSlice`**  
   Makes the Rust `HostOrDeviceSlice` API less verbose by avoiding advanced Rust syntax like `&mut scalars[..]`.  
   Uses simpler method calls like `.func_name()` instead.

3. **Improved Allocation Syntax**  
   Replaces verbose allocation like `DeviceVec::device_malloc().unwrap()` with cleaner `DeviceVec::malloc()`.  
   This makes it visually clear that the memory is allocated on the **device**, not the host — avoiding repetition like `device_malloc()` on a `DeviceVec`.

---

### Why is this change necessary?

- The current memory management API in Rust has long been a candidate for simplification and clarity improvements.
- This PR cleans up commonly used code blocks.
- Helps avoid cases [like this](https://github.com/ingonyama-zk/icicle/compare/develop/vhnat/rust-mem-api?expand=1#:~:text=let%20merkle_tree_min_layer_to_store%20%3D%200%3B), where memory is allocated and data copied to device, but actually is **unused**.

### Does this improve performance?

- No — this is purely for **ergonomic** and **readability** purposes.

### Scope

- Rust wrapper 

---

## 🤔 Suggestions / Open Questions

- **Rename `HostOrDeviceSlice` to `IcicleSlice`?**  
  A shorter, branded, and more intuitive name might improve ergonomics.

- **Confusing Behavior in Default Config**  
  The default config assumes all inputs are on the **host**.  
  However, when supplying a `DeviceVec` to a Rust function:
  - A **failsafe mechanism** overrides the config to: _inputs on device, outputs on device_.
  - As a result, outputs are unexpectedly placed on the device, even if the config specified host outputs.

  This implicit override is **confusing** and potentially error-prone.

### Possible Alternatives

- Respect user-defined config and enforce checks on input location.
- Emit warnings if the actual data doesn't match expectations.
- Provide **explicit wrappers** or helper methods:
  - Accepting `DeviceVec` or `HostSlice` directly.
  - Avoiding the use of dynamic `HostOrDeviceSlice` trait objects.

---

## 🔀 Backend Branches

- **CUDA Backend Branch:** `main` at [`2f6bbdb7`](https://github.com/ingonyama-zk/icicle-cuda-backend/commit/2f6bbdb781e0e7c6ebf0198fd25b845108707ef5)
- **Metal Backend Branch:** _Not specified_

### DIFF
--- a/examples/rust/ntt/src/main.rs
+++ b/examples/rust/ntt/src/main.rs
@@ -94,10 +94,10 @@ fn main() {
     println!("Executing bn254 NTT on device...");
     let start = Instant::now();
     ntt::ntt(
-        scalars.into_icicle_slice(),
+        scalars.into_slice(),
         ntt::NTTDir::kForward,
         &cfg,
-        ntt_results.into_icicle_slice_mut(),
+        ntt_results.into_slice_mut(),
     )
     .unwrap();
     println!(
@@ -112,7 +112,7 @@ fn main() {
     assert_ne!(host_bn254_results, scalars); // check that the results are not the same as the inputs
 
     ntt::ntt_inplace(
-        scalars.into_icicle_slice_mut(),
+        scalars.into_slice_mut(),
         ntt::NTTDir::kForward,
         &cfg,
     )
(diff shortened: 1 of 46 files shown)


examples/rust/ntt/src/main.rs
```diff
@@ -32,6 +39,86 @@ fn try_load_and_set_backend_device(args: &Args) {
     icicle_runtime::set_device(&device).unwrap();
 }
 
+// Add a small extension trait that exposes a convenient one-liner for moving data back to the host.
+trait DeviceVecExt<T: PrimeField + Copy> {
```
yshekel, line 43: Those traits should be in memory.rs right?

vhnatyk:
closed in favor of #1014

PR #954, opened 2025-06-18
[FEAT] Add reinterpretation utilities and polynomial vector operations

yshekel:
- Introduce `reinterpret_slice` and `reinterpret_slice_mut` for view casting `HostOrDeviceSlice` between any two types
- Implement `flatten_polyring_slice` and `flatten_polyring_slice_mut` to abstract flattening of `HostOrDeviceSlice<P>` to `HostOrDeviceSlice<P::Base>`
- Add generic `polyvec_add`, `polyvec_sub`, `polyvec_mul`, `polyvec_mul_by_scalar`, and `polyvec_sum_reduce` over `HostOrDeviceSlice<PolynomialRing>`
- Use flattening logic to delegate polynomial operations to existing base field vector ops

cuda-backend-branch: main
metal-backend-branch: main


wrappers/rust/icicle-core/src/vec_ops/poly_vecops.rs
```diff
@@ -0,0 +1,266 @@
+use super::{add_scalars, mul_scalars, scalar_mul, sub_scalars, sum_scalars, VecOps, VecOpsConfig};
+use crate::{polynomial_ring::PolynomialRing, traits::FieldImpl};
+use icicle_runtime::{
+    errors::eIcicleError,
+    memory::{DeviceSlice, HostOrDeviceSlice, HostSlice},
+    stream::IcicleStream,
+};
+
+// TODO Yuval: should we expose those flatten methods too?
```
yshekel, line 9: TODO Yuval: generalize to two types as a cast<From, To> and move to runtime crate.
output size is computed as sizeof(From)/Sizeof(To)*len
output size can be smaller than input size generally

wrappers/rust/icicle-runtime/src/memory.rs
```diff
@@ -564,3 +564,174 @@ impl<T> DerefMut for DeviceVec<T> {
...
+        Host(&'a mut HostSlice<T>),
+        Device(&'a mut DeviceSlice<T>),
+    }
+
+    /// SAFETY: Caller must ensure layout compatibility between `From` and `To`.
+    fn compute_output_len<From, To>(len: usize) -> usize {
+        let from_size = size_of::<From>();
+        let to_size = size_of::<To>();
+
+        assert!(from_size > 0 && to_size > 0, "Invalid zero-sized types");
```
nonam3e, line 587: maybe use Result here instead of assertion?

yshekel:
> everything is fine, but maybe it is better to remove panics where it's possible

Great observation. I refactored and return Results now so errors returned to used instead of panic

nonam3e (approved):

LeonHibnik (approved):
Nice
TODO add more error types in eIcicileError

PR #955, opened 2025-06-18
determine the scalar size according to configuration if provided

mickeyasa:
scalar size can arrive manually by config

### DIFF
--- a/icicle/backend/cpu/src/curve/cpu_msm.hpp
+++ b/icicle/backend/cpu/src/curve/cpu_msm.hpp
@@ -197,7 +197,7 @@ private:
   void calc_optimal_parameters()
   {
     m_precompute_factor = m_config.precompute_factor;
-    m_scalar_size = scalar_t::NBITS; // TBD handle this config.bitsize != 0 ? config.bitsize : scalar_t::NBITS;
+    m_scalar_size = m_config.bitsize != 0 ? m_config.bitsize : scalar_t::NBITS;
     m_nof_workers = get_optimal_nof_workers(m_config, m_msm_size, m_scalar_size, m_precompute_factor);
 
     // phase 1 properties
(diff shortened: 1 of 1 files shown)


icicle/include/icicle/math/modular_arithmetic.h
```diff
@@ -282,6 +282,14 @@ class ModArith
     rv &= ((1 << digit_width) - 1);
     return rv;
   }
+  HOST_DEVICE_INLINE uint32_t get_scalar_bits(unsigned lsb_idx, unsigned width) const
+  {
+    const uint32_t limb_lsb_idx = lsb_idx / 32;
+    const uint32_t shift_bits = lsb_idx % 32;
+    const uint64_t mask = (1 << width) - 1;
```
yshekel, line 289: (1) Note that I think '1' is an integer, thus 32b, not 64b. This would be '1ULL' or something like that.
(2) what happens when width is too large? maybe assert if you cannot return an error.
(3) this method name is not very clear to me. I mean without the implementation it's not that clear what it does. Maybe rename to 'slice'?

icicle/include/icicle/math/modular_arithmetic.h
```diff
@@ -282,6 +282,14 @@ class ModArith
     rv &= ((1 << digit_width) - 1);
     return rv;
   }
+  HOST_DEVICE_INLINE uint32_t get_scalar_bits(unsigned lsb_idx, unsigned width) const
+  {
+    const uint32_t limb_lsb_idx = lsb_idx / 32;
```
yshekel, line 287: better use 8*sizeof(limb) or something like this

icicle/tests/test_curve_api.cpp
```diff
@@ -78,6 +78,52 @@ class CurveApiTest : public IcicleTestBase
     }
   }
 
+  template <typename A, typename P>
+  void MSM_bitsize()
+  {
+    const int logn = 12;
+    const int batch = 1;
+    const int N = (1 << logn); // - rand_uint_32b(0, 5 * logn); // make it not always power of two
```
yshekel, line 86: remove the comment ?

icicle/backend/cpu/src/curve/cpu_msm.hpp
```diff
@@ -279,16 +285,16 @@ class Msm
         for (int bm_i = 0; bm_i < m_nof_buckets_module; bm_i++) {
           // Avoid seg fault in case precompute_factor*c exceeds the scalar width by comparing index with num additions
           if (m_nof_buckets_module * j + bm_i >= num_bms_before_precompute) { break; }
-
-          uint32_t curr_coeff = scalar.get_scalar_digit(m_nof_buckets_module * j + bm_i, m_c) + carry;
+          const unsigned coeff_width = std::min(m_c, m_scalar_size - scalar_offset);
+          const uint32_t curr_coeff = scalar.get_scalar_bits(scalar_offset, coeff_width) + carry;
+          uint32_t get_scalar_digit_curr_coeff = scalar.get_scalar_digit(m_nof_buckets_module * j + bm_i, m_c) + carry;
```
Koren-Brand, line 290: get_scalar_digit_curr_coeff is unused

icicle/include/icicle/math/modular_arithmetic.h
```diff
@@ -282,6 +282,15 @@ class ModArith
     rv &= ((1 << digit_width) - 1);
     return rv;
   }
+  HOST_DEVICE_INLINE uint32_t get_scalar_bits(const unsigned lsb_idx, const unsigned width) const
+  {
+    ICICLE_ASSERT(width <= 8 * sizeof(*(limbs_storage.limbs))) << "get_scalar_bits::width should be < 32";
```
Koren-Brand, line 287: shouldn't the assertion output the actual size compared? 8 * sizeof(*(limbs_storage.limbs))?

PR #963, opened 2025-06-23
[CI] Move merkle test to hash crate

jeremyfelder:
## Describe the changes

This PR fixes the babybear tests by moving the merkle tree test to the hash crate as it is not specific to babybear


wrappers/rust/icicle-hash/src/tests.rs
```diff
@@ -455,4 +458,73 @@ mod tests {
...
+    fn test_merkle_tree_segfault() {
+        initialize();
+        test_utilities::test_set_main_device();
+
+        let n = 18;
+        let field_element_size = 4;
+        for elements_per_leaf in [1, 5] {
+            // calc tree parameters
+            let leaf_size = field_element_size * elements_per_leaf;
+            let mut test_vec = vec![0 as u8; 1 << n];
```
yshekel, line 472: Note that the original test had 2^n elements of size 4B but here it's 1B per element. Maybe just replace u8 with u32 or keep it as u8 but make the size `(1<<n)*field_element_size`

yshekel (approved):

PR #966, opened 2025-06-25
Feat/challenge space sampling

ChickenLover:
## Describe the changes

## Describe the rational

Why is this change necessary?

Does it increase performance?

## Backend branches

cuda-backend-branch: feat/challenge-space-sampling

metal-backend-branch: main


icicle/include/icicle/random_sampling.h
```diff
@@ -22,4 +22,36 @@ namespace icicle {
   template <typename T>
   eIcicleError random_sampling(
     size_t size, bool fast_mode, const std::byte* seed, size_t seed_len, const VecOpsConfig& config, T* output);
+
+  /**
+   * @brief Sample Labrador Rq challenge polynomials from challenge space C.
```
yshekel, line 27: AFAIK it's not limited to Labrador and the set of {-1,-2,0,1,2} generalizes to other protocols too. @goforashutosh please correct me if wrong

icicle/include/icicle/random_sampling.h
```diff
@@ -22,4 +22,36 @@ namespace icicle {
...
+   * 2. Randomly flips the signs of the coefficients
+   * 3. Permutes the coefficients randomly
+   *
+   * Note: This function does not ensure norm constraints (e.g., τ, T) hold
+   *
+   * @tparam T The element type for the polynomial coefficients.
+   * @param seed Pointer to the seed buffer for deterministic sampling.
+   * @param seed_len Length of the seed buffer in bytes.
+   * @param size Number of polynomials to sample.
+   * @param ones Number of 1s coefficients in each polynomial.
```
yshekel, line 41: maybe mention it's including -1 values too and same of the 2s

icicle/backend/cpu/src/ring/cpu_random_sampling.cpp
```diff
@@ -131,4 +131,168 @@ eIcicleError cpu_random_sampling(
...
+  static const field_t neg_one = field_t::neg(field_t::one());
+
+  const size_t nof_workers = std::min((size_t)get_nof_workers(cfg), total_hashes);
+  const size_t size_per_worker = (size + nof_workers - 1) / nof_workers;
+  const size_t hashes_per_worker = (size_per_worker + polynomials_per_hash - 1) / polynomials_per_hash;
+
+  tf::Taskflow taskflow;
+  tf::Executor executor(nof_workers);
+
+  for (size_t w = 0; w < nof_workers; ++w) {
```
yshekel, line 248: @goforashutosh please review the implementation in terms of its behavior

yshekel (commented):
Looks good to me. Left small comments

icicle/backend/cpu/src/ring/cpu_random_sampling.cpp
```diff
@@ -131,4 +131,168 @@ eIcicleError cpu_random_sampling(
...
+  int j = size_a;
+  int n = size_a + size_b;
+  while (true) {
+    if (random_bit_iterator.next_bit()) {
+      if (j == n) { break; }
+      std::swap(array[i], array[j]);
+      ++j;
+    } else {
+      if (i == j) { break; }
+      ++i;
```
goforashutosh, line 194: I think i increments outside the if-else

icicle/backend/cpu/src/ring/cpu_random_sampling.cpp
```diff
@@ -131,4 +131,168 @@ eIcicleError cpu_random_sampling(
...
+    }
+  }
+  for (i = i ? i : 1; i < n; i++) {
```
goforashutosh, line 197: isn't i already at the correct position?

icicle/backend/cpu/src/ring/cpu_random_sampling.cpp
```diff
@@ -131,4 +131,168 @@ eIcicleError cpu_random_sampling(
...
+      if (limb_idx >= 8) { lfsr_state = lfsr64(lfsr_state); }
+    }
+    return bit;
+  }
+};
+
+// Cross shuffles two adjacent ranges of an array as described in the paper
+// https://arxiv.org/pdf/1508.03167
+template <typename T>
+void merge_shuffle(
```
goforashutosh, line 181: I asked Claude Sonnet to review this against Algo 3 of the paper. It gave this. I think these are all valid: 
Looking at your implementation of the merge shuffle algorithm, I can see several issues when comparing it to the provided pseudocode:

## Issues with the Implementation:

### 1. **Inverted Coin Flip Logic**
Your implementation has the coin flip logic backwards:
- **Pseudocode**: When `FLIP() = 0`, take from first sub-array (increment `i`)
- **Your code**: When `next_bit()` returns `true` (1), you take from second sub-array

### 2. **Random Integer Generation Bug**
The Fisher-Yates shuffle part has a critical bug:
```cpp
for (int b = 0; b < index_bits; b++) {
    m |= random_bit_iterator.next_bit();
    m <<= 1;  // This shifts AFTER setting the bit, losing the last bit
}
```

### 3. **Loop Starting Condition**
```cpp
for (i = i ? i : 1; i < n; i++)
```
This should simply be `for (; i < n; i++)` since `i` is already at the correct position.

### 4. **Incorrect Random Range**
The pseudocode generates `m ∈ {s, ..., i}`, but your modulo operation `m % i` gives range `{0, ..., i-1}`.

## Corrected Implementation:

```cpp
template <typename T>
void merge_shuffle(
    T* array, uint32_t size_a, uint32_t size_b, uint32_t index_bits, RandomBitIterator& random_bit_iterator)
{
    int i = 0;
    int j = size_a;
    int n = size_a + size_b;
    
    // Main merge loop with coin flips
    while (true) {
        if (!random_bit_iterator.next_bit()) {  // Flip = 0: take from first sub-array
            if (i == j) { break; }
            ++i;
        } else {  // Flip = 1: take from second sub-array
            if (j == n) { break; }
            std::swap(array[i], array[j]);
            ++j;
        }
    }
    
    // Fisher-Yates shuffle for remaining elements
    for (; i < n; i++) {
        uint32_t m = 0;
        for (int b = 0; b < index_bits; b++) {
            m <<= 1;  // Shift first
            m |= random_bit_iterator.next_bit();  // Then set bit
        }
        m = m % (i + 1);  // Range {0, ..., i}
        std::swap(array[i], array[m]);
    }
}
```

## Key Corrections:

1. **Fixed coin flip**: `!random_bit_iterator.next_bit()` for correct logic
2. **Fixed bit generation**: Shift before setting the bit
3. **Simplified loop start**: Use existing `i` value
4. **Fixed random range**: `m % (i + 1)` for correct Fisher-Yates range

The algorithm's purpose is to randomly merge two sub-arrays while maintaining a uniform distribution over all possible shuffles, which requires both the correct coin flip logic and proper Fisher-Yates implementation.

icicle/backend/cpu/src/ring/cpu_random_sampling.cpp
goforashutosh: I can't verify the hashing stuff but the rest looks good

icicle/include/icicle/random_sampling.h
goforashutosh: I think the description in brief is good. You can't fix a number of -1s here since the signs are randomly flipped

icicle/backend/cpu/src/ring/cpu_random_sampling.cpp
ChickenLover: It is, but there are rare cases where i will be 0. If I is zero we can't sample a number in range `[0, i)`

icicle/backend/cpu/src/ring/cpu_random_sampling.cpp
ChickenLover: 1. Doesn't make a difference internally but can impact if we want to compare against another merge shuffle implementation which we don't, but wouldn't hurt to change
2. Agree
3. Not true
4. It doesn't make sense to swap A[i] with A[i]

icicle/backend/cpu/src/ring/cpu_random_sampling.cpp
ChickenLover: oh, actually with your previous comment it can't be 0. So yeah

icicle/backend/cpu/src/ring/cpu_random_sampling.cpp
goforashutosh: Agree on 1,2,3.
For 4- it's a probability thing. There needs to be some probability that the element at i doesn't change its position. For example if you permute (1,2) with prob 1/2 it should stay (1,2).

Actually this leads to a more fundamental problem- to sample m you need to use rejection sampling to make sure that it is uniformly random in {0,..., i}. Right now there will be some bias when you simply take modulo. I suggest you look at MergeShuffle code: https://github.com/axel-bacher/mergeshuffle

I don't know how much work this is but you could also try to add a test for merge_shuffle where you shuffle (1,2)/ (1,2,3) a million times and look at the statistics-- they should be close to 1/2 for each permutation of (1,2) and 1/6 for each permutation of (1,2,3)

icicle/include/icicle/operator_norm.h
```diff
@@ -0,0 +1,208 @@
...
+      if (i < j) {
+        ComplexFixed tmp = a[i];
+        a[i] = a[j];
+        a[j] = tmp;
+      }
+    }
+
+    for (size_t len = 2, stage = 0; len <= n; ++stage, len <<= 1) {
+      ComplexFixed wlen = wlen_table[stage];
+      if (inverse) wlen.im = FixedPoint::from_float(-wlen.im.to_float());
```
yshekel, line 170: why need to convert here?

icicle/include/icicle/random_sampling.h
```diff
@@ -22,4 +22,36 @@ namespace icicle {
...
+   * 2. Randomly flips the signs of the coefficients
+   * 3. Permutes the coefficients randomly
+   *
+   * @tparam T The element type for the polynomial coefficients.
+   * @param seed Pointer to the seed buffer for deterministic sampling.
+   * @param seed_len Length of the seed buffer in bytes.
+   * @param size Number of polynomials to sample.
+   * @param ones Number of 1s coefficients in each polynomial.
+   * @param twos Number of 2s coefficients in each polynomial.
+   * @param norm Only sample polynomials for which operator norm is less than or equal to this value. If set to zero,
```
yshekel, line 41: I suggest to rephrase with "rejection sampling"

icicle/include/icicle/random_sampling.h
```diff
@@ -22,4 +22,36 @@ namespace icicle {
...
+   *
+   */
+  template <typename T>
+  eIcicleError sample_challenge_space_polynomials(
+    const std::byte* seed,
+    size_t seed_len,
+    size_t size,
+    uint32_t ones,
+    uint32_t twos,
+    int64_t norm,
```
yshekel, line 54: Better use u64

wrappers/rust/icicle-core/src/random_sampling/mod.rs
```diff
@@ -104,3 +104,132 @@ macro_rules! impl_random_sampling_tests {
...
+/// - `ones`: Number of ±1 coefficients in each polynomial.
+/// - `twos`: Number of ±2 coefficients in each polynomial.
+/// - `norm`: Only sample polynomials for which operator norm is less than or equal to this value. If set to zero,
+///   no norm constraint is applied.
+/// - `output`: Output buffer to store the sampled Rq polynomials. Should be of size `cfg.batch_size`.
+///
+/// # Returns
+/// - `Ok(())` on success, or an error code on failure.
+pub fn challenge_space_polynomials_sampling<T>(
+    size: usize,
```
yshekel, line 142: you don't need size since it is the output.len(). Also please remove from the comment above

icicle/include/icicle/operator_norm.h
ChickenLover: multiple consecutive AI edits without proper review... I need to start reviewing my own code before pushing..

icicle/include/icicle/operator_norm.h
```diff
@@ -0,0 +1,208 @@
...
+        x = ComplexFixed{x.re * scale, x.im * scale};
+    }
+  }
+
+  /// @brief Compute the operator norm of a single polynomial: max |FFT(ψᵢ·aᵢ)| over ℂ
+  inline int64_t operator_norm(const Poly& a)
+  {
+    CPoly complex_a;
+    for (size_t i = 0; i < N; ++i)
+      complex_a[i] = ComplexFixed{FixedPoint::from_float(static_cast<float>(a[i])), FixedPoint::from_float(0.0f)} * twist[i];
```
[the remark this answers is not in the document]
ChickenLover: same as above...

icicle/tests/test_ring_api.cpp
```diff
@@ -1286,3 +1288,85 @@ TEST_F(RingTestBase, RandomSampling)
...
+      const auto& poly = outputs[device_index][i];
+      std::unordered_map<field_t, size_t> coeff_counts;
+      for (int j = 0; j < Rq::d; ++j) {
+        coeff_counts[poly.values[j]]++;
+      }
+      ASSERT_EQ(coeff_counts[field_t::one()] + coeff_counts[neg_one], ones);
+      ASSERT_EQ(coeff_counts[two] + coeff_counts[neg_two], twos);
+      ASSERT_EQ(coeff_counts[field_t::zero()], Rq::d - ones - twos);
+    }
+  }
```
yshekel, line 1335: I think you should maybe compare that devices are consistent with each other using the same seed.
Also consider testing they polynomials are indeed low norm by computing the norm directly, ideally in double precision

icicle/tests/test_ring_api.cpp
```diff
@@ -1286,3 +1288,85 @@ TEST_F(RingTestBase, RandomSampling)
...
+TEST_F(RingTestBase, ComplexFFT_Simple)
+{
+  using namespace opnorm;
+  Poly poly{};
+  for (size_t i = 0; i < N; ++i)
+    poly[i] = i * 100;
+
+  uint64_t opnorm = operator_norm(poly);
+  printf("opnorm: %lu\n", opnorm);
+  ASSERT_LT(opnorm, 152851);             // Python computed 152849.98 but losing precision with f32
```
yshekel, line 1348: let's also assert the low bound since we use fixed precision. Same for the other tests

icicle/backend/cpu/src/ring/cpu_random_sampling.cpp
```diff
@@ -131,4 +132,181 @@ eIcicleError cpu_random_sampling(
...
+          output_polynomial->values, ones, twos, std::ceil(std::log2(ones + twos)), random_bit_iterator);
+        // Do merge shuffle of shuffled 1s and 2s and zeroes
+        merge_shuffle(
+          output_polynomial->values, ones + twos, Rq::d - ones - twos, std::ceil(std::log2(Rq::d)),
+          random_bit_iterator);
+
+        if (norm) {
+          opnorm::Poly poly{};
+          for (int i = 0; i < Rq::d; ++i) {
+            poly[i] = balanced_table.at(output_polynomial->values[i]);
```
yshekel, line 296: Could you shuffle the complex values directly? they are also 64b each, instead of shuffling field elements and then map them and then convert to complex. Or maybe shuffle the real part and then convert to complex?

yshekel (commented):
Left few last comments. We need to make sure OpNorm is close enough to the double precision reference implementation.

icicle/tests/test_ring_api.cpp
```diff
@@ -1286,3 +1288,85 @@ TEST_F(RingTestBase, RandomSampling)
...
+  for (size_t device_index = 0; device_index < s_registered_devices.size(); ++device_index) {
+    outputs[device_index] = std::vector<Rq>(size);
+  }
+
+  size_t ones = 31;
+  size_t twos = 10;
+  int64_t norm = 15;
+
+  const int N = 5;
+  for (int i = 0; i < N; ++i) {
```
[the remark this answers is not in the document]
ChickenLover: Initially it was used to catch some rare segfaults/race conditions, but I think it's safe to remove now as the test size is big enough to catch them anyway

icicle/backend/cpu/src/ring/cpu_random_sampling.cpp
ChickenLover: yeah, but then we need to copy it back into normal non-complex form anyway, so does it actually saves anything?

yshekel (approved):
Approved. Please see my last comment. I think this PR may not compile for other fields that don't have this CONFIG:modulus_u64

PR #971, opened 2025-06-26
[FIX] Use proper feature value propagation for CMAKE_DEPENDENT_OPTION

jeremyfelder:
## Describe the changes

This PR fixes feature variables value propagation when using CMAKE_DEPENDENT_OPTION. The regular `option` command uses CACHE variables when in project mode whereas CMAKE_DEPENDENT_OPTION uses local variables. In order to properly update the values so that they are passed to other functions and cmake files we must set in the PARENT_SCOPE


icicle/cmake/curve.cmake
```diff
@@ -53,8 +50,11 @@ function(setup_curve_target CURVE CURVE_INDEX FEATURES_STRING)
   handle_curve(icicle_curve) # basic curve and field methods, including vec ops
   # Handle features
   handle_msm(icicle_curve "${FEATURES_LIST}")
+  set(MSM ${MSM} PARENT_SCOPE)
```
yshekel, line 53: can you move this into the handlers()?

yshekel (commented):
If values are removed from the cache, will backends still detect the options set here?
Also, in this PR there’s no CI triggered—are CMake file modifications not enough to trigger a CI build?

PR #972, opened 2025-06-26
[FIX] Enable and fix examples

jeremyfelder:
## Describe the changes

This PR:

- Enables the examples workflow
- Fixes golang examples
- Run examples before release via release workflow
- Fixes issue with Golang pqc bindings not linking to libicicle_pqc when only ml-kem is imported
- Fixes the CI build for PQC by enabling PQC feature directly

This PR is built on top of #976


docs/docusaurus.config.ts
```diff
@@ -13,7 +13,7 @@ const ingoPreset = {
   docs: {
     showLastUpdateAuthor: false,
     showLastUpdateTime: true,
-    includeCurrentVersion: false,
+    includeCurrentVersion: process.env.NODE_ENV !== 'production',
```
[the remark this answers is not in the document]
jeremyfelder: It is set automatically by docusaurus in the `build` command: https://github.com/facebook/docusaurus/blob/main/packages/docusaurus/src/commands/build/build.ts#L27

examples/c++/examples_utils.h
```diff
@@ -19,6 +19,8 @@ void try_load_and_set_backend_device(int argc = 0, char** argv = nullptr)
     return;
   }
 
+  if (strcmp(selected_device, "CPU") == 0) { return; }
```
yshekel, line 22: I think better use std::string operator==

jeremyfelder: When running the examples with `-d CPU`, `icicle_load_backend_from_env_or_default` still runs and can load an old and not compatible backend that is in the default location which causes it to crash.

This was just a quick way to fix the examples. I don't think using CPU as a device and loading a backend are mutually exclusive.

LeonHibnik (approved):

yshekel (approved):

PR #976, opened 2025-06-29
[FIX] Replace `dependent_option` with regular `option`

jeremyfelder:
## Describe the changes

This PR fixes the cmake build scripts feature options by reverting a previous change of using dependent_option and using the regular option command

## Describe the rational

cmake_dependent_option is not the right command for our use case (a different default value for the option based on the environment). It is meant to be used to conditionally provide an option. When the condition is met, the option is available and follows the standard `option` command, however, when it is **not** met, it creates a local variable with a default value. 

This presents an issue when functions and other scopes are relying on setting and reading the CACHE values. Due to this, we were forced to use local variables and propagate changes to parent scopes. Furthermore, the local variable shadows any cache entry made by the user which caused the loss of disabling features for developers (outside of the CI env).

Reverting to use the regular `option` in both environments allows us to keep the flexibility for developers and CI and removes the complex shadowing from local variables

### DIFF
--- a/icicle/CMakeLists.txt
+++ b/icicle/CMakeLists.txt
@@ -31,32 +31,23 @@ option(METAL_BACKEND "Branch/commit to pull for METAL backend or `local` if unde
 option(VULKAN_BACKEND "Branch/commit to pull for VULKAN backend or `local` if under icicle/backend/vulkan. Default=OFF" OFF)
 
 option(DISABLE_ALL_FEATURES "Disables all features" OFF)
-# features that some fields/curves have and some don't.
-if(NOT DISABLE_ALL_FEATURES)
-  option(NTT "Build NTT" ON)
-  option(ECNTT "Build ECNTT" ON)
-  option(MSM "Build MSM" ON)
-  option(G2 "Build G2 MSM" ON)
-  option(EXT_FIELD "Build extension field" ON)
-  option(HASH "Build hashes and tree builders" ON)
-  option(POSEIDON "Build poseidon hash" ON)
-  option(POSEIDON2 "Build poseidon2 hash" ON)
-  option(SUMCHECK "Build sumcheck" ON)
-  option(FRI "Build fri" ON)
+if(DISABLE_ALL_FEATURES)
+  set(ENABLE_ALL_FEATURES OFF)
 else()
-  option(NTT "Build NTT" OFF)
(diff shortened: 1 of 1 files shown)


jeremyfelder:
>Also, in this PR there’s no CI triggered—are CMake file modifications not enough to trigger a CI build?
_Originally posted by @yshekel in https://github.com/ingonyama-zk/icicle/pull/971#pullrequestreview-2961970224_

So far our CI has been a single build+test with no standalone build checks. I'm not sure changing the cmake files should trigger tests. We should make sure that the libs build correctly on cmake changes though

icicle/CMakeLists.txt
```diff
@@ -31,18 +30,33 @@ option(CUDA_BACKEND "Branch/commit to pull for CUDA backend or `local` if under
...
-CMAKE_DEPENDENT_OPTION(MSM "Build MSM" OFF "INGO_CI_ENV" ON)
-CMAKE_DEPENDENT_OPTION(G2 "Build G2 MSM" OFF "INGO_CI_ENV" ON)
-CMAKE_DEPENDENT_OPTION(EXT_FIELD "Build extension field" OFF "INGO_CI_ENV" ON)
-CMAKE_DEPENDENT_OPTION(HASH "Build hashes and tree builders" OFF "INGO_CI_ENV" ON)
-CMAKE_DEPENDENT_OPTION(POSEIDON "Build poseidon hash" OFF "INGO_CI_ENV" ON)
-CMAKE_DEPENDENT_OPTION(POSEIDON2 "Build poseidon2 hash" OFF "INGO_CI_ENV" ON)
-CMAKE_DEPENDENT_OPTION(SUMCHECK "Build sumcheck" OFF "INGO_CI_ENV" ON)
-CMAKE_DEPENDENT_OPTION(FRI "Build fri" OFF "INGO_CI_ENV" ON)
-CMAKE_DEPENDENT_OPTION(PQC "Build pqc" OFF "INGO_CI_ENV" CUDA_PQC_BACKEND) # Post-Quantum Cryptography. TODO: default to ON? do we care about CPU too?
+if(NOT DISABLE_ALL_FEATURES)
```
yshekel, line 35: can you make it like
`set(DEFAULT_OPTION_VALUE ...`

And remove this if-else? like the PQC option

LeonHibnik (approved):

yshekel (approved):

PR #977, opened 2025-06-29
[FIX] Matmul rust wrappers identify device memory and assert sizes

yshekel:
cuda-backend-branch: main
metal-backend-branch: main


lnm-at-work (approved):
looks good

PR #984, opened 2025-07-01
[FEATURE] Add matrix transpose support to Rust matrix_ops wrapper

yshekel:


lnm-at-work:
> OPEN: should we decouple matrix ops to seperate traits? Or we want to have them as a bundle such that we cannot have matmul without transpose etc.?

I lean towards keeping them together. I can't think of a structure for which we don't manage to provide both and making users require MatOps on their generics vs. making them require each individual op is cleaner.

nonam3e (commented):
👍

PR #985, opened 2025-07-01
Remove Program from VecOps in Rust

vhnatyk:
## Describe the changes

This PR refactors Rust VecOps trait to remove Program execution

## Describe the rational

Why is this change necessary? 
 - execute_program should be method of Program trait, not VecOps

Does it increase performance? 
 - No

## Backend branches

cuda-backend-branch: main

metal-backend-branch: main


yshekel (approved):
Maybe PR to main instead?

PR #988, opened 2025-07-02
wrapped error type

nonam3e:
This PR adds rust wrapped errors

Haven't changed polynomial api a lot, it seems it won't be that simple to use as now if Results are added there

cuda-backend-branch: main

metal-backend-branch: main

### DIFF
--- a/wrappers/rust/icicle-runtime/src/memory.rs
+++ b/wrappers/rust/icicle-runtime/src/memory.rs
@@ -1,4 +1,4 @@
-use crate::errors::eIcicleError;
+use crate::errors::{eIcicleError, IcicleError};
 use crate::runtime;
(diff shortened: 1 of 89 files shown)


wrappers/rust/icicle-runtime/src/memory.rs
```diff
@@ -286,7 +278,10 @@ impl<T> DeviceSlice<T> {
         }
         if !self.is_on_active_device() {
             eprintln!("not allocated on an active device");
```
yshekel, line 280: can remove the prints?

nonam3e: removed

nonam3e:
moved to https://github.com/ingonyama-zk/icicle/pull/993

PR #990, opened 2025-07-02
V4 Rust wrapper redesign

LeonHibnik:
## Describe the changes
This PR introduces a first-class, idiomatic Rust interface to the Icicle CUDA kernels.
* Trait Hierarchy
* Better readability
* Removed Field and Curve Config 
* Refactored VecOps

## Backend branches

cuda-backend-branch: V4-new

metal-backend-branch: V4-new


nonam3e (commented):
looks good, left some small comments

wrappers/rust/icicle-fields/icicle-koalabear/src/ntt/mod.rs
```diff
@@ -21,4 +14,16 @@ pub(crate) mod tests {
...
+    use super::ExtensionField;
+    use icicle_core::{
+        ntt::{initialize_domain, ntt_inplace, release_domain, NTTConfig, NTTDir, NTTInitDomainConfig},
+        traits::{GenerateRandom, PrimeField},
+    };
+    use icicle_runtime::memory::HostSlice;
+    use risc0_core::field::{
+        baby_bear::{Elem, ExtElem},
+        Elem as FieldElem, RootsOfUnity,
+    };
```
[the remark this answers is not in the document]
LeonHibnik: Removed (also removed in #861 :) )

docs/docs/start/integration-&-support/migrate_from_v3.md
```diff
@@ -66,6 +66,52 @@ FF result = FF::from(5)
 
 ## Migration Guide for Rust
 
+### Arithmetic API
+
+ICICLE v4 implements field arithmetic operations through the `Arithmetic` trait and standard Rust operators. This makes the code more idiomatic and easier to read.
+
+#### Field Arithmetic Operations
```
yshekel, line 73: doesn't it work in v3 already? with the Arithmetic trait and Add/Sub etc.

wrappers/rust/icicle-core/src/balanced_decomposition/mod.rs
```diff
@@ -38,42 +26,33 @@ pub trait BalancedDecomposition<T> {
...
-pub fn count_digits<T>(base: u32) -> u32
-where
-    T: BalancedDecomposition<T>,
-{
+pub fn count_digits<T: PrimeField + BalancedDecomposition>(base: u32) -> u32 {
     T::count_digits(base)
 }
 
-pub fn decompose<T>(
+pub fn decompose<T: PrimeField + BalancedDecomposition>(
```
yshekel, line 41: Here as well it should not be PrimeField.

wrappers/rust/icicle-core/src/balanced_decomposition/tests.rs
```diff
@@ -1,18 +1,17 @@
-use crate::polynomial_ring::PolynomialRing;
-use crate::{balanced_decomposition, traits::GenerateRandom, vec_ops::VecOpsConfig};
+use crate::{balanced_decomposition, field::PrimeField, traits::GenerateRandom, vec_ops::VecOpsConfig};
 
 use icicle_runtime::memory::{DeviceVec, HostSlice};
 
 pub fn check_balanced_decomposition<F>()
 where
-    F: PolynomialRing + balanced_decomposition::BalancedDecomposition<F> + GenerateRandom<F>,
+    F: PrimeField + balanced_decomposition::BalancedDecomposition + GenerateRandom,
```
yshekel, line 7: why make polyring test fields? this doesn't make any sense to me

wrappers/rust/icicle-core/src/ecntt/mod.rs
```diff
@@ -3,64 +3,44 @@ use icicle_runtime::{errors::eIcicleError, memory::HostOrDeviceSlice};
...
 
 #[doc(hidden)]
 pub mod tests;
 
 #[doc(hidden)]
-pub trait ECNTT<C: Curve>: ECNTTUnchecked<Projective<C>, C::ScalarField> {}
-
-#[doc(hidden)]
-pub trait ECNTTUnchecked<T, F: FieldImpl> {
+pub trait ECNTT<T>: PrimeField {
```
yshekel, line 14: This is also strange. Why do you use this style rather than ECNTT<Scalar, Point> and bound them

wrappers/rust/icicle-core/src/rns/mod.rs
```diff
@@ -1,10 +1,10 @@
-use crate::{traits::FieldImpl, vec_ops::VecOpsConfig};
+use crate::{field::PrimeField, vec_ops::VecOpsConfig};
 use icicle_runtime::{eIcicleError, memory::HostOrDeviceSlice};
 
 pub mod tests;
 
 /// Trait for RNS conversions (`Zq <--> ZqRns`)
-pub trait RnsConversion<Zq: FieldImpl, ZqRns: FieldImpl> {
+pub trait RnsConversion<Zq: PrimeField, ZqRns: PrimeField> {
```
yshekel, line 7: Since we  renamed to PrimeField, I think it's a bad name. Zq here is not a prime field, and the rns too is not a prime field. Maybe rename? Or have an integerRing trait such that `PrimeField : IntergerRing`

wrappers/rust/icicle-curves/icicle-bn254/src/ecntt/mod.rs
```diff
@@ -1,8 +1,8 @@
-use crate::curve::{CurveCfg, ScalarCfg, ScalarField};
+use crate::curve::{CurveCfg, ScalarField};
 use icicle_core::ecntt::Projective;
 use icicle_core::impl_ecntt;
 
-impl_ecntt!("bn254", bn254, ScalarField, ScalarCfg, CurveCfg);
+impl_ecntt!("bn254", bn254, ScalarField, CurveCfg);
```
yshekel, line 5: I don't get why CurveCfg remains

wrappers/rust/icicle-rings/icicle-labrador/src/balanced_decomposition/mod.rs
```diff
@@ -1,14 +1,7 @@
...
+// Since PolyRing doesn't implement PrimeField, we can't implement BalancedDecomposition for it.
+// We'll need to implement this differently or through the base field.
 
 #[cfg(test)]
 pub(crate) mod tests {
-    use crate::polynomial_ring::PolyRing;
-    use icicle_core::impl_balanced_decomposition_tests;
-
-    impl_balanced_decomposition_tests!(PolyRing);
+    // Tests are also removed since PolyRing doesn't implement PrimeField
```
yshekel, line 6: What???? This makes me sad

wrappers/rust/icicle-core/src/rns/mod.rs
LeonHibnik: @ChickenLover

examples/rust/arkworks-icicle-conversions/src/main.rs
```diff
@@ -70,7 +73,7 @@ fn incremental_ark_projective_points(size: usize) -> Vec<ArkProjective> {
 fn from_ark<T, I>(ark: &T) -> I
 where
     T: Field,
-    I: FieldImpl,
+    I: BigNum,
```
yshekel, line 76: why no IcicleField? I assumed that BigNum is mainly used internally, not by our users

examples/rust/msm/src/main.rs
```diff
@@ -94,11 +95,11 @@ fn main() {
...
 
         println!("Executing bn254 MSM on device...");
-        msm::msm(scalars, points, &cfg, &mut msm_results[..]).unwrap();
-        msm::msm(scalars, g2_points, &g2_cfg, &mut g2_msm_results[..]).unwrap();
+        msm::msm::<CurveCfg>(scalars, points, &cfg, &mut msm_results[..]).unwrap();
+        msm::msm::<G2CurveCfg>(scalars, g2_points, &g2_cfg, &mut g2_msm_results[..]).unwrap();
 
         println!("Executing bls12377 MSM on device...");
-        msm::msm(
+        msm::msm::<BLS12377CurveCfg>(
```
yshekel, line 102: is it mandatory now to use this type when using MSM? can we get rid of it like it was before? I mean the msm api

examples/rust/sumcheck/src/utils.rs
```diff
@@ -24,9 +25,9 @@ pub fn set_backend_cpu() {
     icicle_runtime::set_device(&device_cpu).unwrap();
 }
 
-pub fn generate_random_vector<F: FieldImpl>(size: usize) -> Vec<F>
+pub fn generate_random_vector<F: Field>(size: usize) -> Vec<F>
```
yshekel, line 28: Isn't this function redundant? I mean it doesn't do anything and it's exactly like generate_random().

icicle/include/icicle/fields/externs.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#define ICICLE_FFI_CONCAT3(a, b, c) a##b##c
+#define ICICLE_FFI_EXPAND_AND_CONCAT3(a, b, c) ICICLE_FFI_CONCAT3(a, b, c)
+
+#define ICICLE_DEFINE_FIELD_FFI_FUNCS(PREFIX, TYPE) \
```
yshekel, line 6: It's strange that we have PREFIX and also ICICLE_FFI_PREFIX. Also it seems we always concat {ICICLE_FFI_PREFIX, PREFIX} so can it be part of the ICICLE_FFI_PREFIX?

ChickenLover: it's C macro shenanigans. I tried using a single prefix but it doesn't work due to how macros work in C

wrappers/rust/icicle-core/src/balanced_decomposition/tests.rs
```diff
@@ -5,15 +5,15 @@ use icicle_runtime::memory::{DeviceVec, HostSlice};
...
+    F: PolynomialRing + balanced_decomposition::BalancedDecomposition<F> + GenerateRandom,
 {
     let batch = 5;
     let size = 1 << 10;
     let total_size = batch * size;
     let bases = [2, 3, 4, 16, 77];
 
     let input = F::generate_random(total_size as usize);
-    let mut recomposed = vec![F::zero(); total_size as usize];
+    let mut recomposed = vec![<F as PolynomialRing>::zero(); total_size as usize];
```
yshekel, line 16: Why is it now like that? I mean <F as PolynomialRing>.
Can we remove the PolynomialRing bound and have this test support any types that can do decomposition and random? should be possible I think

wrappers/rust/icicle-core/src/polynomials/mod.rs
```diff
@@ -1,52 +1,48 @@
...
-    Self::FieldConfig: FieldConfig,
-{
-    type Field;
-    type FieldConfig;
+use crate::ring::IntegerRing;
 
-    fn from_coeffs<S: HostOrDeviceSlice<Self::Field> + ?Sized>(coeffs: &S, size: usize) -> Self;
-    fn from_rou_evals<S: HostOrDeviceSlice<Self::Field> + ?Sized>(evals: &S, size: usize) -> Self;
+pub trait UnivariatePolynomial: Clone + Sized {
+    type Ring: IntegerRing;
```
yshekel, line 6: I think it should be  'type Coeff'  instead of 'type Ring'. It's clearer

wrappers/rust/icicle-core/src/balanced_decomposition/tests.rs
ChickenLover: Yes, removed that, this was also legacy from previous commits

wrappers/rust/icicle-rings/icicle-labrador/src/polynomial_ring.rs
```diff
@@ -1,9 +1,73 @@
...
-use icicle_core::impl_polynomial_ring;
-use icicle_core::traits::FieldImpl;
+use icicle_core::bignum::BigNum;
+use icicle_core::polynomial_ring::PolynomialRing;
+use icicle_core::traits::GenerateRandom;
+use icicle_runtime::errors::{eIcicleError, IcicleError};
 
 // Define the Polynomial Ring Zq[X]/X^d+1
-impl_polynomial_ring!(PolyRing, ScalarRing, 64, -1);
+#[derive(Clone, Copy, Debug, PartialEq)]
```
yshekel, line 8: Why did you remove the macro here?

wrappers/rust/icicle-rings/icicle-labrador/src/ring.rs
```diff
@@ -1,24 +1,23 @@
...
+use icicle_core::bignum::BigNum;
+use icicle_core::{impl_integer_ring, impl_montgomery_convertible};
+use icicle_runtime::{eIcicleError, memory::HostOrDeviceSlice, stream::IcicleStream};
 
 pub(crate) const SCALAR_LIMBS: usize = 2;
 
-impl_scalar_field!("labrador", labrador, SCALAR_LIMBS, ScalarRing, ScalarCfg);
-impl_scalar_field!("labrador_rns", labrador_rns, SCALAR_LIMBS, ScalarRingRns, ScalarCfgRns);
+impl_integer_ring!(ScalarRing, "labrador", SCALAR_LIMBS);
+impl_montgomery_convertible!(ScalarRing, "labrador_scalar_convert_montgomery");
```
yshekel, line 8: Why the montgomery macros are now split? I noticed it in all crates but why?

wrappers/rust/icicle-core/src/poseidon2/mod.rs
```diff
@@ -1,67 +1,67 @@
...
+use crate::{field::Field, hash::Hasher};
+use icicle_runtime::IcicleError;
 
 /// Trait to define the behavior of a Poseidon2 hasher for different field types.
-/// This allows the implementation of Poseidon2 hashing for various field types that implement `FieldImpl`.
-pub trait Poseidon2Hasher<F: FieldImpl> {
-    /// Method to create a new Poseidon2 hasher for a given t (branching factor).
-    fn new(t: u32, domain_tag: Option<&F>) -> Result<Hasher, eIcicleError> {
+/// This allows the implementation of Poseidon2 hashing for various field types that implement `PrimeField`.
+pub trait Poseidon2Hasher: Field {
```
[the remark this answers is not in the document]
ChickenLover: I was thinking of it as of a way to remove redundant template parameters because in all of our codebase it is implemented for a Field over which it operates anyway. But ok, I returned it to have a generic param

wrappers/rust/icicle-rings/icicle-labrador/src/polynomial_ring.rs
ChickenLover: hmm, I don't remember doing this. @LeonHibnik was that your change?

examples/rust/msm/src/main.rs
yshekel: > That's a good question. I also don't like this and I tried to fix that but here is the situation: Previously we had Projective and Affine so that way MSM API could infer the curve type just by looking at projective types. Now Projective and Affine do not have any Curve reference inside (because it's the opposite - curve is created after the Projective and Affine) There are three possible solutions:
> 
> 1. What we currently have - specify CurveConfig type
> 2. Only use types that are taken through C::Type syntax
> 
> ```rust
> let mut msm_results_bls12377 = DeviceVec::<BLS12377G1Projective>::device_malloc(1).unwrap();
> // becomes:
> let mut msm_results_bls12377 = DeviceVec::<<BLS12377CurveCfg as Curve::Projecitve>>::device_malloc(1).unwrap();
> ```
> 
> 3. Change MSM API to have different template params. Like for example Projective, since Projective has Affine and ScalaField definitions inside. But then there is a question - what type should implement the MSM trait? If we keep Curve, than we need to check that curve implements MSM trait in the floating function, which brings us nowhere. If we change it to Projective - we can actually get rid of CurveCfg everywhere, it will work and from code context it sounds really nice, but logically it feels a little wrong to make Projective the curve defining type..
> 4. ~A slightly hacky way - define the actual Affine and Projective types privately and make exposed types as so~ This doesn't work:
> 
> ```
>         pub type $affine_type = $curve::Affine;
>         pub type $projective_type = $curve::Projective;
> ```
> 
> @yshekel wdyt?

I think option (3). You can implement it for projective or for affine or both even (I think it won't matter for users, right?). Then indeed I think we don't need the curve trait, as you mentioned. APIs such as Montgomery conversions are probably implemented for each of them. Others like MSM I guess you can choose which one to use. Let's say projective: now I think you only need to use a type that implements projective.

yshekel (approved):

PR #993, opened 2025-07-03
V4 rust restruct errors

nonam3e:
This PR adds rust wrapped errors

Haven't changed polynomial api a lot, it seems it won't be that simple to use as now if Results are added there

cuda-backend-branch: main

metal-backend-branch: main


nonam3e (commented):
merged ok

nonam3e:
> Is it different from the other PR? I am confused

it's the same, it's just a pr into `V4-rust-restruct` instead of `V4-new`, so after merging `V4-rust-restruct` into `V4-new` no extra changes will be needed, or is there a better approach?

yshekel (approved):

LeonHibnik (approved):

PR #996, opened 2025-07-06
a minor fix

ChickenLover:
## Describe the changes

## Describe the rational

Why is this change necessary?

Does it increase performance?

## Backend branches

cuda-backend-branch: main

metal-backend-branch: main

### DIFF
--- a/wrappers/rust/icicle-core/src/bignum.rs
+++ b/wrappers/rust/icicle-core/src/bignum.rs
@@ -1,0 +1,157 @@
+use std::fmt::{Debug, Display};
+
+use crate::traits::Zero;
+
+pub trait BigNum:
+    Display + Debug + Copy + PartialEq + Clone + Into<Self::Limbs> + From<Self::Limbs> + Send + Sync + Zero
+{
+    const LIMBS_SIZE: usize;
+    type Limbs: AsRef<[u32]> + AsMut<[u32]> + Copy;
+
+    fn limbs(&self) -> &Self::Limbs;
+    fn limbs_mut(&mut self) -> &mut Self::Limbs;
+
+    fn to_bytes_le(&self) -> Vec<u8> {
+        self.limbs()
+            .as_ref()
+            .iter()
+            .flat_map(|&x| {
+                x.to_le_bytes()
+                    .to_vec()
+            })
+            .collect()
+    }
+
+    // please note that this function zero-pads if there are not enough bytes
+    // and only takes the first bytes in there are too many of them
+    fn from_bytes_le(bytes: &[u8]) -> Self {
+        let mut result = Self::zero();
+        let limbs = result
+            .limbs_mut()
+            .as_mut();
+        for (i, chunk) in bytes
+            .chunks(4)
+            .take(Self::LIMBS_SIZE)
+            .enumerate()
+        {
+            let mut chunk_array: [u8; 4] = [0; 4];
+            chunk_array[..chunk.len()].clone_from_slice(chunk);
+            limbs[i] = u32::from_le_bytes(chunk_array);
+        }
+        result
+    }
+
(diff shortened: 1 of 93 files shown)


wrappers/rust/icicle-core/src/polynomials/mod.rs
```diff
@@ -1,50 +1,48 @@
...
-where
-    Self::Field: PrimeField,
-{
-    type Field;
+pub trait UnivariatePolynomial: Clone + Sized {
+    type Ring: IntegerRing;
 
-    fn from_coeffs<S: HostOrDeviceSlice<Self::Field> + ?Sized>(coeffs: &S, size: usize) -> Self;
-    fn from_rou_evals<S: HostOrDeviceSlice<Self::Field> + ?Sized>(evals: &S, size: usize) -> Self;
+    fn from_coeffs<S: HostOrDeviceSlice<Self::Ring> + ?Sized>(coeffs: &S, size: usize) -> Self;
```
yshekel, line 8: I wonder if it should be field or ring. Some apis like divions requires inverse I think

wrappers/rust/icicle-curves/icicle-bls12-377/src/curve.rs
```diff
@@ -1,42 +1,24 @@
...
 pub(crate) const BASE_LIMBS: usize = 12;
 #[cfg(feature = "g2")]
 pub(crate) const G2_BASE_LIMBS: usize = 24;
 
-impl_field!(ScalarField, "bls12_377", SCALAR_LIMBS, true);
-impl_field_arithmetic!(ScalarField, "bls12_377", bls12_377_sf);
-impl_montgomery_convertible!(ScalarField, bls12_377_scalar_convert_montgomery);
-impl_generate_random!(ScalarField, bls12_377_generate_scalars);
+impl_field!(ScalarField, "bls12_377", SCALAR_LIMBS, true, true);
+impl_montgomery_convertible_ffi!(ScalarField, bls12_377_scalar_convert_montgomery);
```
yshekel, line 10: why is it named  *_ffi? I mean other macros don't have this ffi in the name

examples/rust/sumcheck/src/main.rs
```diff
@@ -124,7 +125,7 @@ pub fn main() {
     );
     //try different combine functions!
     let combine_function =
-        <icicle_bn254::program::bn254::FieldReturningValueProgram as ReturningValueProgram>::new_predefined(
+        <icicle_bn254::program::bn254::RingReturningValueProgram as ReturningValueProgram>::new_predefined(
```
LeonHibnik, line 128: Rename to `ReturningValueProgram` maybe? this doesn't make any sense here..

ChickenLover: `ReturningValueProgram` is already taken by the trait

wrappers/rust/icicle-core/src/bignum.rs
```diff
@@ -0,0 +1,187 @@
...
+            }
+        }
+
+        impl icicle_core::bignum::BigNum for $bignum {
+            const LIMBS_SIZE: usize = $num_limbs;
+            type Limbs = [u32; $num_limbs];
+
+            fn limbs(&self) -> &Self::Limbs {
+                &self.limbs
+            }
```
nonam3e, line 120: Are there any cases when this function doesn't work this way? Maybe we can implement it in trait declaration?

examples/rust/sumcheck/src/main.rs
LeonHibnik: `ReturningValProgram` Remove the `Ring` or `Field` prefix

LeonHibnik (approved):
Nice rework of the rust wrapper!

PR #1007, opened 2025-07-10
vecops split

vhnatyk:
## Describe the changes

This PR  splits VecOps Rust trait

## Describe the rational

Why is this change necessary?
Refactoring Rust VecOps trait for modularity and upcoming braking changes

Does it increase performance?
No

## Backend branches

cuda-backend-branch: main

metal-backend-branch: main

### DIFF
--- a/wrappers/rust/icicle-core/src/vec_ops/mod.rs
+++ b/wrappers/rust/icicle-core/src/vec_ops/mod.rs
@@ -37,10 +37,10 @@ impl VecOpsConfig {
 #[doc(hidden)]
 pub trait VecAdd<T> {
     fn add(
-        a: &(impl HostOrDeviceSlice<T> + ?Sized),
-        b: &(impl HostOrDeviceSlice<T> + ?Sized),
-        result: &mut (impl HostOrDeviceSlice<T> + ?Sized),
-        cfg: &VecOpsConfig,
+        _a: &(impl HostOrDeviceSlice<T> + ?Sized),
+        _b: &(impl HostOrDeviceSlice<T> + ?Sized),
+        _result: &mut (impl HostOrDeviceSlice<T> + ?Sized),
+        _cfg: &VecOpsConfig,
     ) -> Result<(), IcicleError> {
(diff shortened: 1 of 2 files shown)


wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -35,91 +35,169 @@ impl VecOpsConfig {
...
+pub trait VecAdd<T> {
     fn add(
         a: &(impl HostOrDeviceSlice<T> + ?Sized),
         b: &(impl HostOrDeviceSlice<T> + ?Sized),
         result: &mut (impl HostOrDeviceSlice<T> + ?Sized),
         cfg: &VecOpsConfig,
-    ) -> Result<(), IcicleError>;
+    ) -> Result<(), IcicleError> {
+        // Default implementation - this should be overridden by specific implementations
+        unimplemented!("VecAdd::add not implemented for this type")
```
yshekel, line 46: (1) Now that you broke the VecOps trait, why do you need a default implementation? to me it makes sense if you have a bundle of vecops and you want to allow partial implementation.
(2) This is a breaking change, and we already released V4 ...

wrappers/rust/icicle-core/src/vec_ops/mod.rs
```diff
@@ -562,7 +659,7 @@ macro_rules! impl_vec_ops_field {
             }
         }
 
-        impl VecOps<$field> for $field {
+        impl VecAdd<$field> for $field {
```
yshekel, line 662: Not sure if we should do it or not but maybe macro per trait is better now

yshekel (approved):

wrappers/rust/icicle-core/src/vec_ops/mod.rs
vhnatyk: I think we do need to have possibility to avoid implementing everything for every field/curve - probably in other PR as you suggested based on priorities

PR #1009, opened 2025-07-15
[FIX] bug in random sampling slow mode

yshekel:
Bug fix for "slow" sampling mode when size is not a power of 2. 

cuda-backend-branch: yshekel/random_sampling_bugfix


goforashutosh (approved):
approving because Yuval asked

PR #1013, opened 2025-07-16
Add files via upload

Shalmoni:
## Describe the changes

## Describe the rational

Why is this change necessary?

Does it increase performance?

## Backend branches

cuda-backend-branch: main

metal-backend-branch: main


yshekel (approved):
Approved. Note that we can make it much faster when we use gpu memory rather than host memory only. It's an optimization that is planned but not yet implemented. Maybe worth mentioning this

PR #1018, opened 2025-07-22
[FEAT] Remove force inline from mul_wide

LeonHibnik:
## Describe the changes

This PR removes the INLINE directive from the mul_wide methods in field extension classes (complex, cubic, and quartic). This change optimizes the build process by allowing the compiler to make better decisions about inlining these functions.

## Describe the rational

Reduce compile time as the compiler doesn't need to inline large functions at every call site

## Backend branches

cuda-backend-branch: main

metal-backend-branch: main

### DIFF
--- a/icicle/include/icicle/fields/complex_extension.h
+++ b/icicle/include/icicle/fields/complex_extension.h
@@ -199,7 +199,7 @@ public:
     }
   }
 
-  constexpr HOST_DEVICE_INLINE Wide mul_wide(const ComplexExtensionField& ys) const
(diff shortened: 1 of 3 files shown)


HadarIngonyama (commented):
Doesn't this affect performance?

LeonHibnik:
> Doesn't this affect performance?
@HadarIngonyama From what I tested, no. ~8 ms over ~1120 ms was the maximum effect I measured in our tests, some runs had no effect

yshekel (approved):

jeremyfelder (approved):

PR #1024, opened 2025-08-06
add image

Shalmoni:
added image to intro.md


lnm-at-work (approved):
I think "device-specific backend" might actually be a bit more accurate.

PR #1026, opened 2025-08-10
fix: Add error checking for CUDA memory allocations in tests

DeVikingMark:
Add proper error checking for all CUDA memory allocation calls in ML-KEM encapsulation tests to prevent silent failures and improve debugging.


DeVikingMark:

yshekel (approved):

LeonHibnik (approved):

PR #1034, opened 2025-09-02
chore(fri-capi): add nullptr guards and early returns in FRI C API

sashass1315:
- Add missing nullptr check for final_poly in fri_proof_get_final_poly.
- Validate result is non-null in fri_proof_get_serialized_size and extension variant.
- Return INVALID_POINTER on null merkle hash handles in fri_merkle_tree_prove/verify and extension variants to avoid UB.
- Aligns pointer-check behavior with conventions in sumcheck and merkle C APIs.


sashass1315:

PR #1043, opened 2026-04-03
Fix reduce_from_bytes to use Barrett reduction instead of buggy iterated subtraction

MavenRain:


MavenRain:

